PR #205, opened 2021-05-10
Set interpolation for Normal Maps to 'closest'

Loewetiger:
Currently, the Texture Interpolation for the Normal Map generated when using extra maps is set to "Linear". This leads to the textures looking like they were badly upscaled (very blurry), even though the Specular and Normal Map are correctly set to "Closest". I am not certain if this was an oversight or done intentionally, but I think it is generally way more desirable to interpolate all the Maps with the "Closest" setting.
I have attached some images, so you can get an idea of the differences:
| SEUS PTGI Shader | Current "Linear" interpolation | New "Closest" interpolation |
| --- | --- | --- |
| SEUS Shader | Current linear interpolation | New closest interpolation |


TheDuckCow:
Thank you for the clear screenshot demonstration, and apologies for the delay! Merging this into the upcoming release.

Issue #274, opened 2022-01-08
Revamp Materials System


StandingPadAnimations:
# Current proposal
Instead of generating materials in Python directly, we should create materials in a blend file that is then sourced by the Prep Materials operator. Each material would have the format `<pack-format>_mcprep-default`. This would make it easier for users to add new pack formats, edit existing pack formats, and reduce overall code size. The downside however is that MCprep dev builds would require a blend file for all pack formats.

In this blend file, there would also be a JSON file that may look something like this (I know this example sucks, but bare with me):
```json
{
    "pack-format-1" : {
        "has_emission" : true
        ...
    }
    ...
}
```

This JSON file will define extra attributes that each material has, such as whether a material has emission nodes or not. We could expand this further to also allow mapping materials to version or even engine.

# Old proposal (kept for transparency)
MCprep's sync materials option allows users to make custom materials that get automatically applied when prepping. But currently, it mainly works on a per-material basis. As Minecraft gets updated and the number of blocks grows, it becomes harder and harder for users to change the material for every block. This also makes using external render engines hard as all materials have to be changed.

A solution to these problems would be a default material. Basically, MCprep would look for a material in the `materials.blend` file with the naming sequence `default_packformat_blenderversion_renderengine`. Then when prepping, it'll just swap textures.

This would also benefit development. It would move the default material setup out of Python, which makes navigating MCprep code much easier. With the packformat part of the name, it would allow users to make completely different materials for different formats. And users in older versions of Blender can make node-setups that work for them using the blenderversion part of the name.

For example, a simple material for 2.8 EEVEE would be named `default_simple_28_eevee` and one for seus 3.0 Cycles would be called `default_seus_30_cycles`

StandingPadAnimations:
For versions like 2.83, it would be an extra digit like `283`. For 2.93.5 however, it would simply be `293`

Roni-Raihan:
If you extract using mineways, the material will be different, but have the same texture file. Only, the location of the uv is different.

StandingPadAnimations:
That has issues when using texture packs

StandingPadAnimations:
And also, it does not work for PBR textures since they require separate maps

TheDuckCow:
I think this is going to be a great redesign, thanks for reporting the idea and the follow up points. It will also help clean up a fair bit of legacy code.

[...]

Issue #239, opened 2021-08-22
Optimizer for MCprep


StandingPadAnimations:
Currently, MCprep does some small optimizations to scenes in Cycles, but it would be nice to have a optimizer built for MC animation. Maybe 2 options(Quality and Speed) for CPU, CUDA, OptiX, and OpenCL. In later versions of Blender, we could also take advantage of adaptive sampling(perhaps a flag that tells MCprep that the user is using a later version of Blender)

TheDuckCow:
Would love to hear your thoughts of all the settings you'd apply! Let me know if you think differently, but this could be great for the "prep world" button (which tbh should probably be renamed to "optimize" or something like that, which is mostly what it does right now). We could definitely have a dropdown property for different "modes", and even be reactive based on Eevee or Cycles being selected.

StandingPadAnimations:
I think it could have light bounce optimization. Firstly by setting max light bounces to 8 and transparency to 12 and some manual input for the other bounces like:
```
How many glossy materials do you have?
[ ] None or not a lot(1 to 2 light bounces)
[ ] a lot(2 to 3 light bounces)
```
Preferably this would be a slider for better visualization(at least for the user). I would also suggest adaptive sampling but it wouldn't be useful for 2.7x users. Ill try working on a concept for how it might look

TheDuckCow:
Awesome, appreciate your willingness to mock it up! And yes, it shouldn't be a problem for use to insert some checks for "is blender 2.7x, then don't show XYZ option", MCprep already does that in a few places.

StandingPadAnimations:
With Blender now getting Cycles X in the 3.0 release, the optimizer may also need another check for Cycles X features and changes

StandingPadAnimations:
Ok, i'vs started work on it while testing some stuff for #273. I just need to iron out some bugs and make the UI easier to work with before submitting a pull request

TheDuckCow:
Amazing, thanks a ton! Let me know if there's anywhere you are looking for inputs from me, and feel free to open a pull request even before it's ready (you can mark as a draft), just so I can see what you are doing and avoid overlap. On Apr 17, 2022 9:23 PM, StandingPad ***@***.***> wrote:
Ok, i'vs started work on it while testing some stuff for #273. I just need to iron out some bugs and make the UI easier to work with before submitting a pull request

StandingPadAnimations:
I will close this now

PR #290, opened 2022-04-18
Cycles Optimization and default materials

StandingPadAnimations:
This aims to add Cycles optimization as a feature to MCprep as mentioned in #239. Some of the features planned include:
- [x] Old Cycles support (includes optimizing tile sizes)
- [x] Cycles X support  
- [x] Material settings for optimization
- [x] Time of Day settings
- [x] Some basic optimizations for the Improve UI feature
- [x] Automatically switching to the GPU if detected
- [x] ~~Option to prioritize speed over quality~~

This feature is not related to Cycles but was mentioned in #274 and will help those wanting to use other render engines:
- [x] Default material definition in materials.blend (this will be made separate from Sync Materials)

These features should help people who have trouble with Cycles render times

~~EDIT: with Blender 3.2, support for the new OBJ importer will be added. It should be relatively simple as I already made a general Blender version function to detect the version being used.~~
- [x] ~~Support for the new C++ based OBJ Importer in Blender 3.2 as mentioned in #305~~ 

EDIT: the new OBJ importer seems to be enabled by default


StandingPadAnimations:
https://user-images.githubusercontent.com/75058058/163898299-fa633983-f3b0-43e8-80e8-34538682ce00.mp4

I've been working on the optimizer, here's a preview of how it looks

and yes, I still have bugs to iron out regarding sampling

StandingPadAnimations:
So I've been talking with the developers of the super fast render addon on how they do testing, and I think I can try to implement a very basic version of one of their tests (adjusting samples and clamping within a set range and see how much brightness changes). However, it may add an extra dependency to MCprep, so I'm still on the fence on that one

StandingPadAnimations:
That dependency will likely be ski-image and/or numpt

TheDuckCow:
Thanks for starting the pr and showing the video, awesome stuff. I'll see about taking a look through.

One thought I have is that the "prep world" button right now mostly acts as an optimizer itself, I wonder if it can be merged into what you are doing here?

And the other guiding advice, is to simplify as much as possible, since many users will not be familiar with all the tradeoffs and be daunted by too many controls (you can always have subpanels nested via `bl_parent_id` for more advanced control). 

Can you share more about what you mean by testing? Is this testing the addon code itself, or more like testing renders to see if changes have substantial impact or not?

StandingPadAnimations:
Most of the features can be implemented in prep world. I'll do that in the next commit

As for the latter question, it's more like testing renders for differences (mostly brightness). If the difference is less then say 10%, then that's the most the render will be optimized. I think I'll make this an advanced feature since it may be confusing for the rest of the userbase

StandingPadAnimations:
Regarding the benchmark idea, I think I'll put it on the backburner since pip in Blender is annoying to deal with on Windows

StandingPadAnimations:
The optimizer now takes place when the world is preped. There is also a new panel for advanced settings

StandingPadAnimations:
I've gotten a proof of concept for default materials. So far, there's no PBR support, but I'll check it off the list since it should be fine for the vast majority of use cases
https://user-images.githubusercontent.com/75058058/164956902-96477a89-cd90-48f8-8e3f-e00c7b18adf9.mp4

StandingPadAnimations:
I've also just removed "sync_default" from the material names (that was for testing purposes) so Sync Materials should still work as intended

TheDuckCow:
I haven't had a real chance to take a look through this yet - if you think it's a good time for me to start an initial review (we can do multiple), let me know!

StandingPadAnimations:
Current state of this pull request regarding features that need to be added
# Cycles Optimization
* Support the older version of Cycles 
* Optimize volumetric materials either using the heterogeneous volume option or using the camera ray trick (if it doesn't affect results much)
# Default materials
* Automatically generate a template for default materials to make it easier user wise (mostly due to the use of Python API stuff regarding engine names)
* Add PBR support (which shouldn't be too bad)
* Allow default materials to be used when prepping without needing a intermediate material (can't remember if I already added this or not)

# Features planned for a later time:
* Ability for developers to override the prep materials function for better engine support. This would mean some other functions would have to made regarding stuff like textures (which is why I won't get to it now). This could be done with a special folder that MCprep can load when Blender starts up (and some Python dictionarys)
* Image difference based optimization: not sure if this will actually be beneficial for the end user, I'll have to experiment on this one. The main issue is trying to install Python modules (though I think there's an easier way in Blender)

Other then everything above, I'd say this pull request is 98% complete

StandingPadAnimations:
I've removed the "Speed over Quality" option from the plan as it wouldn't help much

StandingPadAnimations:
Alright, I've added all I've planned to add for this pull request. Ready when you are @TheDuckCow

TheDuckCow:
Thanks a bunch @StandingPadAnimations starting my initial code review shortly. I'll go in two passes:

1. Start with high level functionality, seeing where things are in the UI, I may request some changes which will all come at once once I release the code review. I may ask for some pivots and moving of pieces from one area to another, splitting or join functions, higher level design stuff etc.
2. After above changes are made, I'll do second pass which is more about particulars of code styles and more minor things; these will be things I could suitably do myself by pushing onto your branch, but you are welcome to make the changes too (I'll ask up front to give the chance; but if it's really minor things, I'll just directly push the changes and share the commit(s) with you for future reference).

StandingPadAnimations:
Just a heads up, I won't be able to make changes to anything UI related (code style related should be fine) from Wednesday of next week to early/mid July since I'll be on vacation

MCprep_addon/materials/default_materials.py
```diff
@@ -0,0 +1,216 @@
+# ##### BEGIN GPL LICENSE BLOCK #####
```
TheDuckCow, line 1: Is there a default blend file that you have locally? You should include this in your commit too - and add it using git lfs add <filename> so it is trcaked as an object file and not tracked for diffs like code.

MCprep_addon/materials/default_materials.py
```diff
@@ -0,0 +1,216 @@
...
+    util.bAppendLink(os.path.join(sync_file, "Material"), import_name, False) # no linking
+
+    imported = set(bpy.data.materials[:]) - set(init_mats)
+    if not imported:
+        return "Could not import " + str(material.name)
+    new_material = list(imported)[0]
+    
+    # checking if there's a node with the label Texture
+    new_material_nodes = new_material.node_tree.nodes
+    if not new_material_nodes.get("Default Texture"):
```
TheDuckCow, line 64: See if we can follow the MCprep convention that already exists for "tagging" the nodes within the material to get the diffuse image. Agreed that this first pass does not need to support PBR, but good to try and follow the convention so it's a more natural upgrade later: https://github.com/TheDuckCow/MCprep/blob/master/MCprep_addon/materials/generate.py#L474

TheDuckCow (changes requested):
Initial round of comments provided here - mix of FYIs, some questions to help my understanding, as well as some specific changes

StandingPadAnimations:
Regarding the optimizer:
* A line of code to check for the auto file size addon and enable it could be added as to not reinvent the wheel. Otherwise it'll just use 256² for GPU and 32² for CPU 
* The day and night options are more light bright and dark options. Generally dark scenes need more samples and higher render settings to account for the lack of light. I'll rename the settings to light and dark.
* Sampling also admittedly isn't the best, so I'll change that.
* Helper functions: didn't know there was one for preferences. I'll be sure to use that

Regarding the default materials:
* I think adding default materials to `generate.py` should be fine.
* When the operator copies the material, it uses the copy as the new material. This could be replace with "copy and paste nodes" 
* I actually didn't know MCprep had standardized names for materials. I'll be sure to use those.

Hopefully I'll get these changes done later this week or early next week.

StandingPadAnimations:
I've pushed some changes mainly in code style as well as adding caching to the bvXX functions

MCprep_addon/MCprep_resources/optimize_scene.py
```diff
@@ -185,13 +199,13 @@ def execute(self, context):
...
                 
-            if util.getCycles() == 2:
+            if util.get_cycles_version() == 2:
                 bpy.context.scene.render.tile_x = 32
                 bpy.context.scene.render.tile_y = 32
         
         elif CyclesComputeDeviceType == "CUDA" or CyclesComputeDeviceType == "HIP":
             if CurrentRenderDevice == "CPU":
-                if bpy.context.preferences.addons["cycles"].preferences.has_active_device():
+                if util.get_preferences("cycles").preferences.has_active_device():
```
TheDuckCow, line 208: Careful - if you are indending to get the Cycles preferences setting, this function `get_preferences` only has an input which is a bpy context or none. I think what you want here is:

```
prefs = util.get_preferences(context)
cprefs = prefs.addons.get("cycles")
if cprefs is not None and cprefs.preferences.has_active_device():
    # do stuff
```

Consider grabbing cycles preference once towards the top of the function, and just referencing it where needed (noting that this usage of ".get" is safe, in case someone has cycles disabled which is possible).

MCprep_addon/materials/default_materials.py
```diff
@@ -126,20 +126,19 @@ def execute(self, context):
             return {'CANCELLED'}
         
         if sync_file == bpy.data.filepath:
-			
             return {'CANCELLED'}
 
         # ------------------------- Find the default material ------------------------ #
-        MaterialName = f"default_{self.SIMPLE if not self.UsePBR else self.PBR}_{self.Engine}"
-        if not default_material_in_sync_library(MaterialName, context):
+        material_name = f"default_{self.SIMPLE if not self.UsePBR else self.PBR}_{self.engine}"
```
TheDuckCow, line 132: I didn't catch this before: f-strings aren't available in older python versions (ie before blender's python in 2.80), plus this comprehension is rather long anyways. I'd use an old-style string formatter:

```
material_name = "default_{form}_{engine}".format(
    form=self.SIMPLE if not self.UsePBR else self.PBR,
    engine=self.engine
)
```

MCprep_addon/materials/default_materials.py
```diff
@@ -170,11 +169,11 @@ def create_default_material(self, context, engine, type):
         engine: the render engine in lowercase
         type: the type of texture that's being dealed with
         """
-        MaterialName = f"default_{type}_{engine}"
-        DefaultMaterial = bpy.data.materials.new(name=MaterialName)
-        DefaultMaterial.use_nodes = True
-        nodes = DefaultMaterial.node_tree.nodes
-        links = DefaultMaterial.node_tree.links
+        material_name = f"default_{type}_{engine}"
```
TheDuckCow, line 172: Same here for f-string

MCprep_addon/util.py
```diff
@@ -183,14 +184,42 @@ def obj_copy(base, context=None, vertex_groups=True, modifiers=True):
...
+
 def bv28():
 	"""Check if blender 2.8, for layouts, UI, and properties. """
 	global BV_IS_28
 	if not BV_IS_28:
-		BV_IS_28 = hasattr(bpy.app, "version") and bpy.app.version >= (2, 80)
+		BV_IS_28 = getBV() >= (2, 80)
 	return BV_IS_28
 
+def bv30():
```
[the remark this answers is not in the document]
TheDuckCow: Thanks for making this change, definitely cleans these functions up a nice bit!

MCprep_addon/MCprep_resources/optimize_scene.py
```diff
@@ -1,58 +1,72 @@
...
-#                   Optimizer code by StandingPad Animations                   #
-# ---------------------------------------------------------------------------- #
 class MCprepOptimizerProperties(bpy.types.PropertyGroup):
 
-	def TimeInScene(self, context):
+	def scene_brightness(self, context):
 		itms = [
-        ("Day", "Day time", ""), 
-        ("Night", "Night time", "")
+        ("bright", "scene is bright", ""),
```
TheDuckCow, line 26: If a user can hove rover to get tooltip text, then it'd be good to fill out that third string; something a bit more verbose like for dark: "Use settings that are optimized for dark scenes\nlike night time, interiors, or caves."

MCprep_addon/load_modules.py
```diff
@@ -47,10 +47,10 @@
 else:
 	from .materials import prep
 
-if "OptimizeScene" in locals():
-	importlib.reload(prep)
+if "optimize_scene" in locals():
+	importlib.reload(optimize_scene)
 else:
-	from .SceneOptimizer import OptimizeScene
+	from .MCprep_resources import optimize_scene
```
TheDuckCow, line 53: Really sorry if I wasn't clear about this - I meant that it could read template blend files from MCprep_resources, but that I'd expect the `optimize_scene.py` file itself to live in the root `MCprep_addon` folder (one level up).

TheDuckCow (changes requested):
Thanks for making these style changes and addressing other comments, getting closer here.

StandingPadAnimations:
Regarding https://github.com/TheDuckCow/MCprep/pull/290#discussion_r874333667, would it make sense to only replace the materials for selected objs? It would help in this case

StandingPadAnimations:
I've updated the pull request with the goal of supporting the new OBJ importer in mind. Here's also an updated list on improvements to the Cycles Optimization and Default materials: 

# Cycles Optimization 
## High Priority 
- [ ] Using existing lists to better optimize the scene
- [ ] Adding proper tooltips for some operators 
- [ ] Taking advantage of the auto tile size addon for older versions of Blender
## Low Priority 
- [ ] Adding a new test for the optimizer (though I'm not sure how this test could be designed)
- [ ] Optimization based on render tests for light paths and sampling (although I would say that it might be best to leave this for later)

# Default Materials
## High Priority
- [ ] Avoid copying so many materials
- [ ] Better template creation for default materials
- [ ] PBR material support
## Low Priority 
- [ ] Allowing materials based on Blender version (this will need some work and honestly is low priority)
- [ ] Allowing custom prep material scripts for those wanting to support render engines that can't use the current system

TheDuckCow:
Thanks for adding the OBJ importer - that's to include for this, though normally would recommend separate branches and PRs for completely unrelated things ( I guess you could argue it's an "optimization" though. Doing a latest code review soon; might be good to not add new functionality, merge, and then do a follow up branch, unless you think you're still to make substantial changes.

With 1.19 out, I do want to do a new MCprep update somewhat soon, and would be nice to have this in as a feature to claim. But will need a week or to make sure everything fits with the UI.

StandingPadAnimations:
> Thanks for adding the OBJ importer - that's to include for this, though normally would recommend separate branches and PRs for completely unrelated things ( I guess you could argue it's an "optimization" though. Doing a latest code review soon; might be good to not add new functionality, merge, and then do a follow up branch, unless you think you're still to make substantial changes.
> 
> With 1.19 out, I do want to do a new MCprep update somewhat soon, and would be nice to have this in as a feature to claim. But will need a week or to make sure everything fits with the UI.

Hopefully, I'll get most or all the necessary fixes done later next week. 

Regarding the OBJ importer, it seems like it's enabled by default in Blender 3.2, so MCprep supports it by default

TheDuckCow:
Great, next week works well timing wise to work on it.

Re OBJ: Ah ok, so it's the same operator call? Any differences in the arg inputs? Mainly want to ensure backwards compatibility is preserved, but that'll come up in my automated testing regardless so don't worry too much yourself on it.

Sending code review shortly

MCprep_addon/util.py
```diff
@@ -186,19 +186,19 @@ def obj_copy(base, context=None, vertex_groups=True, modifiers=True):
         
 
 @functools.cache
-def getBV():
+def getBV(major_version, minor_version):
 	if hasattr(bpy.app, "version"):
-		return bpy.app.version
+		return True if bpy.app.version >= (major_version, minor_version) else False
```
TheDuckCow, line 191: Simplify for the same effect:

```
return bpy.app.version >= (major_version, minor_version)
```

MCprep_addon/util.py
```diff
@@ -186,19 +186,19 @@ def obj_copy(base, context=None, vertex_groups=True, modifiers=True):
         
 
 @functools.cache
-def getBV():
+def getBV(major_version, minor_version):
```
TheDuckCow, line 189: Maybe we rename to: `def min_bv(version)` which better describes which it does - and then version is just a tuple input (so we can be more granular).

MCprep_addon/util.py
TheDuckCow: If changing the input to be a tuple, then this just becomes `return bpy.app.version >= version` where version is a tuple, noting that with python logic all of these are True statements:

```
(1, 2, 3) > (1, 2, 0)
(1, 3) > (1, 2, 3)
(2, 80) > (2, 79)
(3,) > (2, 79)  # Comma forces it to be treated as a tuple
(1, 0, 0) > (1, 0) # Interesting edge case to be aware of.
```

MCprep_addon/materials/generate.py
```diff
@@ -1334,6 +1334,9 @@ def matgen_cycles_simple(
 	else:
 		principled.inputs["Metallic"].default_value = 0
 
+	# Specular tends to cause some issues with how blocks look, so let's disable it
+	principled.inputs["Specular"].default_value = 0
```
TheDuckCow, line 1338: DO we think this is true for _all_ blocks? Does this hold for PBR packs? Generally as I understand, the realistic usage of specular of pbr rendering with cycles's principled node is to leave spec at 0.5 (but then roughness drives how rough ofc, so that may be the main issue here).

TheDuckCow (changes requested):
More minor requests for changes - big thanks again for updating the naming convention, much nicer now and more consistent with the rest of the codebase.

StandingPadAnimations:
Regarding default materials, I think it'll be best to add them in a later update. Adding PBR support would require a rewrite and likely wouldn't be done by the end of next week. Plus, default materials have had quite a bit of bugs I still need to kink out. I'll comment out all UI related things regarding default materials. 

What definitely will be finished by the end of next week is Cycles optimizations. Since that was the original intent of this pull request, it makes sense to go forward with it. I still need to make sure it works on older versions (2.93 seems to return an error regarding a missing import that I'll need to investigate)

TheDuckCow:
Sounds like a plan @StandingPadAnimations!

StandingPadAnimations:
I think the optimizer is ready to be merged. There's still some stuff I want to do like replacing mesh lights with actual lights or doing test renders to further optimize light bounces, but those can be saved for their own pull requests

StandingPadAnimations:
Regarding settings that can be enabled, would it be wise to enable `Automatic Scrambling Distance` if the user is using a GPU and Blender 3.0+?

The main issue is artifacts that may occur, but is it worth risking artifacts for much better performance on the GPU?

Hopefully I'll have an answer before this PR is merged

TheDuckCow:
Hey @StandingPadAnimations - just a heads up that I'm going to make a point to review this over this weekend, and agree we should try to merge before going too much further.

Let's just make a decision for Automatic Scrambling Distance and stick with it. This already will be a button that people have to explicitly press in order to optimize their scenes, so I think it's ok if there's some valid tradeoff optimizations.

Re replacing mesh lights with actual lights: Let's save that for another pull request at another time, as that would potentially tie in with other features.

StandingPadAnimations:
After thinking this over, I think `Automatic Scrambling Distance` isn't worth enabling, mostly because it's more manual then I initially assumed

TheDuckCow:
Ok not a problem! Feel free to do any minor commit to make that the case if not already, either way I'll dive in. I'll do a live test where I install your branch and really play around with it. On Jun 25, 2022 2:27 PM, StandingPad ***@***.***> wrote:
After thinking this over, I think Automatic Scrambling Distance isn't worth enabling, mostly because it's more manual then I initially assumed

TheDuckCow:
FYI I have a local commit I'm attempting to push, I think you need to grant me permissions to push it (so also, avoid making new changes until I can push it to avoid conflict).

TheDuckCow:
Great thanks @StandingPadAnimations! Was able to push - will do live testing shortly and then follow up most likely with minor style adjustments, and then hopefully good to merge unless I come up with some questions I can't reconcile myself.

MCprep_addon/__init__.py
```diff
@@ -34,6 +34,10 @@
 
 """
 
+# ----------------------------- For any developer ---------------------------- #
+# Increment this number for each time you get a "inconsistent use of spaces and tab error"
```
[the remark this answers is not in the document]
StandingPadAnimations: This was an error I got a lot in my IDE (VScode). I assume it's something to do with my IDE settings but I'm not sure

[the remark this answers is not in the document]
TheDuckCow: It's actually probably fixed now, as there _were_ a mixture of spaces and tabs before my most recent commits here (I've changed it all to tabs, to be consistent with the rest of the repo)

StandingPadAnimations:
Before merging, I need to fix some stuff in the optimizer, mostly with light bounces

TheDuckCow:
@StandingPadAnimations as per other comment, hold off on any commits until I reply back again, as I want to push a couple changes in before merging. Will be done by end of day today (likely late though).

StandingPadAnimations:
Some answers to the questions:
* I do agree the material count is not really the best option for optimization, which is why the overall bounces will be divided by 2. Hopefully, I'll be able to optimize based on the camera view for materials so it's more representative of the scene
* I think simplifying materials as an option is a good idea as I myself have been dealing with not wanting that set to 0 in my scene
* Regarding volumetrics pre-3.0 being slower, this is mostly a Cycles thing. Volumetrics pre-3.0 created a lot of noise which meant more samples were needed which equated to longer render times. This was fixed in 3.0 as Cycles now uses ray marching for volumetrics instead of brute force path tracing
* Much of the optimizer is based on much older code, so I'll be cleaning up much of the useless stuff. The reasoning behind numbers being different per backend is based on how much faster one would expect from that backend (for example, OptiX settings are much higher quality since the user isn't likely going to notice, whereas OpenCL uses lower settings because it was much slower then CUDA and OptiX. HIP is treated the same as CUDA)

StandingPadAnimations:
Also looking at the recent commit, I just noticed I forgot to remove the `bv30()` check since the optimizer should work on older versions (that explains my issues with testing this on the 2.93 LTS)

MCprep_addon/optimize_scene.py
```diff
@@ -92,7 +102,10 @@ class MCPrep_OT_optimize_scene(bpy.types.Operator):
 
 	def execute(self, context):
 		global MAX_BOUNCES
```
TheDuckCow, line 104: This `global MAX_BOUNCES` line too can be removed actually, not needed unless you are assigning to it (which isn't advised, since `MAX_BOUNCES` should be thought of as a constant)

MCprep_addon/optimize_scene.py
```diff
@@ -92,7 +102,10 @@ class MCPrep_OT_optimize_scene(bpy.types.Operator):
...
-		MAX_BOUNCES = 8
+		global MIN_BOUNCES
```
TheDuckCow, line 105: Same for all below, don't need these global statements to use them.

MCprep_addon/optimize_scene.py
```diff
@@ -251,8 +244,9 @@ def execute(self, context):
 
 		# Render changes.
 		bpy.context.scene.render.use_motion_blur = MotionBlur
-		bpy.context.scene.render.use_simplify = True
-		bpy.context.scene.render.simplify_subdivision = 0
+		if scn_props.simplify:
```
TheDuckCow, line 247: Thanks for making this change!

TheDuckCow:
Thanks for those replies! And your comment "based on how much faster one would expect from that backend" makes sense to me. If we wanted to go the extra mile, we'd have a sort of benchmark scene and tweak the settings until they all got the same render times on the same machine with the different versions of blender/cycles... but we don't need to do that, it can just be a guiding philosphy.

TheDuckCow (approved):
Marking this as approved. Once it's merged in, I'll make further tweaks. Let me know if there were other changes you wanted to make first (or just have me make right after merging). Remember we want to merge this into dev, not master!

PR #294, opened 2022-05-04
Update README.md

StandingPadAnimations:
Changed the parts regarding Blender versions to include Blender 3.1 as a supported version

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 TO INSTALL MCprep
 ======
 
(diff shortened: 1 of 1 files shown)


TheDuckCow:
Thanks for creating this post proactively! Can we get ahead of the curve here and write in 3.2 as well? All tests pass in the automated script on my machine at least.

TheDuckCow (approved):
Realize I can just commit the change directly so all good now! Thanks again, merging shortly. I like this way of creating branches on MCprep, for individual changes, should probably do more of that myself moving forward.

PR #299, opened 2022-05-15
Effects spawning

TheDuckCow:


TheDuckCow:
Hey @StandingPadAnimations - no obligation, but if you have time and would like to, would love to have you glance through and see if things look good from your standpoint. If you don't have time no worries, I'll just merge.

MCprep_addon/spawner/effects.py
```diff
@@ -0,0 +1,1180 @@
...
+			conf.preview_collections["effects"].load(
+				"effects-{}".format(effect.index), effect_files[e_index], 'IMAGE')
+
+
+# -----------------------------------------------------------------------------
+# Support functions
+# -----------------------------------------------------------------------------
+
+
+def create_auto_footfall_particle_plane():
```
StandingPadAnimations, line 945: Seems like this doesn't do much. It may be best to comment this out so Python doesn't have to create a function object

MCprep_addon/spawner/effects.py
```diff
@@ -0,0 +1,1180 @@
...
+	bl_label = "Spawn Particle Planes"
+	bl_options = {'REGISTER', 'UNDO'}
+
+	location = bpy.props.FloatVectorProperty(
+		default=(0, 0, 0), name="Location")
+	frame = bpy.props.IntProperty(default=0, name="Frame")
+
+	# Importer helper
+	filter_glob = bpy.props.StringProperty(
+		default="*.png;*.jpg;*.jpeg;*.tiff",
```
StandingPadAnimations, line 1121: I think it would be better to put the image formats in a global variable for easy modification

test_files/addon_tests.py
```diff
@@ -1513,6 +1519,161 @@ def model_spawner(self):
 		# Test collection/group added
 		# Test loading from file.
 
+	def geonode_effect_spawner(self):
+		"""Test the geo node variant of effect spawning works."""
+		if bpy.app.version < (3, 0):
```
StandingPadAnimations, line 1524: I think we need a `maxBv` function similar to the `minBv` function added alongside the optimizer. It has a benefit of also being cacheable, something that would help a lot

StandingPadAnimations (commented):
Most of it seemed pretty good, though there are a few things I noticed

StandingPadAnimations:
if you don't mind, I'll be pushing a few commits to this branch

StandingPadAnimations:
Actually most of the changes should be added once this is merged (mainly within util.py)

test_files/addon_tests.py
TheDuckCow: Good catch to ask about this function. I don't think a maxBv is needed, as we can take any min bv and just inver the logic. You can see the change I made in [this commit](https://github.com/TheDuckCow/MCprep/commit/0a134a68636663db2b0ef66c46600d3601eab35f).

MCprep_addon/spawner/effects.py
TheDuckCow: Good idea, I've made the update now.

MCprep_addon/spawner/effects.py
TheDuckCow: Fair call - better to make it an enhancement issue vs deadcode, removed in the meantime.

TheDuckCow:
Thanks for the review @StandingPadAnimations, super nice to get another set of eyes and feedback on these things. I'm going to merge in the meantime to dev.

Issue #309, opened 2022-06-28
Adding title screen camera setup


Tiefseetauchner:
I made a little script that creates Cameras to render the title panorama, and I had the idea of implementing it in MCprep, but I didn't wanna spend too much time implementing it and learning real addon dev and then have you guys say you don't want it, so I made it a [basic gist](https://gist.github.com/Tiefseetauchner/384d9b2a04ffec15e461e96bbc6bf956) (with usage instructions)

The idea is, you spawn in 6 cameras in the right orientation, make the render res square, render them, and they should line up

I feel like adding it in here might fit, but again, idk if you want that functionality, and I don't know where you'd want to implement it. Tho I'm happy to implement it, I have nothing to do atm. I'd personally put it in the Shift+A menu somewhere, but again, idk

*Fingerguns*

StandingPadAnimations:
Looks pretty useful for resource pack developers, and the code seems to be mostly complete. @TheDuckCow Would it be possible to squeeze this in the next update?

TheDuckCow:
Thanks for sharing this! I was going back and forth thinking, hmm, it's very niche and how will people discover this... Any code added to MCprep needs to have corresponding tests created, and continually maintained. 

That being said, it's < 100 lines of code and is not complex! The main question I'm looking for brainstorming help on is, where do we add this feature? Maybe another button under world tools? Or another advanced featured hidden under the World Imports panel (e.g. under Animate Textures or Combine Textures)?

Then to make it an end-to-end feature, probably would want to have a filebrowser popup where you can select the output folder.

----
ps love your subtle promotion in the default description, I'll make a point to keep that if it stays in its current form ;)

pps hmmm maybe in the farther future, I start releasing custom MCprep panorama textures along with the version artwork of https://bit.ly/MCprepSplash... that could be a fun way to use the feature and help continually test it.

TheDuckCow:
> I'd personally put it in the Shift+A menu somewhere, but again, idk

I'm actually thinking of something more where it just uses the current position of the active camera, spawns in all the cameras, renders out the stuff, and then cleans up after itself (so the pano cameras don't remain after). This way it's truly an end to end experience, just pick your folder output and let it run. So aside from where the user clicks the button to start this custom render (maybe even from the render window??), does this user flow make sense? Would you use it like that?

Tiefseetauchner:
> Maybe another button under world tools? 

That was actually my first idea, but it kinda feels odd and might confuse people

[...]

PR #314, opened 2022-07-05
Adding title screen camera setup (#309)

Tiefseetauchner:
PR for #309 

Implemented main idea

# Known issues

1. When pressing "render panorama" and the file explorer icon in there to select the folder, the modal dialogue closes (ugh Blender), and when pressing accept, Blender has a segfault. Not even with a lot of info, just saying "Segmentation Fault (Core Dumped)" and the crashreport isn't helpful either:
```
# Blender 3.2.0, Commit date: 2022-06-08 10:22, Hash e05e1e369187

# backtrace
./blender(BLI_system_backtrace+0x20) [0xc1491e0]
./blender() [0x11d8e2a]
/lib/x86_64-linux-gnu/libc.so.6(+0x42520) [0x7fc07d01c520]
./blender(BLI_findstring+0x35) [0xc0a7535]
./blender(IDP_AddToGroup+0x15) [0x1209ed5]
./blender(RNA_property_string_set+0xe3) [0x1af1583]
./blender() [0x9d64fae]
./blender() [0x17a1621]
./blender() [0x17a47dc]
./blender() [0x17a513d]
./blender(wm_event_do_handlers+0x37a) [0x17a5b9a]
./blender(WM_main+0x20) [0x179aa30]
./blender(main+0x322) [0x10d47e2]
/lib/x86_64-linux-gnu/libc.so.6(+0x29d90) [0x7fc07d003d90]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x80) [0x7fc07d003e40]
./blender() [0x11d570c]

# Python backtrace
```
2. Currently, the Render window isn't being shown, and the program just freezes. That's because I'm currently using 'EXEC_DEFAULT' instead of 'INVOKE_DEFAULT'. The latter does not work, the script crashes saying when starting the second render. I shall investigate that later on (Maybe I'll have to put up signs in the neighborhood, my dog "window" from "context" is missing, have you seen them?).

```
  File "/home/richards/.config/blender/3.2/scripts/addons/MCprep_addon/world_tools.py", line 1097, in render_next_in_queue
    bpy.ops.render.render('INVOKE_DEFAULT', write_still=True, use_viewport=True)
  File "/home/richards/Documents/Blender/blender-3.2.0-linux-x64/3.2/scripts/modules/bpy/ops.py", line 113, in __call__
    ret = _op_call(self.idname_py(), C_dict, kw, C_exec, C_undo)
RuntimeError: Operator bpy.ops.render.render.poll() Missing 'window' in context
```

# Things to do better

1. The formatting is a mess, I come from C# and Javaland with spaces
2. The camera creation is currently hardcoded, due to it requiring certain angles. I could devise a smart formula with bit operations and magic, but I honestly think this is the more reasonable approach performance wise. But I shall everyone else be the judge of that.
3. render_queue_cleanup just exists because currently, I pop the elements from render_queue. Should I do that? I don't know. But I am. Maybe consider having an index instead
4. Both the cancel and finish callbacks call the function with three arguments. Self, scene, and... something. I don't know what it is. It's dummy to me now. Maybe find out what it is (bpy api doc is unhelpful as usual)
5. The default path shouldn't be /tmp, I'd rather it be populated with the current path set in the Output properties (also I just realised I don't buffer that yet, so that'll have to be changed)
6. Documentation
7. Tests
Need I say more
8. `# Do the renderage` is not a particularly helpful comment 
9. `World settings and lighting` is not the name of the category, I just forgot to commit that change, I'll amend it later 

Anyways, I think that's as bad as I can talk about my PR today, if someone happens to look at this, attempt not to look at the commit history, early versions are like gore, you don't wanna see it, you get ill, but you're drawn in by the horrors I produced (yes it was worse than the current one, no I have not read Clean Code, see https://i.redd.it/tdclszvedfc51.jpg)


TheDuckCow:
Thanks for sharing this @Tiefseetauchner and getting a solid start! I will do my best to take a review of what you have so far, but let me know if there's any specific areas you are seeking feedback around. Give me a day or two as I'm caught up with a few things right now.

I'll think about the things you mentioned in the top post about issues and to do better and see what suggestions I have.

MCprep_addon/world_tools.py
```diff
@@ -973,6 +974,126 @@ def world_time_update(self, context):
 	return
 
 
+class MCPREP_OT_render_panorama(bpy.types.Operator):
```
TheDuckCow, line 977: Probably the most helpful thing here is to switch this to also inherent from the `ImportHelper` class blender provides. This will change the invoke experience to be directly a file selection operation in a modal window. There is still a sidebar for additional properties in this modal, but most users probably won't see those so I wouldn't depend on it (but we can start there).

See how this is utilized with the swap texture pack button, same concept:
https://github.com/TheDuckCow/MCprep/blob/55719ce2b1faa3d007f6b2c69c7d0083ac26a90a/MCprep_addon/materials/prep.py#L342

(ignore the other class inherence McprepMaterialProps in this above example)

MCprep_addon/world_tools.py
```diff
@@ -973,6 +974,126 @@ def world_time_update(self, context):
...
+	"""Render the Panorama images for a texture Pack"""
+	bl_idname = "mcprep.render_panorama"
+	bl_label = "Render Panorama"
+	bl_description = "Render Panorama for texture Pack"
+	bl_options = {'REGISTER', 'UNDO'}
+
+	panorama_resolution = bpy.props.IntProperty(
+		name="Render resolution",
+		description="The resolution of the output images",
+		default=512
```
TheDuckCow, line 987: Does this value match the default for most resource packs/Vanilla?

MCprep_addon/world_tools.py
```diff
@@ -973,6 +974,126 @@ def world_time_update(self, context):
...
+	)
+
+	save_path = bpy.props.StringProperty(
```
TheDuckCow, line 990: Per [example here](https://github.com/TheDuckCow/MCprep/blob/55719ce2b1faa3d007f6b2c69c7d0083ac26a90a/MCprep_addon/materials/prep.py#L356) along with according blob filter settings, would be good to stay consistent in path namings, so I'd use `filepath` (even though it's really a directory).

MCprep_addon/world_tools.py
```diff
@@ -973,6 +974,126 @@ def world_time_update(self, context):
...
+		name="Output Folder",
+		description="The folder to output the renders to",
+		subtype="DIR_PATH",
+		default="/tmp\\"
```
TheDuckCow, line 994: User should be forced to select, so I'd just assign an empty string `""` and then if the folder doesn't already exist in the execute function, `return {'CANCELLED'}` after doing a `self.report({"ERROR"}, "Directory {x} does not exist")` for the user.

MCprep_addon/world_tools.py
```diff
@@ -973,6 +974,126 @@ def world_time_update(self, context):
...
+	render_queue = []
+	render_queue_cleanup = []
+
+	camera_data = None
+
+	old_res_x = None
+	old_res_y = None
+	original_cam = None
+
+	def invoke(self, context, event):
```
TheDuckCow, line 1006: Won't need to manually specify invoke once you inherent from ImportHelper above.

MCprep_addon/world_tools.py
```diff
@@ -973,6 +974,126 @@ def world_time_update(self, context):
...
+		return context.window_manager.invoke_props_dialog(self, width=400 * util.ui_scale())
+
+	def draw(self, context):
```
TheDuckCow, line 1009: This is good to keep though, remains on the sidebar, but again discovery may be low. Would only need the panorama resolution though, as the save_path (now filepath) becomes the main part of the modal.

MCprep_addon/world_tools.py
```diff
@@ -973,6 +974,126 @@ def world_time_update(self, context):
...
+		# Save old Values
+		self.original_cam = bpy.context.scene.camera
+		self.old_res_x = bpy.context.scene.render.resolution_x
+		self.old_res_y = bpy.context.scene.render.resolution_y
+
+		self.camera_data = bpy.data.cameras.new(name="panorama_cam")
+		self.camera_data.angle = math.pi / 2
+
+		pi_half = math.pi / 2
+		self.render_queue.append({
```
TheDuckCow, line 1023: Without reading too deeply or doing further analysis, I think your issue with crashing may be that these variables and references to self only exist within the context of this instance of the operator run. You probably want to define a global class that gets instances once and referenced through a global variable which holds any queue objects, so they can safely live outside the operator.

If you do it that way, some of these utility functions you define below could be pushed into this helper, non-operator class.

TheDuckCow (changes requested):
Left some initial comments. Start with what I mention here around the modal and ImportHelpter, and then I'm happy to dive a bit deeper around the point of the render queue. As I said, I know it's finicky, so if you can't quite get it to work after like a half hour or so, let me have at it and see what more we can do (outside of my one comment already suggesting one approach to use a new standalone, non-operator helper class).

MCprep_addon/world_tools.py
Tiefseetauchner: I feel stupid for not finding this, I searched for forever... Thank you, that pretty much fixes it. I'm a bit unhappy with the sidebar, but people will just render 1024 I guess, unless they know what they are looking for anyways

MCprep_addon/world_tools.py
Tiefseetauchner: Ok, what I gather from everywhere is that render() tries to get the active window with bpy.context.window (well, not really cuz it's C and not py but yk), and that window is somehow set to None at some point during/after the render. Which doesn't happen with EXEC_DEFAULT

So it would appear that I have to pass an [overriding context](https://docs.blender.org/api/current/bpy.ops.html#overriding-context) to the render function. 

But for some ungodly reason, ImportHelper seems to yeet the object reference at some point. 

```
ReferenceError: StructRNA of type MCPREP_OT_render_panorama has been removed
```

So now, instead of having the issue that the window doesn't exist anymore, I have the issue that my whole class doesn't exist anymore, which probably means that I'll have to refactor it into a global variable after all

TheDuckCow: Not sure if this helps, but when I encountered something related myself, I used an approach where for the _first_ render, I used invoke default, but then all following renders used exec default. You can see my [example here](https://github.com/TheDuckCow/blender-community-render/blob/main/community_render.py#L1336).

Tiefseetauchner: I tried this, after rendering 4 cameras, it crashes my blender. I have a very unstable configuration of Nvidia Drivers on Linux, that don't work reliably regardless, so I will try this as soon as I'm back on a stable (AMD) system. Thank you for the hint tho!

TheDuckCow:
Sorry for the radio silence on my side, it's been quite a past couple weeks. I'm making a point to launch the new MCprep update by Aug 1st (for real this time), and I hope we can include these changes! I'll get to code review and seeing about troubleshooting later today. Worst case scenario we don't sort it out in time, and it gets pushed to the next patch release (which can be soon after).

TheDuckCow:
Still looking through and will spend more time tomorrow seeing if there's a way we can better make it interactive (view intermediate renders and cancel part way through) - but that aside, I was able to use the operator and it worked well!

Screenshot of my title screen after making a render using my generic test scene export:

Screen Shot 2022-07-30 at 11 22 16 PM

MCprep_addon/world_tools.py
TheDuckCow: I was finding the same, so I skipped the approach. Just assed a new commit which uses a different approach of using timers to offset when the actual render starts, as otherwise I think the render complete callbacks essentially act recursively (render finishing calls a new callback before the end of the prior render finish callback).

TheDuckCow (approved):
Working well - I'm going to merge this later, feel free to take a look at the couple commit changes I made on your fork. If you have outstanding work or changes you want to make (not saying there are any), we can do that as another pull request.

Awesome contribution, I'm planning to showcase the new MCprep splash are with a pano render thanks to this!

Tiefseetauchner:
Coolio! Thanks for the merge!

I'll inquire on the Blender bug tracker whether the current behavior of "INVOKE_DEFAULT" is intended, and if there ends up being a prettier solution, I'll make a second PR!

Thanks very much!

PR #323, opened 2022-07-30
Added bug report template

StandingPadAnimations:


.github/ISSUE_TEMPLATE/Bug-Report.yml
```diff
@@ -0,0 +1,66 @@
...
+        - 3.3.0 
+        - 3.2.5.1
+        - 3.2.5
+    validations:
+      required: true
+  - type: dropdown
+    id: version
+    attributes:
+      label: Blender Version
+      description: What version of Blender are you using?
```
TheDuckCow, line 43: Is it possible to not hard code this? we will forever be out of date, rather just leave it a text entry field and give a sample format of entry. We might also want to allow something like "n/a" if it's a feature request (though that would be a different template I suppose).

.github/ISSUE_TEMPLATE/Bug-Report.yml
```diff
@@ -0,0 +1,66 @@
...
+        value: "Step 1: "
+      validations:
+        required: true
+  - type: dropdown
+    id: version
+    attributes:
+      label: MCprep Version
+      description: What version of MCprep are you using?
+      options:
+        - 3.3.1 (Default)
```
TheDuckCow, line 33: I'm also thinking maybe we should make this not be an explicit list of options too since people can always look up their version, if it's required they'll enter it.

.github/ISSUE_TEMPLATE/Bug-Report.yml
```diff
@@ -0,0 +1,66 @@
...
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What happened?
+      description: Also tell us, what did you expect to happen?
+      placeholder: Tell us what you see!
+      value: "A bug happened!"
+    validations:
+      required: true
+    - type: textarea
```
TheDuckCow, line 19: Can we add one more question which is "Did the issue continue after you restarted blender?"
and they have to select Yes or no (but if they select no, then they really shouldn't continue filling out the form).

So many bug reports have been resolved due to this simple thing, often after someone did an update but didn't restart blender.

TheDuckCow (approved):
Minor changes requested, but looking good! Approving understanding you'll make the adjustments, this is a rare case where I assume it indeed is necessary to commit directly to master right?

StandingPadAnimations:
I've made all the changes, tested the issue template, and will now merge

PR #334, opened 2022-08-16
Optimizer Improvements for 3.4.2

StandingPadAnimations:
This PR shall track the progress of optimizer improvements for MCprep 3.4.2.

# New UI
The current UI for the optimizer is a bit hard to understand right now. First off, there are certain boolean properties that could be better as enum properties. Take for instance the Caustics setting. When set to true, caustic optimization is disabled, but the user might think it means to enable caustics optimization.

That's just a small example of a UI issue in the optimizer, so the UI will need to be slightly reworked

- [x] Turn confusing boolean properties into enum properties
- [x] Figure out better names for settings

# ~~New Feature: Optimizer JSON Settings~~ JSON Settings shall be replaced with the new Advanced Settings panel
~~This aims to add some internal settings of the MCprep optimizer to a JSON file that can be modified by the user to do some fine-tuning. Some of the settings that will be included (along with their default values) are:~~
```json
{
    "MAX_BOUNCES": 8,
    "MIN_BOUNCES": 2,
    "MAX_FILTER_GLOSSY": 1.0,
    "MAX_STEPS": 200,
    "MIN_SCRAMBLING_MULTIPLIER": 0.35,
    "SCRAMBLING_MULTIPLIER_ADD": 0.05,
    "MCPREP_HOMOGENOUS_VOLUME" : "MCPREP_HOMOGENOUS_VOLUME",
    "MCPREP_NOT_HOMOGENOUS_VOLUME" : "MCPREP_NOT_HOMOGENOUS_VOLUME"
}
```

~~This allows the user to edit stuff such as the maximum and minimum amount of bounces, the max volume stepping size, and even the names for the optimizer to use as hints.~~

~~Now this is not the same as going into the code and adjusting everything towards your own hardware (not by a long shot, that's not what the MCprep optimizer was designed for), but it does give the user more control. As new versions come out, more internal things will be exposed in this basic config file.~~

- [x] Load JSON settings to internal variables
- [x] Allow new scene brightness presets

# Redundant Optimization for Blender 3.0 and Above
The optimizer currently uses tricks that are common in older versions of Blender. While most are still useful, some aren't. In particular, disabling caustics seems to not do much in terms of performance.

This is actually a good thing, as it shows the improvements that have been made to Cycles. However, these redundant optimizations may actually cause some issues. For examples, here are some comparison images of a glass like material with and without caustics enabled (here it's mostly a brightness and vibrancy difference in the right eye):
Without Caustics             |  With Caustics
:-------------------------:|:-------------------------:
image |  image

While this may seem like a small difference, it's not hard to imagine a scene with a lot of reflective/refractive materials looking completely different. 

A combination of a version check and a backend check (MCprep already checks the backend used for rendering anyway) to set different defaults should be sufficient.

- [x] Create a system for MCprep to check the backend at the UI level (currently we can only check at the operator level, so we may need some “caching” system)
- [x] Determine which settings are redundant in Blender 3.0+ 

# Bounce Optimization
Right now, MCprep optimizes 2 types of bounces:
* Glossy
* Transmissive

However, there are a few issues. Firstly, I feel that modifying transmissive bounces is useless for MCprep materials and could instead be replaced with transparent bounce optimization (which would solve a common issue MCprep users have). Secondly, sometimes a scene may need fewer bounces, and sometimes a scene may need more. While the new config file will help a lot in that regard, it's on a general basis and not a scene by scene basis.

- [x]  Figure out a proper way to set transparent bounces 
- [x] Add some settings to allow further fine-tuning at the scene level


StandingPadAnimations:
Whoops, accidentally set the base branch to master, so there may be some missing things (though I've fixed the PR to use the dev branch)

StandingPadAnimations:
So far, disabling caustics by default seem to be the only redundant setting. That should be an easy fix.

StandingPadAnimations:
For scene level fine tuning, I have an idea for another panel that's by default hidden (only visible if you enable it in preferences)

This would require some way to store data across sessions though

StandingPadAnimations:
Actually ignore that above comment, property groups already save properties on a per-file basis

TheDuckCow:
Just to check, still looking for me to review now?

StandingPadAnimations:
> Just to check, still looking for me to review now?
Yep, just need a preliminary review

StandingPadAnimations:
While this PR is mainly optimizer improvements, I have pushed some commits to make MCprep use Alpha Blend for the blend mode and None for the shadow mode as Alpha Hash is too noisy. 

Let me know if this needs to be changed back @TheDuckCow

StandingPadAnimations:
Also, JSON settings will be replaced with the new Advanced panel, which I feel is easier to reach and more intuitive.

StandingPadAnimations:
All advanced settings should be finished now

StandingPadAnimations:
For the most part, I think this PR is done. I do plan on adding a feature called Easy DoF but that should be its own PR

@TheDuckCow, ready for review

StandingPadAnimations:
The only feature missing from this PR is scene brightness presets, and I don't think I'll add them anytime soon since the advanced panel should be enough for most scenes

StandingPadAnimations:
> For the most part, I think this PR is done. I do plan on adding a feature called Easy DoF but that should be its own PR
> 
> @TheDuckCow, ready for review

I spoke too soon, I still need to add a feature that disables mesh lights (it wasn't part of the original plan so I forgot)

StandingPadAnimations:
Alright, I've now added the “Remove Emission” feature and began work on the “Remove Useless Nodes” feature

I think I'll start moving some features of the optimizer from Cycles to EEVEE (mainly the 2 new features) as they would have benefit in shader compilation times for SEUS and Specular materials

StandingPadAnimations:
Here's a preview of what I'm working on
https://user-images.githubusercontent.com/75058058/187323933-5a728d3d-c7e2-4f2c-9ae6-6966f21028c7.mp4

StandingPadAnimations:
Alright, here are the main new features added this week:
* Removing emission nodes
* Removing useless nodes (for Specular and SEUS)
* Adding a feature where if the user accidentally clicks an MTL file, the corresponding OBJ file will be selected instead

@TheDuckCow what are your thoughts?

MCprep_addon/util.py
```diff
@@ -202,6 +201,17 @@ def bv30():
 	"""Check if we're dealing with Blender 3.0"""
 	return min_bv((3, 00))
 
+BLEND_MODE = "HASHED"
+SHADOW_MODE = "HASHED"
+def change_blend(self, context):
```
TheDuckCow, line 206: Thanks for making this change and, as discussed, making hashed default.

MCprep_addon/util_operators.py
```diff
@@ -86,6 +86,10 @@ def execute(self, context):
 
 			# make it solid mode, regardless of cycles or eevee
 			context.scene.display.shading.type = 'SOLID'
+
+		# TODO: Confirm this works across all versions of Blender
```
TheDuckCow, line 90: Just confirmed this works down to at least blender 2.79 (even if Blender Render engine is active; blender 2.78 is probably also fine with it)

MCprep_addon/conf.py
```diff
@@ -70,6 +70,12 @@ def init():
 		"MCprep_resources",
 		"mcprep_data_update.json")
 
+	global optimizer_json_path
```
TheDuckCow, line 73: Where does this get used? I recall our discussion of this being used as an override, but I'm not seeing where this gets picked up in the code.

MCprep_addon/materials/generate.py
```diff
@@ -1522,16 +1529,19 @@ def matgen_cycles_principled(
...
+				mat.shadow_method = util.SHADOW_MODE
 
 			# best if there is no partial transparency
 			# material.blend_method = 'CLIP' for no partial transparency
 			# both work fine with depth of field.
 
 			# but, BLEND does NOT work well with Depth of Field or layering
 
+			# This prevents transparency issues with Alpha Blend
+			mat.show_transparent_back = False
```
TheDuckCow, line 1543: Ah, what happens here? Do you have a screenshot of with this off vs on in a relevant context? Just curious.

MCprep_addon/materials/prep.py
```diff
@@ -119,6 +124,12 @@ def pack_formats(self, context):
 		description="Change the pack format when using a PBR resource pack.",
 		items=pack_formats
 	)
+	blendMode = bpy.props.EnumProperty(
```
TheDuckCow, line 127: Per further above, does this completely replace the need for having the config file for your optimizer settings? (or are you planning to do more with that? either way is fine, just would rather not merge in code that "might get used" as it's hard to tell what's safe to delete at any given moment in time)

MCprep_addon/world_tools.py
```diff
@@ -271,15 +272,16 @@ class MCPREP_OT_import_world_split(bpy.types.Operator, ImportHelper):
 	@tracking.report_error
 	def execute(self, context):
 		# for consistency with the built in one, only import the active path
+		if self.filepath.lower().endswith(".mtl"):
+			filename = Path(self.filepath)
+			new_filename = filename.with_suffix(".obj")
+			self.filepath = str(new_filename) # Change it from MTL to OBJ, this will be checked with the rest of the if clauses
```
TheDuckCow, line 278: Nit: move the # Change it.. comment to the line above, to avoid exceeding the 79 character line

MCprep_addon/optimize_scene.py
```diff
@@ -26,58 +26,77 @@
 MAX_BOUNCES = 8 # 8 is generally the standard for light bounces
```
TheDuckCow, line 26: File-level comment:

What are some basic tests we can run and add to the test runner, to make sure this is working across versions? This is probably a decent time to add something even basic, so I can validate it continues to work across versions.

TheDuckCow (changes requested):
Sorry I've been slammed recently @StandingPadAnimations. taking a high level look now and replying:

- I'm not sure about the emission nodes thing. Is that a bandaide around a lower level issue? Should we just have an option to prep materials with emission materials off?
- Removing "useless" nodes I'm supportive of.
- Great improvement to auto-select the OBJ!

Left some other comments, nothing major.

MCprep_addon/conf.py
StandingPadAnimations: Whoops, that was back when I was thinking of adding JSON configs to the optimizer (which now uses the new advanced panel). Thanks for catching that

MCprep_addon/materials/generate.py
StandingPadAnimations: Right now I don't have a screenshot on hand but without disabling this, any materials using alpha transparency (when using Alpha Blend) will be really messed up. This option does not affect Alpha Hashed.

MCprep_addon/materials/prep.py
StandingPadAnimations: JSON config was replaced with an advanced panel that only shows up if the user enables it in user preferences. Optimizer configuration allows the user to find tune and control the internal parts of the optimizer (such as max and minimum bounces, volumetric stepping rate, etc.)

StandingPadAnimations:
I think as part of this update I'll port some optimizer features for EEVEE (or at least ones that are compatible in both Cycles and EEVEE, like simplify)

MCprep_addon/materials/prep.py
TheDuckCow: Ok I think that's ok then, I was sort of thinking having a config file for a specific feature felt a little excessive and not that user friendly anyways.

MCprep_addon/materials/generate.py
TheDuckCow: Great, good to know it doesn't affect alpha hashed, thanks!

MCprep_addon/conf.py
TheDuckCow: Sure thing! Reminder to delete when you get to updating this next :)

MCprep_addon/util_operators.py
TheDuckCow: (We can remove this comment since this has been confirmed now)

MCprep_addon/optimize_scene.py
TheDuckCow: Reminder to think of this, I'd like to have at least one unit test so we can have confidence when testing new (and old) blender versions accordingly.

StandingPadAnimations:
Blender 3.4 brought path guiding to Cycles. More research will need to be done on how it impacts scenes that use MCprep

StandingPadAnimations:
Some details on this PR since it's been a while since any work has been done on this, the following goals are in progress:
- [x] Finer adjustment of how the optimizer tweaks values
- [x] Leaving caustics alone in Blender 3.0
- [ ] Porting some settings to EEVEE
- [ ] Creating a test file for MCprep tests
- [ ] Fixing bugs related to volume settings not being applied
- [ ] [**MAYBE**] A way for users to send their own files through email to allow improvements

In addition, the following improvements are being changed:
*  Removal of mesh lights (being moved to material generation and would require a separate PR IMO)
* Clamping indirect as it's too destructive in certain scenes. Implementing this properly would require some external Python libraries though, and would likely be put in its own PR
	* Improved bounce optimization would require the same things as clamping indirect. Sadly I'm not sure how to optimize scenes further
	* Perhaps a hidden setting in user preferences that requires running Blender with elevated permissions
		* Alternatively we could use some file path tricks [like mentioned in this Stack Overflow post](https://stackoverflow.com/questions/11161901/how-to-install-python-modules-in-blender/11488685#11488685), but I do not feel comfortable in creating new files in random locations on the user's system. In addition, I feel like this may open a security hole, but we could assume the user takes liability for enabling the feature and make it easy to disable and remove the modules.

I also feel that setting this as a deadline for 3.4.2 is a bit limiting, so I'll just mark this as "General Optimizer Improvements" and not keep MCprep 3.4.2 dragged down by this PR

StandingPadAnimations:
As mentioned [here](https://standingpadanimations.github.io/posts/goodbye-for-now/), I won't be working on MCprep anymore, so this PR will just be left in this state.

I believe this will eventually get done in the future, either by the open source community or myself (if I decide to come back) but for now I'll just set this as a draft

StandingPadAnimations:
Closing this in favor of using future improvements made in [Kaion](https://github.com/StandingPadAnimations/MCprep-Kaion)

PR #339, opened 2022-08-25
add effects to user preferences and advanced option

zNightlord:
For certain user want to change the effect path to another location through preferences and through the advanced dropdown and resolve the symlink issue, user created or when the addon added through vscode debug extension. https://github.com/TheDuckCow/MCprep/issues/338

I'm not sure about adding reset path for consistency.


TheDuckCow (approved):
This looks great, thanks for the contribution! (and apologies for the delayed reply)

TheDuckCow:
Merging this, thanks again

PR #348, opened 2022-10-16
Internal Refactoring for 3.4.3

StandingPadAnimations:
# List of Deprecated Features Planned for MCprep 3.4.3
## `conf.dev` -> `conf.env.dev_build`
```py
from .conf import env
env.dev_build
```

## `conf.v` -> `conf.env.verbose`
```py
from .conf import env
env.verbose
```

## `conf.vv` -> `conf.env.very_verbose`
```py
from .conf import env
env.very_verbose
```

## `conf.data` -> `conf.env.data`
```py
from .conf import env
env.data
```

## `conf.json_data` -> `conf.env.json_data`
```py
from .conf import env
env.json_data
```

## `conf.json_path` -> `conf.env.json_path`
```py
from .conf import env
env.json_path
```

## `conf.use_icons` -> `conf.env.use_icons`
```py
from .conf import env
env.use_icons
```

## `conf.preview_collections` -> `conf.env.preview_collections`
```py
from .conf import env
env.preview_collections
```

## `conf.icons_init()` -> `conf.env.icons_init()`
```py
from .conf import env
env.icons_init()
```

## `conf.loaded_all_spawners` -> `conf.env.loaded_all_spawners`
```py
from .conf import env
env.loaded_all_spawners
````

## `conf.skin_list` -> `conf.env.skin_list`
```py
from .conf import env
env.skin_list 
```

## `conf.rig_categories` -> `conf.env.rig_categories`
```py
from .conf import env
env.rig_categories 
```

## `conf.entity_list` -> `conf.env.entity_list`
```py
from .conf import env
env.entity_list
```

## `conf.material_sync_cache` -> `conf.env.material_sync_cache`
```py
from .conf import env
env.material_sync_cache
```

## `conf.register()` ->`conf.MCprepEnv.__init__(self)`
```py
# conf.py
env = MCprepEnv()
```

## `conf.log(statement, vv_only=False)` -> `conf.env.log(self, statement, vv_only=False)`
```py
from .conf import env
env.log("Debug statement!")
```

PR #357, opened 2022-12-07
Using new mix node in Blender 3.4

zNightlord:
This adds `create_node()` to helps with node generation between version with different socket index

### DIFF
--- a/MCprep_addon/materials/generate.py
+++ b/MCprep_addon/materials/generate.py
@@ -1090,7 +1090,7 @@ def texgen_specular(mat, passes, nodeInputs, use_reflections):
 
 	nodeSpecInv = create_node(nodes, "ShaderNodeInvert", name = "Specular Inverse", label = "Specular Inverse", location = (-80, -280))
 	nodeNormalInv = create_node(nodes, "ShaderNodeRGBCurve", name = "Normal Inverse", label = "Normal Inverse", location = (-380, -500))
-	nodeNormal = create_node(nodes, "ShaderNodeNormalMap", location = (-80, 140))
+	nodeNormal = create_node(nodes, "ShaderNodeNormalMap", location = (-80, -500))
 
 	# Sets values
 	nodeSaturateMix.inputs[0].default_value = 1.0 # Graystyle Blending
--- a/MCprep_addon/materials/generate.py
+++ b/MCprep_addon/materials/generate.py
@@ -964,6 +964,7 @@ def create_node(tree_nodes, node_type, **attrs):
 		tree_nodes: the material node tree's nodes
 		node_type: the type of the node
 		**attrs: set attributes if that node type has (eg: location, name, blend_type...)
+		"hide_sockets" to hide the sockets only display linked when need
 	"""
 	if node_type == 'ShaderNodeMixRGB': # MixRGB in 3.4
 		if util.min_bv((3, 4, 0)):
@@ -976,6 +977,9 @@ def create_node(tree_nodes, node_type, **attrs):
 	for attr, value in attrs.items():
 		if hasattr(node, attr):
 			setattr(node, attr, value)
+		elif attr == 'hide_sockets': # option to hide socket for big node 
+			node.inputs.foreach_set('hide', [value] * len(node.inputs))
+			node.outputs.foreach_set('hide', [value] * len(node.outputs))
 	return node
 
 def get_node_socket(node, is_input=True):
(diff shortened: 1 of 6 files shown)


StandingPadAnimations:
@zNightlord can this be merged into the dev branch? Master is what we use for releases

StandingPadAnimations:
I'll do further review later this week, but through my quick skim of the code here's what I did notice the use of `bpy.app.version`. MCprep provides a utility function called `minBv` which essentially does the same thing, just more readable and takes into account version differences in the API

StandingPadAnimations:
Just requesting a review from @.TheDuckCow for this pull request, though I'll also be reviewing the code and noting where changes have to be made to account for #348

MCprep_addon/materials/generate.py
```diff
@@ -958,6 +958,27 @@ def set_saturation_material(mat):
...
+			inputs = [0,6,7]
+			outputs = [2]
+		else:
+			node = tree_nodes.new('ShaderNodeMixRGB')
+			inputs = [0,1,2]
+			outputs = [0]
+		sockets= {
+			"inputs": inputs,
+			"outputs": outputs
+		}
```
TheDuckCow, line 976: I need to live test a little, but this feels a bit funny to me. I think it's important that a function generally has the same return structure regardless of how it runs (here, sometimes it returns one value, sometimes two, which makes it unclear what you get back down the line).

I'll probably propose an edit later today unless you beat me to it.

Did the socket _names_ change, or just the order? If the names we need to use are the same

TheDuckCow (changes requested):
Marking to request changes, but I recognize I'm not being very explicit yet as I need to take a look myself. The main comment is below, about having consistent return times. I don't love the idea of having to have every single node return both its reference _and_ a sockets function just due to this one change, but the higher-level point for me is that functions have a consistent return structure regardless of what happens inside (ie number of return vars)

Regardless, thanks for jumping so quick on this!

MCprep_addon/materials/generate.py
zNightlord: For the sockets part, I understand. It's quite odd for me sometimes returns one, two.
The new Mix changes the socket index and the name. If you use the socket name, you will get the result of a float mix instead of RGBA. 
image

StandingPadAnimations: This may be the C++/Rust part of my brain acting up, but couldn't we just either:
A. Always return a tuple regardless of the amount of objects, where the first item is the node and the second object depending on the version can be the sockets 
B. Use a small wrapper class to handle this situation

Knowing how CPython works though, the C++/Rust part of my brain is screaming "unnecessary copy!" as I write this

zNightlord: Maybe we could split the sockets part out to a different function like `get_node_socket("ShaderMixNode", "outputs",1)` 
It will do versioning for the socket of that node type seperate.

zNightlord:
Oops.

StandingPadAnimations:
Alright I'll start my review (sorry for the delay, I was installing Arch on my system)

StandingPadAnimations:
Note that anything I mark as `NOTE_FOR_INTERNAL_REFACTORING` is something I'll deal with as part of #348

MCprep_addon/materials/generate.py
```diff
@@ -958,6 +958,32 @@ def set_saturation_material(mat):
 		sat_node.mute = not bool(saturate)
 		sat_node.hide = not bool(saturate)
 
+def create_node(tree_nodes, node_type, **kwargs):
+	"""Function to create node"""
+	if node_type == 'ShaderNodeMixRGB': # For MixRGB in 3.4 become Legacy 
+		if util.min_bv((3, 4, 0)):
```
StandingPadAnimations, line 964: `NOTE_FOR_INTERNAL_REFACTORING` (sorta)

There are plans for `MCprepEnv` to check for 2.8/3.0. 3.4 is still debatable as that brought a couple of big changes that make MCprep have to do workarounds, but at the same time it's not as big as say 2.8 or 3.0 were

MCprep_addon/materials/generate.py
```diff
@@ -958,6 +958,32 @@ def set_saturation_material(mat):
...
+		node = tree_nodes.new(node_type) 
+	for attr, value in kwargs.items():
+		if hasattr(node, attr):
+			setattr(node, attr, value)
+	return node
+	
+def get_sockets_node(node, is_input=True):
+	n_type = node.bl_idname
+	if n_type == 'ShaderNodeMix' or n_type == 'ShaderNodeMixRGB':
+		if util.min_bv((3, 4, 0)): # For MixRGB in 3.4 become Legacy
```
StandingPadAnimations, line 979: `NOTE_FOR_INTERNAL_REFACTORING` (sorta)

Same as my above comment

MCprep_addon/materials/generate.py
```diff
@@ -1191,6 +1219,8 @@ def texgen_seus(mat, passes, nodeInputs, use_reflections):
 	nodeNormalInv.mapping.curves[1].points[0].location = (0, 1)
 	nodeNormalInv.mapping.curves[1].points[1].location = (1, 0)
 
+	saturateMixIn =  get_sockets_node(nodeSaturateMix) #sockets["inputs"]
```
StandingPadAnimations, line 1222: Love the comments, keep them up

MCprep_addon/materials/generate.py
```diff
@@ -1571,31 +1603,37 @@ def matgen_cycles_original(
...
-	nodeFalloff = nodes.new("ShaderNodeLightFalloff")
-	nodeLightPath = nodes.new("ShaderNodeLightPath")
-	nodeEmit = nodes.new("ShaderNodeEmission")
-	nodeEmitCam = nodes.new("ShaderNodeEmission")
-	nodeMixCam = nodes.new("ShaderNodeMixShader")
-	nodeMixEmit = nodes.new("ShaderNodeMixShader")
-	nodeTrans = nodes.new("ShaderNodeBsdfTransparent")
-	nodeMixTrans = nodes.new("ShaderNodeMixShader")
-	nodeOut = nodes.new("ShaderNodeOutputMaterial")
+	# Shader
```
StandingPadAnimations, line 1606: Again, good job at commenting these sections. Makes it easier to maintain

MCprep_addon/materials/generate.py
```diff
@@ -958,6 +958,32 @@ def set_saturation_material(mat):
...
+		if hasattr(node, attr):
+			setattr(node, attr, value)
+	return node
+	
+def get_sockets_node(node, is_input=True):
+	n_type = node.bl_idname
+	if n_type == 'ShaderNodeMix' or n_type == 'ShaderNodeMixRGB':
+		if util.min_bv((3, 4, 0)): # For MixRGB in 3.4 become Legacy 
+			if node.data_type == 'RGBA': 
+				inputs, outputs = [0,6,7], [2]
```
StandingPadAnimations, line 981: I'm not sure where these numbers come from. Perhaps add a comment explaining where the numbers come from?

StandingPadAnimations (commented):
Overall, some good things. There are a couple of sections that may need to be changed for internal refactoring, depending on what we decide for 3.4 version detection

TheDuckCow:
Sorry for the long delay getting back to you - I'll try to finish my review towards the end of this weekend, and then hopefully we can just merge it and address any follow ups ad hoc so we don't lose track. Thanks again for going through the iterations here!

zNightlord:
> Sorry for the long delay getting back to you - I'll try to finish my review towards the end of this weekend, and then hopefully we can just merge it and address any follow ups ad hoc so we don't lose track. Thanks again for going through the iterations here!

No problem. 
I just add in some small adjustment for nodegroup from the node group test

TheDuckCow:
FYI I'm almost but not quite done with some local updates I'm going to push onto the branch before approving/merging, so please avoid making any changes until I get that in (just didn't quite get it done tonight).

TheDuckCow:
Hey @zNightlord can you add me as an editor to your [fork here](https://github.com/zNightlord/MCprep/)? (settings deeplink: https://github.com/zNightlord/MCprep/settings/access). Then I can push the commit I've got locally. On that, I'll approve and we can merge it in!

TheDuckCow (approved):
Approved.

I was quite hesitant about having the *attrs argument for the node function. But in the end, it does provide upside, and from a code completion standpoint, we don't have static typing really with the node objects being passed around anyways. 

All in, thanks for making this change, glad to have it merged in 👍  (oh and local testing went well enough, I'll do some more testing in older blender versions too; but all automated tests through 2.80+ work fine).

zNightlord:
Alright I don't have write access. So you can merge it now

PR #369, opened 2023-01-31
Option to remove emission from prep materials

StandingPadAnimations:
Changed the branch so reopening

Work on this feature began in [Kaion](https://github.com/StandingPadAnimations/MCprep-Kaion/tree/dev?rgh-link-date=2023-01-23T01%3A40%3A53Z), and now I'm just refining this feature.

This adds an option to prep materials to not add emissive nodes, which can help to reduce noise in Cycles and makes nodes less complex in EEVEE.

PR #370, opened 2023-01-31
Moon lamp

StandingPadAnimations:
This adds a minor change where soft shadows is now enabled for EEVEE when using the MC sky (though this should be merged after #369 since this include the commits from that)


TheDuckCow:
For when I get to testing (sorry it's taken so long), I need to use the files here:
https://github.com/TheDuckCow/MCprep/issues/112#issuecomment-1410919329

StandingPadAnimations:
> I get some weird popping of the sun at 4am and 5am for eevee and cycles respectively with these blends, is there a way we could smooth the function so it fades out?

In Cycles, we could use a map range node, but not in EEVEE

> Also for some reason, the moon wasn't appearing.. and worse yet, blender 3.5, I'm not getting the sun or moon lamps moving with the rotating empty, haven't dug deeper but something is further off there.
Strange, seems to be working on my end

StandingPadAnimations:
Wait, what the f- (excuse my language but this thing I found in the Blender docs is amazing)
image

This may require some extra code, but I think this will make our lives easier

StandingPadAnimations:
@TheDuckCow I should mention that this doesn't work well, but the general gist is there:
```py
def mcprep_sun_ease(world_time):
    energy = 0
    normalized_world_time = world_time % 24
    if normalized_world_time <= 18 and normalized_world_time >= 6:
        energy += 0.5 * normalized_world_time if energy != 10 else 0
    else:
        energy -= 0.5 * normalized_world_time % 12 if energy != 0 else 0
    return energy 

def mcprep_moon_ease(world_time):
    normalized_world_time = world_time % 24
    energy = 5 if normalized_world_time <= 6 else 0
    if normalized_world_time >= 18 or normalized_world_time <= 6:
        energy += 0.25 * normalized_world_time % 12 if energy != 2 else 0
    else:
        energy -= 0.25 * normalized_world_time if energy != 0 else 0
    return energy 


import bpy

# Add functions defined in this script into the drivers namespace.
bpy.app.driver_namespace["mcprep_sun_ease"] = mcprep_sun_ease
bpy.app.driver_namespace["mcprep_moon_ease"] = mcprep_moon_ease
```

Most notably it isn't smooth and doesn't clamp properly, but the main time points 6 and 18 (when the sun and moon are equal) are the main markers for when to start increasing/decreasing

StandingPadAnimations:
It is weird when making a change though; you have to click the brightness values of both the sun and moon each time you reregister it

TheDuckCow:
Wow never knew you could make functions available like that to python, very interesting indeed.

Appreciate the investigation. I'm thinking I'll still hold off on this for the 3.4.2 release, but will aim to better test and understand what's going on so it can be in the release first thing after. The main changes I needed for the upcoming release are both in, so I want to just do final tidying up.

StandingPadAnimations:
> Wow never knew you could make functions available like that to python, very interesting indeed.

Yeah, although Blender will complain the expression is "slow". Since it's in the Blender docs and even a template in the Blender editor (`Templates > Python > Driver Functions`), we shouldn't have to worry about Blender changing something in the future (and hopefully should work in older versions).

>Appreciate the investigation. I'm thinking I'll still hold off on this for the 3.4.2 release, but will aim to better test and understand what's going on so it can be in the release first thing after.

Yeah I think we should wait, since determine the correct values to increase and decrease by will take a while (not to mention making everything smooth and not janky like in the snippet I posted)

StandingPadAnimations:
Ok so after some thinking, I figured it would actually be easier to use quadratics in this case (you know, `ax^2 + bx + c`). We would need 1 for the sun cycle, and 2 for the moon cycle (since midnight is 0)

StandingPadAnimations:
Ok, so after doing 3 separate rounds of quadratic regression, we get the following (rounded to 3 decimal places):
$scy(x) = -0.073x^2+1.762x-5.161$
$mfh(x) = -0.014x^2+0.831x-10.166$
$mlh(x) = -0.012x^2-0.146x+1.590$

$scy(x)$ being the sun cycle, $mfh(x)$ being the first half of the moon cycle (since midnight is 0), and $mlh(x)$ being the latter half of the moon cycle.

Now assuming I didn't make a mistake typing all that in, our Python functions for the drivers would look like this:
```py
def mcprep_sun_ease(world_time):
    x = world_time % 24
    return (-0.0073)(x**2)+(1.762)(x)-(5.161)

def mcprep_moon_ease(world_time):
    x = world_time % 24
    if x >= 0:
        return (-0.012)(x**2)-(0.164)(x)+(1.590)
    else:
        return (-0.014)(x**2)+(0.831)(x)-(10.166)


import bpy

# Add functions defined in this script into the drivers namespace.
bpy.app.driver_namespace["mcprep_sun_ease"] = mcprep_sun_ease
bpy.app.driver_namespace["mcprep_moon_ease"] = mcprep_moon_ease
```

Much simpler, and less likely to have an error (unless a typo occurs or Python decides to not follow basic algebra)

StandingPadAnimations:
Here are the updated files:
[Cycles](https://cdn.discordapp.com/attachments/741468426404495410/1084521702026903652/clouds_moon_sun.blend)
[EEVEE](https://cdn.discordapp.com/attachments/741468426404495410/1084521044657840139/clouds_moon_sun_eevee.blend)

MCprep_addon/world_tools.py
```diff
@@ -1206,6 +1206,19 @@ def execute(self, context):
...
+	
+	# We want to only return this value if it's above 2, otherwise ignore it
+	return f_x if f_x >= 2 else 0
+
+def mcprep_moon_ease(world_time):
+    x = world_time % 24
+    if x >= 18 and x <= 23.999:
+        return (-0.014)*(x**2)+(0.831)*(x)-(10.166)
+    else:
+        return (-0.012)*(x**2)-(0.164)*(x)+(1.590)
```
TheDuckCow, line 1221: So I'm realizing now, one issue with this approach of defining drivers, is that it will not stand on its own. ie if sent to another computer to render, if it doesn't have MCprep installed, it will fail to work.

Something for us to think about further. We could potentially have this as a script file which is actually generated into a text block that lives with the blend file for instance, but I do want to have a chance to play around with this more first before you go down this path.

StandingPadAnimations:
Alright, I've went ahead and calculated a new quadratic for the sun cycle. Specifically, I also added that inputting 5 and 19 should be 0, which returns the following:
$scy(x)=-0.105x^2+2.511x-9.265$

Now this is approximate, and my calculator is also saying $R^2=0.91$, which is not ideal, but better then nothing. More importantly, $scy(5)$ and $scy(19)$ will be less then 1, which allows disabling the sun fully to be easier.

As such, the driver function for the sun should now be:
```py
def mcprep_sun_ease(world_time):
    x = world_time % 24
    f_x = (-0.105)*(x**2)+(2.511)*(x)-(9.265)
    if f_x < 1:
        return 0
    return f_x
```

MCprep_addon/world_tools.py
StandingPadAnimations: > Something for us to think about further. We could potentially have this as a script file which is actually generated into a text block that lives with the blend file for instance

We could also append the text file from the file since it's included

StandingPadAnimations:
I've went ahead and added documentation on how this works: https://github.com/StandingPadAnimations/MCprep-Kaion/blob/moon-lamp/docs/implementation.md#day-night-cycle-implementation

StandingPadAnimations:
Wait what am I doing, we have angle as well

That would simplify it down to 2 quadratics

StandingPadAnimations:
Alright, so I haven't figured out a way to use angle for quadratics, but I did ask a math nerd and they said it would be simpler to just invert one of the quadratics

StandingPadAnimations:
After some attempts at inverting, I think it's simpler to just use 3 equations and switch with a if statement

TheDuckCow:
Sorry that we had to let this one sit for so long - are those blend files still the most up to date ones? Anything else pending from your side? If not, I'll re-assign myself for review.

Will this solution still require having text py files packaged as well (so we don't conflict with things like render farms)?

StandingPadAnimations:
Blend files are up to date. I still need to add embedding driver functions to the file

TheDuckCow:
Embedded drive in addition to, or instead of, a script that needs to be attached to the blend file to stand on its own [without MCprep installed afterwards]? Hopefully we can limit the dependency as much as possible. Ideal solution would still just be a driver on its own with no extra script necessary, but I realize it's tricky.

StandingPadAnimations:
Embedded into the file and set to auto-register to make blend files portable

TheDuckCow:
Ok, let's move this PR to draft until we have that aspect integrated.

StandingPadAnimations:
@TheDuckCow finally added driver function embedding in 054bb6c

StandingPadAnimations:
I'll also be moving the base to `milestone-3-5-0`, which includes updating with that branch

StandingPadAnimations:
Alright I've updated everything and verified it works

TheDuckCow:
Would be good to rebase one more time (or merge in, whatever your fancy) the `milestone-3-5-0` branch, now that we merged the annotations branch (and separately check that detail of it not actually importing the moon).

TheDuckCow (changes requested):
As discussed, StandingPad checking why the moon doesn't import. Marking as request changes on that basis, will re-review when there are updates (and the rebase is done again, so I can live test it in 2.80 where there might be some incompatibility).

StandingPadAnimations:
Ok looking into it the sun isn't even appended properly for Cycles; in Cycles the sun lamp has a black body node setup that's missing when the MC sky is created

StandingPadAnimations:
Ok from what I can tell, we're not importing the sun and moon explicitly, we're depending on Blender to implicitly import them (the sky depends on a property from the time object, which would implicitly also import objects it has a relationship with)

PR #373, opened 2023-02-15
Using the C++ OBJ importer and Fixing Issues with Importing when using a Non-standard colorspace

StandingPadAnimations:
This comes at the cost of having a split OBJ, but it solves 2 issues:
* Performance
* Errors involving colorspaces like ACES and AgX

Perhaps we could put this behind a user prefs option, so users can use the legacy importer if they need to (though idk what stops a user from going into edit mode, selecting the OBJ, and splitting by material)


StandingPadAnimations:
As mentioned https://github.com/TheDuckCow/MCprep/issues/231#issuecomment-1431578239, the colorspace issue lies in the MTL file. Hopefully I'll have a working fix by the end of today

StandingPadAnimations:
The root cause has finally been fixed. We do this in 2 stages.

The first stage is done in the following lines of code: 
```py
			BLENDER_STANDARD = (
				"Standard",
				"Filmic",
				"Filmic Log",
				"Raw",
				"False Color"
			)
			MTL = self.filepath.rsplit(".", 1)[0] + '.mtl'
			LINES = None
			if bpy.context.scene.view_settings.view_transform not in BLENDER_STANDARD:
				with open(MTL, 'r') as mtl_file:
					LINES = mtl_file.readlines()
					for index, line in enumerate(LINES):
						if line.startswith("map_d"):
							print(line)
							LINES[index] = "# " + line
				with open(MTL, 'w') as mtl_file:
					mtl_file.writelines(LINES)
```
Here we check to see if the view transform is part of Blender. If it isn't, we read the MTL in, edit some lines, and read it out. The second stage is the following:
```py
		if bpy.context.scene.view_settings.view_transform not in BLENDER_STANDARD:
			for index, line in enumerate(LINES):
				if line.startswith("# map_d"):
					LINES[index] = line[2:] # remove # and the space
			with open(MTL, 'w') as mtl_file:
				mtl_file.writelines(LINES)
```
This occurs at the end of the operator, and basically resets the MTL.

This may mean importing will most definitely be slightly slower, at least prior to Blender 3.1. Since we also take advantage of the C++ importer in Blender 3.1+, the overall performance should be much better

StandingPadAnimations:
Ok we should start weighing the tradeoffs of using the C+j OBJ importer for 3.1 - 3.4, since feature wise the C++ importer loses a major feature (split groups) that was only added in 3.5

StandingPadAnimations:
I think I'll keep it enabled in 3.1 - 3.4. The loss of functionality is made up for by the massive performance improvements.

I've noticed with even small OBJs the performance goes from seconds to instant imports, so who knows how this scales

TheDuckCow:
That's really awesome to hear. In terms of splitting the world back into by material, we could just run that function on import (that was the actual reason for using an custom operator for import, in the early days of 2.8 when everything was joined together). It will bring back a little bit of that slowness, but might still be worth it.

Thanks a ton for this, will try to get to review this. The imminent release urgency dropped as my tutorial sequence is not blocked on the need to fix sync materials (that still should be fixed tho). Will try to review this and your other change soon.

MCprep_addon/world_tools.py
```diff
@@ -310,8 +310,34 @@ def execute(self, context):
...
+			MTL = self.filepath.rsplit(".", 1)[0] + '.mtl'
+			LINES = None
+			if bpy.context.scene.view_settings.view_transform not in BLENDER_STANDARD:
+				with open(MTL, 'r') as mtl_file:
+					LINES = mtl_file.readlines()
+					for index, line in enumerate(LINES):
+						if line.startswith("map_d"):
+							LINES[index] = "# " + line
+				with open(MTL, 'w') as mtl_file:
+					mtl_file.writelines(LINES)
```
TheDuckCow, line 329: Would prefer to move all of this outside the try-except (though, with the open commands, should have their own). Good to keep try/excepts as narrow as possible, so it's also clearer to see what things could go wrong we are accounting for.

MCprep_addon/world_tools.py
```diff
@@ -310,8 +310,34 @@ def execute(self, context):
...
+
+			res = None
+			if util.min_bv((3, 5)):
+				res = bpy.ops.wm.obj_import(
+					filepath=self.filepath, use_split_groups=True) # Returns functionality missing in 3.1 - 3.4 
+			if util.min_bv((3, 1)):
```
TheDuckCow, line 335:
```suggestion
			elif util.min_bv((3, 1)):
```

Otherwise wouldn't it import twice?

MCprep_addon/world_tools.py
```diff
@@ -310,8 +310,34 @@ def execute(self, context):
...
+				res = bpy.ops.wm.obj_import(
+					filepath=self.filepath)
```
TheDuckCow, line 337: I'm going to say: let's not use the new importer if we can't have split by groups (materials, in practice). It breaks other parts of MCprep in a tough way.

I did a speed test of using new C++ with no splitting (~instance), to using C++ and then split materials python op (~13s) to the current way (~2.8s). Maintaining functionality which users are already expecting in 3.4 I think prevails, so just bucket this in the block below. And we can celebrate the improvement in 3.5 🥳

MCprep_addon/world_tools.py
```diff
@@ -394,6 +420,14 @@ def execute(self, context):
 
 		addon_prefs = util.get_user_preferences(context)
 		self.track_exporter = addon_prefs.MCprep_exporter_type  # Soft detect.
+
+		if bpy.context.scene.view_settings.view_transform not in BLENDER_STANDARD:
+			for index, line in enumerate(LINES):
+				if line.startswith("# map_d"):
+					LINES[index] = line[2:] # remove # and the space
+			with open(MTL, 'w') as mtl_file:
+				mtl_file.writelines(LINES)
```
TheDuckCow, line 429: I'm mildly wary about editing the mtl files and entirely breaking them. To confirm: The new importer doesn't help auto resolve color space changes at all?

If we are going to do this, let's 1) make a robust unit test around it, and 2 put this into its own function (so we can test in isolation). Happy to go over with you how to use the unit test script. It's rather non standard, but works well enough.

TheDuckCow (changes requested):
Largely good, but worried about editing mtl's and one reversion for 3.4 branch.

MCprep_addon/world_tools.py
StandingPadAnimations: I think what we could also do is copy the MTL, and either rename the original as something like `name.mtl.original`, or create a new folder and put the original MTL there (just to make the user's directory clean). Both have an additional benefit of avoiding redundant editing (in other words, if it's already been edited, we can skip it)

StandingPadAnimations:
Alright, MTL files should now be copied to a new folder called `ORIGINAL_MTLS` before they're edited

StandingPadAnimations:
Weird situation: Not sure why but somehow Kaion's master branch was merged with `support-all-color-management`. I'm looking into the issue, but I can't find the PR that somehow did it

StandingPadAnimations:
Ight, I still haven't found the reason for this, so sadly I'll have to close this for now

Seems like Github did something to the remote repo that caused this issue

StandingPadAnimations:
Ok so turns out when I forced reverted the merge commit, it didn't properly revert the merge. I'll look into reverting the force push (Github thankfully stored the commit in question)

@TheDuckCow do you know how to properly undo a merge?

StandingPadAnimations:
Ok so I've reverted the merge. Sadly the history remains but oh well. I'll continue to see if I can remove the extra commits that were introduced, but hopefully this should be fine now

StandingPadAnimations:
Good news, I've split this into 2 branches, and with some cherrypicking was able to move some commits over. I still need to reimplement MTL conversion but #382 now tracks support for the new C++ OBJ importer

StandingPadAnimations:
Alright, I've now restored both changes, and they have their own PRs

PR #374, opened 2023-02-21
Desaturation fix

StandingPadAnimations:
As mentioned in https://github.com/TheDuckCow/MCprep/issues/368#issuecomment-1438924696, the desaturation bug that's been recently plaguing MCprep for a while is caused by using an incorrect index. This PR should hopefully fix it (fingers crossed, I haven't tested it yet)

### DIFF
--- a/MCprep_addon/materials/generate.py
+++ b/MCprep_addon/materials/generate.py
@@ -955,10 +955,8 @@ def set_saturation_material(mat):
 		if len(desat_color) == 3:
 			desat_color += [1]  # add in alpha
 
-		if util.min_bv((3, 4)):
-			sat_node.inputs[6].default_value = desat_color
-		else:
-			sat_node.inputs[2].default_value = desat_color
+		sat_node_in = get_node_socket(node, is_input=True) # Get the node sockets in a version agnostic way	
+		sat_node.inputs[sat_node_in[2]].default_value = desat_color
 		sat_node.mute = not bool(saturate)
 		sat_node.hide = not bool(saturate)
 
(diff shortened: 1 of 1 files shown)


MCprep_addon/materials/generate.py
```diff
@@ -954,7 +954,11 @@ def set_saturation_material(mat):
 			return  # requires regenerating material to add back
 		if len(desat_color) == 3:
 			desat_color += [1]  # add in alpha
-		sat_node.inputs[2].default_value = desat_color
+
+		if util.min_bv((3, 4)):
+			sat_node.inputs[6].default_value = desat_color
```
TheDuckCow, line 959: On a spot check, shouldn't this be index 7? Index 6 is actually the "line in", ie the connected texture, while 7 would be the assignment of the actual color to apply.

```suggestion
			sat_node.inputs[7].default_value = desat_color
```

That being said, should we also consider using this new infra for version agnostic node selection? tbh i actually find your current code change easier to follow... but I guess if we want to be consistent we'd do something like below (95% sure this isn't quite right):

```
sat_node_in = get_node_socket(sat_node)
...
sat_node.inputs[sat_node_in[2]
```

TheDuckCow (changes requested):
Good fix to make, spot check the value and consider using the more generic selector

StandingPadAnimations:
Made the necessary change. `get_node_socket` returns a list, so I think the code was meant to access index 2 of that list (which would indeed be index 7 of the Mix node)

TheDuckCow (approved):
Good to go - please merge at your earliest convenience!

PR #380, opened 2023-03-06
Set default particle planes folder and fixed emitter material

TheDuckCow:
Set default particle planes folder, fixed emitter material, and now updated folder for particle planes is remembered.

Added scene variable `mcprep_particle_plane_file` which is initially unset, only assigned after a confirmed, non-cancelled particle plane effect placement.


StandingPadAnimations (approved):
Looks fine to me!

PR #381, opened 2023-03-07
Added some guides on commits in CONTRIUTING.md

StandingPadAnimations:


CONTRIBUTING.md
```diff
@@ -171,6 +171,41 @@ One other detail: MCprep uses git lfs or Large File Storage, to avoid saving bin
...
+MTL files. This brings a quality of life improvement to change the extension
+if the file selected is an MTL, since MTLs share the same name as their corresponding
+OBJ files
+```
+
+The first line is a summary of the changes, and should be less then 50 characters. The rest should justify the changes. Convince us why these changes are important and why they've been made this way.
+
+Git won't automatically wrap messages either, so each line should have a limit of 72 characters.
+
+Here's a template I found that can help (modified for simplicity) by using # to define which is the limit Git can display for each line:
```
TheDuckCow, line 195:
```suggestion
Here's a template some MCprep developers found that can help (modified for simplicity) by using # to define which is the limit Git can display for each line:
```

Suggesting this to keep the guide feel neutral, since it is indeed the result of multiple of our contributions.

TheDuckCow (approved):
Nice improvement 👍

PR #382, opened 2023-03-08
C++ OBJ importer

StandingPadAnimations:
This introduces support for the C++ OBJ importer for Blender 3.5 and above

### DIFF
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -313,10 +313,7 @@ class MCPREP_OT_import_world_split(bpy.types.Operator, ImportHelper):
 			res = None
 			if util.min_bv((3, 5)):
 				res = bpy.ops.wm.obj_import(
-					filepath=self.filepath, use_split_groups=True) # Returns functionality missing in 3.1 - 3.4 
-			if util.min_bv((3, 1)):
-				res = bpy.ops.wm.obj_import(
-					filepath=self.filepath)
+					filepath=self.filepath, use_split_groups=True) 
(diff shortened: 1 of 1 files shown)


MCprep_addon/world_tools.py
```diff
@@ -310,8 +312,13 @@ def execute(self, context):
 		obj_import_mem_msg = (
 			"Memory error during OBJ import, try exporting a smaller world")
 		try:
-			res = bpy.ops.import_scene.obj(
-				filepath=self.filepath, use_split_groups=True)
+			res = None
+			if util.min_bv((3, 5)):
+				res = bpy.ops.wm.obj_import(
+					filepath=self.filepath, use_split_groups=True)
```
TheDuckCow, line 318: I think this diff hunk can be reverted. This "use_split_groups" works in 3.5 and equally in 3.4 and earlier, as written this would make 3.4 and earlier stop splitting out by objects from the source, which would not be good.

TheDuckCow (approved):
Approving with the confidence that you'll make the once change (reversion) below to maintain consistent backwards behavior (it's actually less code too)

MCprep_addon/world_tools.py
StandingPadAnimations: Not sure what you mean, though I did catch the use of the wrong operator for 3.4 and below. In my testing, `use_split_groups` doesn't seem to work in 3.4

StandingPadAnimations:
Anyway, I'll merge this now

PR #384, opened 2023-03-10
Mark weather effect particle object as fake user

TheDuckCow:
Necessary, since imported particle objects are not loaded into the scene and thus would disappear after a file reload.


TheDuckCow:
FYI @StandingPadAnimations  - but I'm just going to merge since it's a 4 line change and I have it working locally.

PR #386, opened 2023-03-10
Code compatibility cleanup

TheDuckCow:
Addressed a few things to improve:

- Removed an f-string reference (invalid syntax in blender 2.7's version of python)
- Disabled the cycles optimizer for blender 2.7 (the draw function wasn't working even due to missing icon names, so nobody could have been using it so far there anyways; I did not address an issue where it doesn't seem to work in Blender 2.80, but is fine in 2.90)
- General style improvements in the files touched, mostly whitespaces.

### DIFF
--- a/MCprep_addon/mcprep_ui.py
+++ b/MCprep_addon/mcprep_ui.py
@@ -768,15 +768,16 @@ class MCPREP_PT_world_imports(bpy.types.Panel):
 			col.operator(
 				"mcprep.improve_ui", text="Improve UI", icon='SETTINGS')
 
-		# Optimizer Panel.
-		row = col.row(align=True)
-		icon = "TRIA_DOWN" if scn_props.show_settings_optimizer else "TRIA_RIGHT"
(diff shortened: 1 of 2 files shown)


StandingPadAnimations (approved):

TheDuckCow:
@StandingPadAnimations we certainly could. There are formatters out there for python, I'm just wary of doing such a large scale bulk change.

I'm not, however, opposed to giving it a try and a standalone branch. We would just need to halt any in progress dev/branches, try this out, if we like it merge it, and then start any new work. Anything that was in progress would have impossible merge conflicts. There's also the issue of renamed properties too, which the first time people update (...and don't restart blender like they should) would run into bugs of missing property names.

I'd say low priority for that reason in particular, but I could be convinced.

StandingPadAnimations:
I think what we could also do is make PEP8 formatting a "good first contribution" for people (who may not know a lot in terms of writing new code) wanting to contribute

PR #388, opened 2023-03-12
Add MTL conversion unit tests

TheDuckCow:
Will need to rebase once https://github.com/TheDuckCow/MCprep/pull/383 is merged, as that is the base for this branch.

These tests are being added to add extra sureness that the expected changes are happening when editing an mtl file.

### DIFF
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -178,8 +178,6 @@ def convert_mtl(filepath):
 
 	# This represents a new folder that'll backup the MTL filepath
 	original_mtl_path = Path(filepath).parent.absolute() / "ORIGINAL_MTLS"
-	# TODO: make sure this works in 2.7x. It should since 2.8 uses 3.7 but
-	# we should confirm nonetheless
 	original_mtl_path.mkdir(parents=True, exist_ok=True)
 
 	mcprep_header = (
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -167,7 +167,6 @@ def convert_mtl(filepath):
 	Returns:
 		True if success or skipped, False if failed, or None if skipped
 	"""
-	print("Convert MTL's spaces??", BUILTIN_SPACES)
 	mtl = filepath.rsplit(".", 1)[0] + '.mtl'
 	lines = None
 	copied_file = None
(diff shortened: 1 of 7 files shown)


MCprep_addon/world_tools.py
```diff
@@ -151,72 +160,72 @@ def convert_mtl(filepath):
...
+	if bpy.context.scene.view_settings.view_transform in BUILTIN_SPACES:
+		return None
 
-		mcprep_header = (
-			"# This section was created by MCprep's MTL conversion script\n",
-			"# Please do not remove\n",
-			"# Thanks c:\n"
-		)
+	# This represents a new folder that'll backup the MTL filepath
+	original_mtl_path = Path(filepath).parent.absolute() / "ORIGINAL_MTLS"
```
TheDuckCow, line 180: @StandingPadAnimations is there a reason to use this over something more like:

```suggestion
	original_mtl_path = Path(os.path.join(filepath, "ORIGINAL_MTLS")).parent.absolute()
```

I just worry about the slash being OS dependent. But if it worked for you (on a windows machine?) then I'm good with it.

TheDuckCow: And I did test on 2.7x, works fine.

TheDuckCow: Well I tested my change above, and it actually breaks the unit tests (a sign they are working), not sure exactly why though, I'm also just less familiar with the Path module.

So, really the main thing is if you could confirm you were testing on windows, then I'm all set.

StandingPadAnimations: I'm on Linux but the [Python documentation](https://docs.python.org/3/library/pathlib.html#operators) says that it should be fine.

TheDuckCow: Thanks for the doc reference, good to know. Unless you have any particular complaints here, I'll go ahead and merge in a moment.

StandingPadAnimations:
I just need to check the Blender C++ OBJ code to see if indeed MTL conversion is needed for 3.5 and above

StandingPadAnimations:
Alright, I don't see any mention of "Non-Color Data" anywhere, so it's hard for me to determine if indeed the C++ OBJ importer has an intended behavior change. For now I think it's somewhat safe to say that it's an oversight, but I can't be certain

StandingPadAnimations (approved):

PR #394, opened 2023-03-14
Updated version in README

StandingPadAnimations:
This seems to have been overlooked in the previous update


TheDuckCow (approved):
Let's mark it as through 3.5! I did testing against that. Good catch (and solid rare use of pushing to master directly)

PR #395, opened 2023-03-14
Updated version in README

StandingPadAnimations:
This seems to have been overlooked in the previous update


TheDuckCow:
Ah I meant, you could directly update to say 3.5 (and, that you could also push to master)

PR #397, opened 2023-03-15
Fix MTL conversion bug with Mineways MTLs

StandingPadAnimations:
There's been a issue recently where Mineways will name MTLs weirdly. For
instance:
	- "Beach Town.obj" : "Beach_Town.mtl"
	- "hello there.obj" : "hello_there.mtl"

The MTL conversion function simply replaces the extension of
self.filepath, assuming that the MTL has the same name. However, since
Mineways adds underscores, convert_mtl would throw an exception when
attempting to read the file. The current workarounds are to:
	- Remove the underscores
	- Don't add spaces to OBJ names (ideally no one should be adding
	  spaces to any filename because of issues like this)
	- Use jmc2OBJ, which as far as I know doesn't have this weird
	  behavior

This is more of a band-aid, and ideally we should check for underscores
in the MTL name (we can take advantage of pathlib for this), but for now
if an exception occurs due to a weirdly named MTL, we'll just print the
error, return False, and call it a day.

### DIFF
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -170,8 +170,13 @@ def convert_mtl(filepath):
 	mtl = filepath.rsplit(".", 1)[0] + '.mtl'
 	lines = None
 	copied_file = None
-	with open(mtl, 'r') as mtl_file:
-		lines = mtl_file.readlines()
+
+	try:
+		with open(mtl, 'r') as mtl_file:
+			lines = mtl_file.readlines()
+	except Exception as e:
+		print(e)
+		return False
 
 	if bpy.context.scene.view_settings.view_transform in BUILTIN_SPACES:
 		return None


MCprep_addon/world_tools.py
```diff
@@ -170,8 +170,13 @@ def convert_mtl(filepath):
 	mtl = filepath.rsplit(".", 1)[0] + '.mtl'
 	lines = None
 	copied_file = None
-	with open(mtl, 'r') as mtl_file:
-		lines = mtl_file.readlines()
+
+	try:
+		with open(mtl, 'r') as mtl_file:
+			lines = mtl_file.readlines()
+	except Exception as e:
```
TheDuckCow, line 177: Can we do more narrow exception handling? e.g. file not found, and then still have a catch all at the end.

MCprep_addon/world_tools.py
```diff
@@ -170,8 +170,13 @@ def convert_mtl(filepath):
 	mtl = filepath.rsplit(".", 1)[0] + '.mtl'
 	lines = None
 	copied_file = None
-	with open(mtl, 'r') as mtl_file:
-		lines = mtl_file.readlines()
+
+	try:
+		with open(mtl, 'r') as mtl_file:
```
TheDuckCow, line 175: I'd also argue we may as well do the check for which files exist a couple lines before.

If we know blanket-wise it converts spaces to underscore, let's just add it in now:

```python
mtl_underscore = os.path.join(
    os.path.dirname(mtl_underscore),
    os.path.basename(mtl_underscore).replace(" ", "_")
if not os.path.isfile(mtl) and os.path.isfile(mtl_underscore):
    mtl = mtl_underscore
```

.. but would be good to sanity check it too, of course.

TheDuckCow (changes requested):
Hey, am I reading this right, that this would be causing complete import errors for a large swath of users? If so, we might need to do an emergency patch release. Unfortunately my error analytics pipeline slightly broke so we're flying blind for a couple days until I fix that (cloud related issue only).

I also should have thought to error handle that and later lines.

MCprep_addon/world_tools.py
StandingPadAnimations: I've went ahead and implemented an existence check in the beginning of the function, but with pathlib:
```py
mtl = Path(filepath.rsplit(".", 1)[0] + '.mtl')
if not mtl.exists():
	mtl_underscores = Path(mtl.parent.absolute()) / mtl.name.replace(" ", "_")
	if mtl_underscores.exists():
		mtl = mtl_underscores
	else:
		return False
```

Sidenote: Pathlib seems to be a lot easier and more readable when doing file operations, we should use it more

TheDuckCow (approved):
Didn't live test with a model, but all unit tests are passing on this branch. Good to merge imo, and will continue to plan for another release next week as scheduled.

Issue #403, opened 2023-03-19
Warden causing Reload Mob errors in 2.8x and 2.9x


TheDuckCow:
For some reason, 2.8x and 2.9x users are encountering errors when importing specifically the warden rig:

```
File "<addon_path>/mobs.py", line 206, in execute
 update_rig_list(context)
 File "<addon_path>/mobs.py", line 148, in update_rig_list
 _add_rigs_from_blend(blend_path, blend_name, category)
 File "<addon_path>/mobs.py", line 65, in _add_rigs_from_blend
 with bpy.data.libraries.load(path) as (data_from, data_to):
 OSError: load: C:\Users\{REDACTED}\AppData\Roaming\Blender Foundation\Blender\2.91\scripts\addons\MCprep_addon\MCprep_resources\rigs\hostile\Warden - DigDanAnimates.blend failed to open blend file
```

I see no reports for Blender 3.x implying it's a compatibility problem. This is likely rig asset fix we need to make.

This accounts for 122 of the 704 bugs reported in the last few days. As a breakdown of current bug reporting:

- 494 (70%) are due to the material conversion issue.
- 122 for THIS warden rig issue (17%)
- 76 for the block material failing to import (10%)
- Only 12 for truly anything else (~2%) (mostly reloading errors or more unique fatal issues like MCprep props not being registered, I'd bucket under reloading blender issues)

TheDuckCow:
There is also a more general improvement that should happen, apparently the issue is happening actually during "mob reloading", so really any flaky mob model will break the loading of all of them. We should likewise make mob lading much safer during the reading of a blend file, fail by skipping that single entry.

StandingPadAnimations:
> I see no reports for Blender 3.x implying it's a compatibility problem. This is likely rig asset fix we need to make.

*I sense a perfect "first contribution" type issue for new developers*

TheDuckCow:
I would agree would ordinarily be a good first contrib, but there's some strong urgency behind this one that we'll want to resolve it in the next couple days.

I just did a test myself, and I'm not seeing any issues loading the warden rig there. I am using blender 2.90. Looking back at the error reporting database, I see:

- Only one error for a linux user (likely a fluke)
- 153 (all others) are from windows users,
- 0 Mac user reports (but mac users are also already a minority).
- Reports coming from blender versions including the full list below

Full blender version list with at least 1+ report (so once again, seems like a pretty even spread of version ≥ 2.80 and < 3.0)

```
[...]

PR #405, opened 2023-03-21
Fix bug in spawn_util

CFeeney333:
Return False not True in order to skip the current file (this_file)

### DIFF
--- a/MCprep_addon/spawner/spawn_util.py
+++ b/MCprep_addon/spawner/spawn_util.py
@@ -158,13 +158,6 @@ def check_blend_eligible(this_file, all_files):
 				# instead, so return False for the current file
 				return False
 
-			# tuple_ver = tuple_from_match(matches)
-			# res = util.min_bv(tuple_ver)
-			# # If the suffix = 3.0 in `afile` file, and current blender is
-			# # below 3, then `afile` is the file that should be loaded, and the
-			# # current file `this_file` is to be skipped.
-			# return not res
-
 	# If no matches (the most common case), then this file is eligible.
 	return True
 
--- a/MCprep_addon/spawner/spawn_util.py
+++ b/MCprep_addon/spawner/spawn_util.py
@@ -146,12 +146,24 @@ def check_blend_eligible(this_file, all_files):
 		base_afile = os.path.splitext(afile)[0]
 		matches = find_suffix.search(base_afile)
 		if matches:
-			tuple_ver = tuple_from_match(matches)
-			res = util.min_bv(tuple_ver)
-			# If the suffix = 3.0 in `afile` file, and current blender is
-			# below 3, then `afile` is the file that should be loaded, and the
-			# current file `this_file` is to be skipped.
-			return not res
+			# afile is guaranteed to have the same basename, not be the same file, and has a suffix
+			# because it has a suffix, the recursively called function will not reach the containing for loop
+			if not check_blend_eligible(afile, all_files):
+				# if afile is not eligible, we can continue to the next iteration to see if there is another
+				# file with a suffix that is eligible
+				# previously this only allowed for 1 other file with a suffix to be checked
+				continue
+			else:
+				# if a file with a suffix is eligible here, we should not use the current file, but should use afile
+				# instead, so return False for the current file
+				return False
+
+			# tuple_ver = tuple_from_match(matches)
+			# res = util.min_bv(tuple_ver)
+			# # If the suffix = 3.0 in `afile` file, and current blender is
+			# # below 3, then `afile` is the file that should be loaded, and the
+			# # current file `this_file` is to be skipped.
+			# return not res
 
 	# If no matches (the most common case), then this file is eligible.
 	return True
--- a/MCprep_addon/spawner/spawn_util.py
+++ b/MCprep_addon/spawner/spawn_util.py
@@ -151,7 +151,7 @@ def check_blend_eligible(this_file, all_files):
 			# If the suffix = 3.0 in `afile` file, and current blender is
 			# below 3, then `afile` is the file that should be loaded, and the
 			# current file `this_file` is to be skipped.
-			return res
+			return not res
 
 	# If no matches (the most common case), then this file is eligible.
 	return True
--- a/MCprep_addon/spawner/spawn_util.py
+++ b/MCprep_addon/spawner/spawn_util.py
@@ -98,21 +98,29 @@ def filter_collections(data_from):
 
 
 def check_blend_eligible(this_file, all_files):
-	"""Returns true if the path blend file is ok for this blender version.
+	"""Returns true if this_file is the BEST blend file variant for this rig.
 
 	Created to better support older blender versions without having to
-	compeltely remake new rig contirbutions from scratch. See for more details:
+	completely remake new rig contributions from scratch. See for more details:
 	https://github.com/TheDuckCow/MCprep/issues/317
 
-	If the "pre#.#.#" suffix is found in any similar prefix named blend files,
-	evaluate whether the target file itself is eligible. If pre3.0.0 is found
-	in path, then this function returns False if the current blender version is
-	3.0 to force loading the non pre3.0.0 version instead, and if pre3.0.0 is
-	not in the path variable but pre#.#.# is found in another file, then it
-	returns true. If no pre#.#.# found in any files of the common base, would
-	always return True.
+	This function searches for "pre#.#.#" in the filename of passed in files,
+	where if found, it indicates that file should only be used if the current
+	running blender version is below that version (non inclusive). The function
+	then finds the highest versioned file and checks if this_file is that
+	the same one. If so, it returns true, otherwise returns false. Most rigs
+	do not have versioned names are are assumed to work with the latest blender
+	version, hence they are treated as a "max_ver" to never get thrown out.
+
+	Examples, presuming current blender is v2.93
+	for list ["rig pre2.8.0", "rig"]
+		-> returns true for "rig"
+	for list ["rig pre2.8.0", "rig pre3.0.0", "rig"]
+		-> returns true for "rig pre3.0.0"
+
 	"""
 	basename = os.path.splitext(this_file)[0]
+	max_ver = (99, 99, 99)  # Standin for an impossibly high blender version.
 
 	# Regex code to any matches of pre#.#.# at very end of name.
 	# (pre) for exact match,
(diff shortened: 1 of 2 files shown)


CFeeney333:
So it seems it was just a case of a logic error. If the version of 'afile' is greater than the blender version, then this is the version that should be used. 'this_file' should be discarded, so False should be returned

This works in blender 2.93.15 now. However, this is the only version that I have tested it with

StandingPadAnimations (approved):

TheDuckCow:
Wow not sure how that snuck through. I do find it rather confusing though that I can't produce the error locally - but this indeed works for you locally? Can you share a screenshot of your hostile mobs folder to show all files present?

And thanks a million for diving in and proactively making a PR, hoping we can approve and merge this today (just bear with me while I attempt another live test later).

CFeeney333:
I'm away from the machine where I did the testing, so I can't take screenshots now (it's a college computer). However, I can confirm from memory that there were three warden rigs in the hostile mobs folder - pre-2.8.0, pre-3.0.0 and an unversioned one. Reload mobs worked on blender 3.4, and I downloaded 2.93 specifically to test the error, and that's the screenshot in #403. I can also confirm that after the changes, reload mobs worked in blender 3.3 and 2.93, when I tested it with the changes on my own machine. I can only imagine there's a discrepancy between the rigs we have or the blender versions we're using...

TheDuckCow:
Thanks for sharing the details. Do you remember the detail of whether it was an MCprep _upgrade_ or a fresh install in that 2.93 instance?

I'm looking back, we had this line of code/entire function last modified ahead of the 3.4.0 release - which was the same release that the warden was introduced (https://github.com/TheDuckCow/MCprep/issues/318) with 3 split up files even at that time. Even the lines of code where we call `check_blend_eligible` was unmodified since that release 8 months ago. I'm really head scratching why we haven't gotten this bug in v3.4.0 and v3.4.1 before, and yet it's such a sudden surge now.

Furthermore, testing this change just based on our unit tests, I actually start getting this assertion to fail in our unit test runner. It's true there could be something wrong with the unit test, but I need to figure out first what is going wrong better. 
https://github.com/TheDuckCow/MCprep/blob/master/test_files/addon_tests.py#L718

Also, I head turned slightly - funny that pull/405 is for issue 403, so close to being one and the same!

TheDuckCow:
Alright, I've figured out the culprit - the algorithm is too greedy it seems up front in returning "true, is eligible", and shouldn't really be exiting early until it's tested all files in the list. I'm going to check out your branch and make some contributions, so you can still get the code contribution credit, just be aware I'm making a number of changes.

I'm creating a new unit test called `check_blend_eligible_middle` which tests this function for this scenario of 2.9x user, when there are 3 files present, the equivalent of `pre2.8`, `pre3.0`, and `(current live)`, asserting that it _only_ returns true for the `pre3.0` scenario (right now, it's returning true for the "current" one as well, hence blender 2.9x users files are trying to open the post 3.0 versions, as far as I understand it).

Still doesn't explain why this hasn't been an issue till now, but I at least understand where things are going wrong. Could always be the order that files were listed or loaded in rig lists? So it more just _happened_ to work before.

CFeeney333:
OK, so when I ran it in 2.93, I ran it from pycharm. I cloned the repo to my projects folder. From a while back, I had copied the mcprep_resources from the 3.4.1 addon to this repo. I used a pycharm plugin blend charm to mark the mcprep addon dir as a blender addon which meant I could run/debug the addon from the ide. The plugin installs the addon to the instance of blender you are using when you hit run. So if you were hoping I used a conventional method of running blender, I didn't. It would be curious to see, though, if its just the order of the blend files in the folder that causes some users to get an error as you were saying

CFeeney333:
Ok. After another look, I realised that the code only allowed for 1 check for an eligible versioned file, even if there were two. So I guess my previous fix just happened to work because of the order of the files in the list. The new fix should work, regardless of the order of the files, as it was originally intended to do. At this stage, I wouldn't be surprised at all if the error only popped up depending on the order of the files in the list. I wonder if a simple sort() would have worked as a hacky fix

MCprep_addon/spawner/spawn_util.py
```diff
@@ -146,12 +146,24 @@ def tuple_from_match(match):
...
+			if not check_blend_eligible(afile, all_files):
+				# if afile is not eligible, we can continue to the next iteration to see if there is another
+				# file with a suffix that is eligible
+				# previously this only allowed for 1 other file with a suffix to be checked
+				continue
+			else:
+				# if a file with a suffix is eligible here, we should not use the current file, but should use afile
+				# instead, so return False for the current file
+				return False
+
```
CFeeney333, line 160: If there are no eligible versioned files, it will fall through to return True at the end of the function anyway

MCprep_addon/spawner/spawn_util.py
```diff
@@ -146,12 +146,24 @@ def tuple_from_match(match):
...
+			# tuple_ver = tuple_from_match(matches)
```
StandingPadAnimations, line 161: Just curious why this section is commented out. If it's not necesary, then just remove it

StandingPadAnimations (changes requested):
Mostly some stuff regardding comments, but otherwise looks good

MCprep_addon/spawner/spawn_util.py
```diff
@@ -146,12 +146,24 @@ def tuple_from_match(match):
 		base_afile = os.path.splitext(afile)[0]
 		matches = find_suffix.search(base_afile)
 		if matches:
-			tuple_ver = tuple_from_match(matches)
-			res = util.min_bv(tuple_ver)
-			# If the suffix = 3.0 in `afile` file, and current blender is
-			# below 3, then `afile` is the file that should be loaded, and the
-			# current file `this_file` is to be skipped.
-			return res
+			# afile is guaranteed to have the same basename, not be the same file, and has a suffix
```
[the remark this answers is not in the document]
CFeeney333: Will do. Keep to 80?

TheDuckCow:
Thanks for reviewing, and @CFeeney333 for making further changes. I'm going to ask to hold on for a bit until I can tack on a couple changes I've made and am continuing on with locally, I agree with your assessment of it being more chance that it was working _at all_ before.

TheDuckCow:
Hey @CFeeney333 can you give me write/push access to your [MCprep fork](https://github.com/CFeeney333/MCprep)? Then I can push my one commit up and we can merge, and you'll appear in the history correctly.

CFeeney333:
https://github.com/CFeeney333/MCprep/invitations

TheDuckCow (approved):
So of course I'd approve the changes I've made. Seeking input from @StandingPadAnimations 

I don't like how complicated this function became, but even after spending some time looking at it, I didn't see a way to (substantially) simplify it. We have to look at all filenames in the list before we can determine if this_file should return true. That was really the crux of the issue before, it was returning true (I imagine) for the `warden.blend` when it hadn't yet checked to see if it should prefer `warden pre3.0.0.blend`.  Unit tests now demonstrate this working as intended.

MCprep_addon/spawner/spawn_util.py
```diff
@@ -129,32 +137,59 @@ def tuple_from_match(match):
...
+	if len(other_eligible) == 1:
+		# Only this_file is eligible, and thus must be "the one".
+		return True
+
+	# If multiple files, then we should use the one that is the largest version
+	# without being equal to or greater than the current running blender ver.
+	sorted_eligible = sorted(other_eligible, key=lambda x: x[0])
+	latest_allowed = None
+	for tple, afile in sorted_eligible:
+		if util.min_bv(tple, inclusive=True) is False:
```
StandingPadAnimations, line 187: Same thing here

StandingPadAnimations (changes requested):
I'll have to look further in depth later today, but here's the things I found at this moment

MCprep_addon/spawner/spawn_util.py
```diff
@@ -129,32 +137,59 @@ def tuple_from_match(match):
...
-			res = util.min_bv(tuple_ver)
-			# If the suffix = 3.0 in `afile` file, and current blender is
-			# below 3, then `afile` is the file that should be loaded, and the
-			# current file `this_file` is to be skipped.
-			return res
-
-	# If no matches (the most common case), then this file is eligible.
-	return True
+			res = util.min_bv(tuple_ver, inclusive=True)
+			if res is False:
```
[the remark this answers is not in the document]
TheDuckCow: Yup, I spent a lot of time starting and it and it ended up helping me read it having the argument specifically there. I nearly just removed the function call in favor of bpy.app.version to make it easier to read, but we can do that in the 3.5 release (even though bpy.app.version is safely available in 2.79, no reason to break the compatibility lower just for this).

TheDuckCow:
FYI @StandingPadAnimations likely to merge tomorrow, no sweat if you don't have time. Happy to hear input too from you @CFeeney333 to verify this still makes it work well on your side, that would actually be a good sanity check since I still have yet to produce the error myself.

StandingPadAnimations:
Whoops, forgot about this. Looking over it again I think it should be fine to merge (we can ignore min_bv for now, I'll be doing some version check changes in 3.5 anyway)

StandingPadAnimations (approved):

CFeeney333:
https://user-images.githubusercontent.com/96747360/227708156-10af286f-5219-499b-8e52-ef352de104bf.mp4

Ig this is the quickest way to verify. My dev branch was updated with your changes before recording that, and blender 3.3.5 works with both branches

image

order in blend files in list (2.93, dev branch)

TheDuckCow:
Darn, even with my unittest still not working. I'm quite sure though we still need to loop over files. Thanks for the screenshot, I've extract that and will do some more testing on it today, creating or updating the test script.

TheDuckCow:
Alright, this is progress at least: The unit tests now fail for the expected versions of blender, consistent with your screen recording (even though blender 2.93 still works fine for me). Next stop: Actually making the function work as expected.

Test results csv
```
blender	failed_test	short_err
(3, 5, 0)	ALL PASSED	-
(3, 4, 0)	ALL PASSED	-
(3, 2, 1)	ALL PASSED	-
(3, 3, 1)	ALL PASSED	-
(3, 1, 0)	ALL PASSED	-
(3, 0, 0)	ALL PASSED	-
(2, 93, 0)	check_blend_eligible_real	DiaDanAnimates.blend
(2, 90, 1)	check_blend_eligible_real	DiaDanAnimates.blend
(2, 80, 75)	check_blend_eligible_real	DiaDanAnimates.blend
```

This is after running the command `./run_tests.sh -all -run check_blend_eligible_real` 

I realize the "run all" control does not exist yet in the windows bat script, so one would have to manually specify the first entry in their `blender_execs.txt` file to make it run that version of blender. For context, this is my "blender_execs.txt" file, and hence these are the versions that get tested when I run the script above.

```
/Applications/blender 3.5/Blender.app/Contents/MacOS/Blender
/Applications/blender 3.4/Blender.app/Contents/MacOS/Blender
/Applications/blender 3.2/Blender.app/Contents/MacOS/Blender
/Applications/blender 3.3/Blender.app/Contents/MacOS/Blender
/Applications/blender 3.1/Blender.app/Contents/MacOS/blender
/Applications/blender 3.0/Blender.app/Contents/MacOS/Blender
/Applications/blender 2.93/Blender.app/Contents/MacOS/Blender
/Applications/blender 2.90/Blender.app/Contents/MacOS/Blender
/Applications/blender 2.80/Blender.app/Contents/MacOS/Blender
```

CFeeney333:
Wait, the occasion that the exception occurs is the code without the changes. You are aware of that?

I just noticed in the test results it says the rig is by 'DiaDanAnimates' and not 'DigDanAnimates'. Does that mean anything?

TheDuckCow:
Alright I take it back, the only reason my tests were failing was because I had a typo in the rig list I copied from you (my image-to-text reader picked up `a` instead of `g` in one spot). Once that typo was fixed, all tests now pass. And this new test is a live integration test essentially, opening in multiple versions of blender and validating the right file is picked. I've also checked in the UI that in 2.93 for me it's picking up the right one (screenshot below).

@CFeeney333 I'm really hoping this is somehow it was not working on your machine due to `__pycache__` files or something like that in the addons directory. Can you try completely uninstalling MCPrep (delete the addon from the addons folder entirely), restart, and then re-open again after having done a git pull once more? (I technically only changed test files though, but just to be safe).
 
picking-correct

I don't want to push this release out if we think this isn't working. And I know the original logic was flawed for sure, it needs to operate over the entire list before checking if a given file is the right one. So even though your initial change was working for you, I believe it's still only by chance in essence (no different to how it was working before). Though let me know if you somehow see it differently and I'm misreading all of this still somehow.

TheDuckCow:
> Wait, the occasion that the exception occurs is the code without the changes. You are aware of that? ie. doesn't that mean the function now works as expected? (The exception occurs on my master branch as I hadn't merged the changes)
> 
> I just noticed in the test results it says the rig is by 'DiaDanAnimates' and not 'DigDanAnimates'. Does that mean anything?

OH just now saw this, totally misunderstood that! So, to your understanding, it's all working now? And yes, twas a typo.

CFeeney333:
No worries. Yes, it's all working now...

PR #407, opened 2023-03-25
Forcing users to restart if there is any update detected.

TheDuckCow:
This applies to both manual as well as addon-updater updates. The reason for this change is to avoid errors that will happen in future versions when we rename modules, remove functions, and generally revamp the structure of MCprep. Many users encounter errors due to not restarting Blender already, even with 'typical' updates.

closes https://github.com/TheDuckCow/MCprep/issues/402

### DIFF
--- a/MCprep_addon/mcprep_ui.py
+++ b/MCprep_addon/mcprep_ui.py
@@ -62,11 +62,10 @@ def addon_just_updated():
 	# thorough check by seeing if the mcprep_data_update.json exists (ie it
 	# hasn't been renamed yet to mcprep_data.json, which happens on init after
 	# an install/update)
-	global last_check_for_updated
 	check_interval = 5  # Time in seconds
-	if time.time() - check_interval > last_check_for_updated:
+	if time.time() - check_interval > conf.last_check_for_updated:
 		check_for_updated_files()
-		last_check_for_updated = time.time()
+		conf.last_check_for_updated = time.time()
 	return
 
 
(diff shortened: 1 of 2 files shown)


MCprep_addon/mcprep_ui.py
```diff
@@ -845,6 +903,10 @@ class MCPREP_PT_bridge(bpy.types.Panel):
 	bl_category = "MCprep"
 
 	def draw(self, context):
+		if addon_just_updated():
```
StandingPadAnimations, line 906: It would be nice if there was a short comment above these explaining what they do for any future developers

MCprep_addon/mcprep_ui.py
```diff
@@ -2007,6 +2106,10 @@ def register():
 		bpy.types.IMAGE_MT_uvs.append(mcprep_uv_tools)
 	# bpy.types.IMAGE_MT_image.append(mcprep_image_tools) # crashes, re-do ops
 
+	# Used to avoid checking for update file on disk too frequently.
+	global last_check_for_updated
```
StandingPadAnimations, line 2110: Wouldn't this be better off in `conf.py`?

StandingPadAnimations (changes requested):
Just a couple of minor things needed

MCprep_addon/mcprep_ui.py
TheDuckCow: Good call, definitely better there

MCprep_addon/mcprep_ui.py
TheDuckCow: Do you think it's sufficient to have that comment in the code of those functions? Otherwise we'll be having copies of the comment in 10x places, adding to the boilerplate.

StandingPadAnimations: Should be fine as long as it's explained what the functions do and why they do that they do

StandingPadAnimations (approved):

StandingPadAnimations:
Looks good now

MCprep_addon/mcprep_ui.py
TheDuckCow: So... would you modify the docstrings that are there now I guess is the question :)

PR #423, opened 2023-05-11
Big formatting commit

StandingPadAnimations:
In the past, I said we should use darker for only formatting changes, and while that's true, I think it makes more sense to just begin running an autoformatter and have a large "formatting commit"

For one thing, this is already a massive MCprep update, and already have Python changes from newer versions. In addition, as the author of Black says on his blog
(https://lukasz.langa.pl/36380f86-6d28-4a55-962e-91c2c959db7a/):

> When you make this single reformatting commit, everything that comes
> after is semantic changes so your commit history is clean in the sense
> that it actually shows what changed in terms of meaning, not style.
> There are tools like darker that allow you to only reformat lines that
> were touched since the last commit. However, by doing that you forever
> expose yourself to commits that are a mix of semantic changes with
> stylistic changes, making it much harder to see what changed.

In other words, if we decide to only format changes and keep the rest of the code untouched, we end up making it harder to understand changes. In addition, there would be consistency issues. I think it makes sense to just bite the bullet, create a massive formatting commit, and in the future use autoformatting as a part of our toolchain.

None of the actual underlying code has been changed, just the formatting. No renamed properties, no renamed classes, etc, so there shouldn't be any issues.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -207,7 +207,7 @@ Add this to a file called .gitmessage, and then execute the following command:
 To use for each commit, you can use `git config --local commit.verbose true` to tell Git to perform a verbose commit all the time for just the MCprep repo.
 
 ## Dependencies
-If you're using an IDE, it's recommened to install `bpy` as a Python module. In our experience, the [fake-bpy package](https://github.com/nutti/fake-bpy-module) seems to be the best. In addition, we also use `darker` to perform PEP8 formatting on changed code (this is a requirement, we expect you to use `darker` on your changes)
+If you're using an IDE, it's recommened to install `bpy` as a Python module. In our experience, the [fake-bpy package](https://github.com/nutti/fake-bpy-module) seems to be the best. In addition, we also use `black` to perform PEP8 formatting on changed code (this is a requirement, we expect you to use `darker` on your changes)
 
 It's also recommened to use a virtual environment (especially if you're on Linux) as to avoid issues with system wide packages and different versions of `bpy`. [See this for more details](https://realpython.com/python-virtual-environments-a-primer/)
 
@@ -226,7 +226,7 @@ If you decide to use Poetry, then simply run the following command:
 
 `poetry install`
 
-To enable the virtual environment, run `poetry shell`, then type `exit` when you're done. When you make a change to a file, **please please please** run `poetry run darker <changed file>` to apply PEP8 formatting to the changes (or if you've enabled the virtual environment, you can just type `darker <changed file>`)
+To enable the virtual environment, run `poetry shell`, then type `exit` when you're done. When you make a change to a file, **please please please** run `poetry run black <changed file>` to apply PEP8 formatting to the changes (or if you've enabled the virtual environment, you can just type `darker <changed file>`)
 
 ### Manual: Requirements.txt Edition
 First create a virtual environment:
@@ -247,7 +247,7 @@ Install dependencies:
 
 `python3 -m pip install -r requirements.txt`
 
-### Manual: Setting up `bpy` + `darker` Manually Edition
+### Manual: Setting up `bpy` + `black` Manually Edition
 First, we need to come up with a name. For MCprep development, it's recommended to use the following convention:
 `mcprep_venv_<version>`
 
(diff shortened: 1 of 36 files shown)


StandingPadAnimations:
@TheDuckCow just curious on your thoughts. I think we should start using auto formatting. That means a massive commit, but future code will comply with PEP8. I'm using https://github.com/psf/black to format the code, and in theory we could just add a pre-commit hook for formatting (or a GitHub Action) so future code will be formatted.

StandingPadAnimations:
Funnily enough the auto-formatter actually caught an `elif` statement that I missed earlier, all the more reason to start adding this to our toolchain:
https://github.com/TheDuckCow/MCprep/blob/99ebce857903fd06166359ef9188260ddbe1430a/MCprep_addon/mcprep_ui.py#L2085

TheDuckCow:
Hey @StandingPadAnimations sorry I didn't get to commenting before you went ahead on this. I was going to say I'm not supportive of moving to black for a number of reasons.

1. It'll be jarring for new contributors as an extra requirement to get started
2. While it is a popular linting/formatting convention, it's not the _most_ popular anyways.
3. I'll use my owner hammer to say, I also don't have the mental capacity to be trying to get used to another linting system, especially such a strict one like black (even if it is "auto formatting"). I'd rather we enforce the looser but widely accepted pep8 linting standard only, or at least build into our compile script pep8 checking to push out warnings over changed diffs.
4. Being a strict autoformatting, it can't account for local or custom style - for instance, how even blender has its own guideline around treating enum strings with single quotes and other "real" strings with double quotes. Having a strict autoformatter gets in the way of anything like that / I'm sure will lead to some 'in fighting' with the formatter.
5. A massive change like this, even if auto formatted, is basically impossible to QC fully. We've already made massive changes with recent recore and the milestone branch as you've said. Are tests are already not fully working. Making this change name would just make getting those in order even more difficult. We would also completely invalidate any in progress PRs making merging impossible basically (lesser concern I suppose, since we could always re-run the auto formatting later).
    - If we were to do a similar formatting to pep8 standards, it too would be a massive change. I'm likewise not supportive of doing that at this time, at least not until we have functional unittests that everyone can run. Same point applies of it would invalidate any open PRs or in progress work, so should we ever do it, it would need to be at a time that everything else not yet merged we are ok with completely dropping.
    - Just to put some weight behind this perspective so it doesn't feel arbitrary, it is a hotly debated topic of incremental vs bulk style changes. I'm on the side and regularly practice incremental changes.

So all in, I do appreciate the idea and your perspective of the value this would give, but I feel it's a step too far. If you're still invested in trying to make this better, I'd see what tooling we could make with the rebuild of compile to support diff style checking for pep8 (or is that already set up with poetry? I haven't had time to even check that).

Not closing the PR w/o merging yet to give you and maybe @zNightlord a chance to counter reply, but otherwise expecting to.

StandingPadAnimations:
Addressing some of the concerns:
> 1. It'll be jarring for new contributors as an extra requirement to get started

But the benefit would be consistent style among contributed code. We could alternatively use GitHub Actions or pre-commit hooks to make it easier. 

> 2. While it is a popular linting/formatting convention, it's not the most popular anyways.
> 3. I'll use my owner hammer to say, I also don't have the mental capacity to be trying to get used to another linting system, especially such a strict one like black (even if it is "auto formatting"). I'd rather we enforce the looser but widely accepted pep8 linting standard only, or at least build into our compile script pep8 checking to push out warnings over changed diffs.

Black is a PEP8 formatter, although having a compiler check might be a good compromise

> 4. Being a strict autoformatting, it can't account for local or custom style - for instance, how even blender has its own guideline around treating enum strings with single quotes and other "real" strings with double quotes.

This I'm curious on, seems weird Blender has that issue

Regardding number 5, this is the way I see it:
- I agree with the change part, so perhaps we could introduce this near the end of the 3.5 development cycle
- Black verifies that the formatted code is valid, so that's a non-issue
- When it comes to PRs, I think there's 2 ways to mitigate it:
    - Incrementally run the autoformatter on a handful of files
    - Give advanced warning and do it on all files

The problem with incremental changes is consistency. Sections won't match, Python will throw tabs vs spaces errors (we already get that a lot), and arguably the code will be even more unreadable. To me, it makes more sense to just make a giant commit to make things consistent, add the formatter requirement, and be done with it.

PR #418, opened 2023-04-29
MCprep 3.5 - Annotate and docummented util functions

zNightlord:
As part of #401 add annotations to common functions
And #315 add a README docummented util.py

PR #425, opened 2023-05-11
Annotate documment functions

zNightlord:
Based on #418


MCprep_addon/README.md
```diff
@@ -0,0 +1,362 @@
+### Utilities
```
TheDuckCow, line 1: Either the filename or location seems off. README generally refers to a user, whereas `CONTRIBUTING.md` in the root level is for developers. Since this is developer documentation, I think it should probably [live here](https://github.com/TheDuckCow/MCprep/tree/master/docs) and call it dev utils, but open to other ideas.

MCprep_addon/conf.py
```diff
@@ -36,10 +45,10 @@ class MCprepEnv:
 	def __init__(self):
 		self.data = None
 		self.json_data = None
-		self.json_path: Path = Path(os.path.dirname(__file__), "MCprep_resources", "mcprep_data.json")
-		self.json_path_update: Path = Path(os.path.dirname(__file__), "MCprep_resources", "mcprep_data_update.json")
+		self.json_path: PathLike = Path(os.path.dirname(__file__), "MCprep_resources", "mcprep_data.json")
```
TheDuckCow, line 48: Can you share more of the rational and reason for changing to pathLike for typing? Just curious what problems it solves (in some cases I understand making typing more general, but we _do_ know it should be a Path class instance)

MCprep_addon/materials/default_materials.py
```diff
@@ -20,15 +20,18 @@
...
 from .. import tracking
 from .. import util
 from . import sync
 
-from ..conf import env
+from ..conf import env, Engine
 
 
-def default_material_in_sync_library(default_material, context):
+def default_material_in_sync_library(default_material: list, context: Context) -> bool:
```
TheDuckCow, line 34: Typically for typing of lists and dicts etc in python, you'd use the `from typing import List` or `Dict` or a more general structure called `Sequence`, I think `List` would be fine here rather than the built in. These are special cases which is why the typing library has the special imports.

Technically the _reason_ for doing this is outdated since python 3.8 [per here](https://stackoverflow.com/questions/39458193/using-list-tuple-etc-from-typing-vs-directly-referring-type-as-list-tuple-etc).

Right now we'll have build errors on Blender 2.80 which uses python 3.7 (different example, but shows that we _will_ need to make this fix everywhere, to uphold the standard of support back to 2.80)

```
  File "/Users/patrickcrawford/Library/Application Support/Blender/2.80/scripts/addons/MCprep_addon/materials/generate.py", line 251, in PrepOptions
    passes: dict[str]
TypeError: 'type' object is not subscriptable
```

In this example, we should use `from typing import Dict` and then use `Dict[str, Any]` or something like that.

TheDuckCow (changes requested):
Didn't go all the way through, but did notice we have breakage on blender 2.80 with some of the typing used. TL;DR will need to replace any list, dict, tuple's with the variants from the `typing` module. Might be easier if you manage to test with a version of blender 2.80 on your machine, as startup errors tell you right away where anything may still be a problem :)

MCprep_addon/README.md
zNightlord: Okay I rename and move it there for now.

MCprep_addon/conf.py
zNightlord: That was me being indecisive. I made it like that for later change. So change back to Path.

zNightlord:
I think the List situation is good now, unless I miss some.

Is it just me or mcprep_ui has some issues with register properties? 
I have to comment out `default=addon_prefs.<attr>` line for it to stop prompting error

MCprep_addon/spawner/effects.py
```diff
@@ -1038,10 +1047,7 @@ def effects_enum(self, context):
 			elist.append((
 				str(effect.index),
 				effect.name + short_type,
-				"Add {} {} from {}".format(
-					effect.name,
-					display_type,
-					os.path.basename(effect.filepath))
+				f"Add {effect.name} {display_type} from {os.path.basename(effect.filepath)}"
```
StandingPadAnimations, line 1050: To be honest, I think for this specific use case, I prefer having the old `.format` syntax. It allows us to put the variables on their own lines, which tends to be nicer in editors

MCprep_addon/spawner/meshswap.py
```diff
@@ -742,9 +745,7 @@ def execute(self, context):
 			instancing = sum(t3s) - sum(t2s)
 			cleanup = t5 - t4
 			total = tprep + loop_prep + face_process + instancing + cleanup
-			env.log("Total time: {}s, init: {}, prep: {}, poly process: {}, instance:{}, cleanup: {}".format(
-				round(total, 1), round(tprep, 1), round(loop_prep, 1),
-				round(face_process, 1), round(instancing, 1), round(cleanup, 1)))
+			env.log(f"Total time: {round(total, 1)}s, init: {round(tprep, 1)}, prep: {round(loop_prep, 1)}, poly process: {round(face_process, 1)}, instance:{round(instancing, 1)}, cleanup: {round(cleanup, 1)}")
```
StandingPadAnimations, line 748: As I've stated in an earlier comment, `.format` would work better here as it allow us to split these variables on their own lines. This line is now so long that on GitHub's split diff, it has to be wrapped twice

MCprep_addon/spawner/mobs.py
```diff
@@ -444,19 +449,17 @@ def execute(self, context):
 						os.mkdir(dst)
 					except OSError as exc:
 						if exc.errno == errno.EACCES:
-							print("Permission denied, try running blender as admin")
```
StandingPadAnimations, line 447: I'm curious why this was removed? I would reword the print as "Permission denied, verify that you have sufficient permissions to access this file" (since running Blender as admin/root is a terrible idea), but removing this just doesn't make sense.

StandingPadAnimations (changes requested):
Good overall, just need to make some changes. I think the big thing is trying to know when to use f-strings vs the old `.format` syntax. In a couple of cases, there were some `.format` strings that, when converted to f-strings, were way too long to be displayed properly in an editor. 

In general, if the resulting f-string is extremely long and requires scrolling/word wrap do display, then it's better to use `.format` as you can split the different variables in individual lines. This is especially true if we also have to perform operations on those variables for the string.

TheDuckCow (changes requested):
Are you able to run the tests? At a minimum with a bulk change like this, needs to compile and run without errors - on trying to load this branch with a fresh pull just now, I get console printouts (this is fatally killing the addon):

```
    from .spawner import meshswap
  File "/Users/patrickcrawford/Library/Application Support/Blender/3.5/scripts/addons/MCprep_addon/spawner/meshswap.py", line 748
    			env.log(f"Total time: {}s, init: {}, prep: {}, poly process: {}, instance:{}, cleanup: {}".format(
    			                                                                                          ^
SyntaxError: f-string: empty expression not allowed
```

TheDuckCow:
> In general, if the resulting f-string is extremely long and requires scrolling/word wrap do display, then it's better to use `.format` as you can split the different variables in individual lines. This is especially true if we also have to perform operations on those variables for the string.

I agree with this comment, and I see you made a few of these changes already, thanks for that!

TheDuckCow:
Please also do a git merge from dev, so you have the latest files already in your branch, so when it comes to merging to dev, we have confidence it is going to work well.

MCprep_addon/conf.py
```diff
@@ -19,6 +19,19 @@
...
 import bpy
+from bpy.utils.previews import ImagePreviewCollection
+
+from typing import Union, Literal, Tuple, List, Dict
+from mathutils import Vector
+
+VectorType =  Union[Tuple[float, float, float], Vector]
+
+Form = Literal["mc", "mineways", "jmc2obj"]
+# Eevee Next now exists as an engine option in Blender code
```
StandingPadAnimations, line 30: I would suggest changing the comment to the following. It better justifys the reason for having `BLENDER_EEVEE_NEXT` as an option.
```suggestion
# EEVEE Next is the next generation EEVEE. So in preperation for that, 
# we've added "BLENDER_EEVEE_NEXT" as an Engine option
```

MCprep_addon/conf.py
```diff
@@ -82,9 +94,9 @@ def __init__(self):
 		# no blend files etc), then it would continue to ask to reload spanwers.
 		self.loaded_all_spawners: bool = False
 
-		self.skin_list: list = []  # each is: [ basename, path ]
-		self.rig_categories: list = []  # simple list of directory names
-		self.entity_list: list = []
+		self.skin_list: List[Skin] = []  # each is: [ basename, path ]
```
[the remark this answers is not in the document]
StandingPadAnimations: Personally I think this is fine. Python doesn't perform any static analysis at runtime (this is mostly for IDEs) and it's not like Swift where the compiler will just give up if it takes too long.

Source for that latter statement (from one of my friends who uses Swift):

MCprep_addon/conf.py
```diff
@@ -140,7 +153,7 @@ def icons_init(self):
 				'IMAGE')
 		except Exception as e:
 			self.log("Old verison of blender, no custom icons available")
-			self.log("\t" + str(e))
+			self.log(f"\t {e}")
```
StandingPadAnimations, line 156: The `\t` isn't necessary and will most definitely cause problems with terminal output
```suggestion
			self.log(e)
```

MCprep_addon/materials/default_materials.py
```diff
@@ -39,7 +43,7 @@ def default_material_in_sync_library(default_material, context):
 	return False
 
 
-def sync_default_material(context, material, default_material, engine):
+def sync_default_material(context: Context, material: Material, default_material: str, engine: Engine) -> Union[Material, str, None]:
```
StandingPadAnimations, line 46: `Union[Type, None]` should be replaced with [`Optional[Type]`](https://docs.python.org/3.7/library/typing.html#typing.Optional) as it's more concise. For this case, I would suggest the following (not just for this, but future cases too):
```suggestion
def sync_default_material(context: Context, material: Material, default_material: str, engine: Engine) -> Optional[Union[Material, str]]:
```

`Optional[Union[Material, str]]` states to the programmer "This function can return either a Material or a string, but it can also return a Nonetype

MCprep_addon/spawner/effects.py
```diff
@@ -31,7 +37,9 @@
 
 from . import spawn_util
 
-from ..conf import env
+from ..conf import env, VectorType
+
+ListEffectsAssets = TypeVar("ListEffectsAssets") # Check spawn_util.py ListEffectsAssets
```
StandingPadAnimations, line 42: Put the comment before the line, otherwise it'll be too long in an editor. I would also expand this comment a little
```suggestion
# Check spawn_util.py for the
# definition of ListEffectsAssets
ListEffectsAssets = TypeVar("ListEffectsAssets")
```

MCprep_addon/spawner/entities.py
```diff
@@ -127,10 +129,10 @@ def updateEntityList(context):
...
 class face_struct():
 	"""Structure class for preprocessed faces of a mesh"""
-	def __init__(self, normal_coord, global_coord, local_coord):
-		self.n = normal_coord
-		self.g = global_coord
-		self.l = local_coord
+	def __init__(self, normal_coord: VectorType, global_coord: VectorType, local_coord: VectorType):
+		self.n: VectorType = normal_coord
+		self.g: VectorType = global_coord
+		self.l: VectorType = local_coord
```
StandingPadAnimations, line 135: This should be a `dataclass` now that we can use them
```py
@dataclass
class face_struct():
	n: VectorType
	g: VectorType
	l: VectorType
```

I would also personally make the class be defined as the following to follow naming convention:
```py
@dataclass
class FaceStruct:
	normal_coord: VectorType
	global_coord: VectorType
	local_coord: VectorType
```

MCprep_addon/spawner/item.py
```diff
@@ -19,8 +19,13 @@
 import os
 
 import bpy
+from bpy.types import (
+  Context, Object
+)
```
StandingPadAnimations, line 24: Just import `Context` as `Object` on its own gets confusing (for a reason I'll get to later)
```suggestion
from bpy.types import Context
```

MCprep_addon/spawner/item.py
```diff
@@ -84,22 +89,21 @@ def reload_items(context):
...
-			"item-{}".format(i), item_file, 'IMAGE')
+			f"item-{i}", item_file, 'IMAGE')
 
 	if mcprep_props.item_list_index >= len(mcprep_props.item_list):
 		mcprep_props.item_list_index = len(mcprep_props.item_list) - 1
 
 
-def spawn_item_from_filepath(
-	context, path, max_pixels, thickness, threshold, transparency):
+def spawn_item_from_filepath(context: Context, path: Path, max_pixels: int, thickness: float, threshold: float, transparency: bool) -> Tuple[Optional[Object], Optional[str]]:
```
StandingPadAnimations, line 106: Replace `Object` with `bpy.types.Object`. Normally I would prefer the former, but `Object` can be confused with the `Object` type in Python (I actually had to do a double take on this for this very reason)
```suggestion
def spawn_item_from_filepath(context: Context, path: Path, max_pixels: int, thickness: float, threshold: float, transparency: bool) -> Tuple[Optional[bpy.types.Object], Optional[str]]:
```

MCprep_addon/spawner/mcmodel.py
```diff
@@ -20,16 +20,25 @@
 import json
 from mathutils import Vector
 from math import sin, cos, radians
+from pathlib import Path
 
 import bpy
 import bmesh
+from bpy.types import (
+  Context, Object, Material
+)
```
StandingPadAnimations, line 29: Again as suggested earlier, ditch the use of `Object` on its own and replace it with `bpy.types.Object`
```suggestion
from bpy.types import (
  Context, Material
)
```

MCprep_addon/spawner/mcmodel.py
```diff
@@ -208,29 +217,29 @@ def read_model(context, model_filepath):
...
+	env.log(f"\nfile: {model_filepath}", vv_only=True)
 	# env.log("parent:" + str(parent))
 	# env.log("elements:" + str(elements))
 	# env.log("textures:" + str(textures))
 
 	return elements, textures
 
 
-def add_model(model_filepath, obj_name="MinecraftModel"):
+def add_model(model_filepath: Path, obj_name: str="MinecraftModel") -> Object:
```
StandingPadAnimations, line 242: To prevent confusion with the Python `Object` type
```suggestion
def add_model(model_filepath: Path, obj_name: str="MinecraftModel") -> bpt.types.Object:
```

MCprep_addon/spawner/meshswap.py
```diff
@@ -23,11 +23,15 @@
 import time
 
 import bpy
+from bpy.types import (
+  Context, Object, Collection
+)
```
StandingPadAnimations, line 28: Same thing as mentioned earlier
```suggestion
from bpy.types import (
  Context, Collection
)
```

MCprep_addon/spawner/meshswap.py
```diff
@@ -159,10 +163,10 @@ def updateMeshswapList(context):
...
 class face_struct():
 	"""Structure class for preprocessed faces of a mesh"""
-	def __init__(self, normal_coord, global_coord, local_coord):
-		self.n = normal_coord
-		self.g = global_coord
-		self.l = local_coord
+	def __init__(self, normal_coord: VectorType, global_coord: VectorType , local_coord: VectorType) -> None:
+		self.n: VectorType = normal_coord
+		self.g: VectorType = global_coord
+		self.l: VectorType = local_coord
```
StandingPadAnimations, line 169: Make this a dataclass
```py
@dataclass
class face_struct():
	n: VectorType
	g: VectorType
	l: VectorType
```

MCprep_addon/spawner/meshswap.py
```diff
@@ -754,10 +759,10 @@ def execute(self, context):
 		elif self.runcount == 1:
 			self.report({'INFO'}, "Swapped 1 object")
 			return {'FINISHED'}
-		self.report({'INFO'}, "Swapped {} objects".format(self.runcount))
+		self.report({'INFO'}, f"Swapped {self.runcount} objects")
 		return {'FINISHED'}
 
-	def prep_obj_list(self, context):
+	def prep_obj_list(self, context: Context) -> List[Object]:
```
StandingPadAnimations, line 765: Replace `Object` with `bpy.types.Object`
```suggestion
	def prep_obj_list(self, context: Context) -> List[bpy.types.Object]:
```

MCprep_addon/spawner/spawn_util.py
```diff
@@ -20,13 +20,21 @@
 import re
 
 import bpy
+from bpy.types import (
+  Context, Object, Collection,
+  BlendDataLibraries
+)
```
StandingPadAnimations, line 26: Like mentioned earlier
```suggestion
from bpy.types import (
  Context, Collection, BlendDataLibraries
)
```

MCprep_addon/spawner/spawn_util.py
```diff
@@ -227,7 +234,7 @@ def attemptScriptLoad(path):
 	text.use_module = True
 
 
-def fix_armature_target(self, context, new_objs, src_coll):
+def fix_armature_target(self, context: Context, new_objs: List[Object], src_coll: Collection) -> None:
```
StandingPadAnimations, line 237:
```suggestion
def fix_armature_target(self, context: Context, new_objs: List[bpy.types.Object], src_coll: Collection) -> None:
```

MCprep_addon/spawner/spawn_util.py
```diff
@@ -338,11 +345,11 @@ def get_rig_from_objects(objects):
 	return prox_obj
 
 
-def offset_root_bone(context, armature):
+def offset_root_bone(context: Context, armature: Object) -> bool:
```
StandingPadAnimations, line 348:
```suggestion
def offset_root_bone(context: Context, armature: bpy.types.Object) -> bool:
```

MCprep_addon/world_tools.py
```diff
@@ -19,13 +19,17 @@
 import os
 import math
 from pathlib import Path
+from typing import List, Optional
 import shutil
 
 import bpy
+from bpy.types import (
+  Context, Object, Camera
+)
```
StandingPadAnimations, line 28: Just like with past suggestions
```suggestion
from bpy.types import (
  Context, Camera
)
```

MCprep_addon/world_tools.py
```diff
@@ -993,7 +999,7 @@ def execute(self, context):
 		self.track_param = engine
 		return {'FINISHED'}
 
-	def create_sunlamp(self, context):
+	def create_sunlamp(self, context: Context) -> Object:
```
StandingPadAnimations, line 1002: To prevent confusion
```suggestion
	def create_sunlamp(self, context: Context) -> bpy.types.Object:
```

MCprep_addon/world_tools.py
```diff
@@ -1182,7 +1188,7 @@ def cleanup_scene(self):
 		if self.open_folder:
 			bpy.ops.mcprep.openfolder(folder=self.filepath)
 
-	def create_panorama_cam(self, name, camera_data, rot, loc):
+	def create_panorama_cam(self, name: str, camera_data: Camera, rot: VectorType, loc: VectorType) -> Object:
```
StandingPadAnimations, line 1191: Again to prevent confusion
```suggestion
	def create_panorama_cam(self, name: str, camera_data: Camera, rot: VectorType, loc: VectorType) -> bpy.types.Object:
```

StandingPadAnimations (changes requested):
Finished reviewing. There's a couple of things related to annotations and comments.

For comments, make sure they aren't long, so try to keep them on their own lines. In addition, comments should be clear. Instead of `Check <file> <type>`, do `Check <file> for definition of <type>` (preferably broken up so things don't get too long).

For type annotations, be as concise and clear as possible. If a bpy type can get confused with a Python type, then make it clear it's a bpy type (like doing `bpy.types.Object` over `Object`). On that same note, `Optional[Union[Type1, Type2]]` is more concise over `Union[Type1, Type2, None]` and conveys the same information. Try using dataclasses where they make sense too, it's less boilerplate then writing a class with a `__init__` function that only holds variables.

StandingPadAnimations:
> Please also do a git merge from dev, so you have the latest files already in your branch, so when it comes to merging to dev, we have confidence it is going to work well.

Personally if it's to keep a branch up to date, I think a rebase would make more sense (merges should be reserved for feature branches being merged upstream, not just to update a branch), but that brings up the whole merge vs rebase debate

zNightlord:
I was a bit confused by entities.py has another `FaceStruct`, so it is removed.

TheDuckCow:
> Personally if it's to keep a branch up to date, I think a rebase would make more sense (merges should be reserved for feature branches being merged upstream, not just to update a branch), but that brings up the whole merge vs rebase debate

It depends on how you do it, I've made the mistake of using a rebase where it should have been a git merge and results in merge conflicts which basically duplciates commits in the history. It's worth calling out that when pressing "resolve conflicts" in the actual github interface, it is the equivalent to running: `git checkout annotate-documment; git merge dev` which leaves both branches in place but just brings any new changes in the history of `dev` into the chain of `annotate-documment`

@zNightlord and I should actually say, the could should be to merger `theduckcow:milestone-3-5-0` into your local branch, so that I have the fixes to the build script so I don't borq my addons folder again.

Please also do a check where you run and install MCprep with Blender 2.80 - I think some annotations might not be valid back that far.

TheDuckCow:
I'd like to get this merged in soon, and avoid having it drag on too much more! Thanks for putting up for the back and forth feedback here, moving forward I'll suggest we try and break changes into multiple separate parts to avoid these sorts of mega commits of 1K+ lines (refactoring will always be large, but there are ways to break it up).

TheDuckCow:
Just checking @zNightlord do you have time to make updates in the next week and a half (before the dev meeting kickoff)? If not, I can try putting some time aside to fix issues I've run into with older blender versions. You might still need to add me as a direct collaborator on your fork though.

zNightlord:
Alright, `Literal` did not exist in py3.7 so using typing extension instead and for `NodesModifier` that is for 3.0  

There are some lines like `default=addon_prefs.skin_path` in mcprep_ui causing some register problem. I'm not really sure why but comment it out, annotate is good with 2.80.

TheDuckCow:
Thanks for making the updates, I'm going through now also trying to see what is causing that line to fail. Issues during property registration are usually somewhat opaque and hard to track down as it's a sign of an issue somewhere else, but I'll let you know what I come across.

TheDuckCow:
FYI I pushed one commit @zNightlord, which makes it work quite a bit better for me now. But now when I load on blender 2.80, I'm getting this error - are you testing on 2.80 and having it work? I get this as soon as I try to enable the addon.

```
  File "/Applications/Blender 2.80/blender.app/Contents/Resources/2.80/scripts/modules/addon_utils.py", line 351, in enable
    mod = __import__(module_name)
  File "/Users/patrickcrawford/Library/Application Support/Blender/2.80/scripts/addons/MCprep_addon/__init__.py", line 59, in <module>
    from . import load_modules
  File "/Users/patrickcrawford/Library/Application Support/Blender/2.80/scripts/addons/MCprep_addon/load_modules.py", line 28, in <module>
    from . import conf
  File "/Users/patrickcrawford/Library/Application Support/Blender/2.80/scripts/addons/MCprep_addon/conf.py", line 28, in <module>
    from typing_extensions import Literal
ModuleNotFoundError: No module named 'typing_extensions'
```

For what it's worth, I have Blender 2.80 (sub 75) local.

TheDuckCow:
As a compatibility reminder, blender 2.80's python is `3.7.0` so keep in mind the feature set we have available with that.

TheDuckCow:
Doing some reading, seems like one option could be to have a `pip3 install typing_extensions` version packaged with the addon, and then do a relative import to get the typing extensions that way (I see that `typing_extensions` is referenced in some places, but we shouldn't presume or attempt to ad hoc install pip, I'd rather we just package the code with the addon itself in a modules or similar folder).

But all of that just seems horribly complicated. What if instead of using Literal anywhere, we just define python Enum values to use instead? (and utilize their own built in ability to fetch the string name, if ever we need to pass as a string)? What do you think @zNightlord?

Only files currently using Literal are: `MCprep_addon/util.py`, `MCprep_addon/conf.py`, and `MCprep_addon/materials/sequences.py` and all feel like they should easily be Enums (not exactly sure of the typing implication though)

StandingPadAnimations:
I've only used Python enums once but they make me miss C++ enums (which in turn make me miss Rust enums). I don't know what it is about them (maybe the `Class.ENUM` syntax) but I dislike working with Python enums. That being said, they're a better option in this circumstance.

zNightlord:
I'm not sure why my download in the first place somehow has typing extension in it.
I'm good with using enum option.

TheDuckCow:
Ok great - let me take a stab at making this change then, as I will likely run into some other fixes for getting things working. thanks both for weighing in!

TheDuckCow (approved):
We made it folks. After 51 comments, 38 commits, and yes over a thousand lines changed, I'm ready to approve and merge this.

Since this is just going into the milestone branch, I'm using my owner authority to go ahead and merge. I should say, while all _automated tests_ pass, I do have some errors or issues that cropped up that I'll be creating a fix list for, but the good news is Blender 2.80 is definitely working fine for the most common use cases. There are some issues affecting all versions (e.g. an issue with animate textures exists that I didn't get to). We can see about resolving these in _new, dedicated_ branches for each issue so we can split up the work. I'll also be working on re-creating the test suit anyways, to use a more standard library. This will make it easier for others to run tests and validate their own work as well.

Proof of all tests passing for me locally, noting:
- Loss of data is nothing to be overcome
- img_sequence_effect_spawner Apparently this crashing is nothing new, perhaps the feature should just be disabled for blender 2.8x.

Raw test results:

```
blender	failed_test	short_err
(3, 6, 0)	ALL PASSED	-
(3, 5, 1)	ALL PASSED	-
(3, 4, 0)	ALL PASSED	-
(3, 2, 1)	ALL PASSED	-
(3, 3, 1)	ALL PASSED	-
(3, 1, 0)	ALL PASSED	-
(3, 0, 0)	ALL PASSED	-
(2, 93, 0)	qa_effects	 loss of data!\n \n 
(2, 93, 0)	qa_rigs	 loss of data!\n \n 
(2, 80, 75)	img_sequence_effect_spawner	 consistent crashing
(2, 80, 75)	qa_effects	 loss of data!\n \n 
(2, 80, 75)	qa_rigs	 loss of data!\n \n 
```

Issue #452, opened 2023-07-25
Code to manually offset bmesh's scaling problem. It's messy in principle but it works perfectly.


FigureOfStickman:
### bmesh scale fix

Hi! I hope this isn't inappropriate. I've never really used GitHub before.

I added this bit of code in my copy of the addon to rescale spawned blocks so that they aren't 1.008m^3. It's been working great so far, so I thought you guys may want to check it out. It's been bothering my adderall-riddled brain and [StandingPadAnimations said it was a known issue.](https://github.com/Moo-Ack-Productions/MCprep/issues/451#issuecomment-1648790199)

Here's a screenshot of the code I added to [mcmodel.py](https://github.com/Moo-Ack-Productions/MCprep/blob/a8facb4a3b8fd7f0f5d22577e4d53a57f742190a/MCprep_addon/spawner/mcmodel.py#L4) and what it looks like in Blender. I guess it automatically applies the scale because the operation is being performed on the raw mesh, so that cleans it up a bit for the user.

<img width="742" alt="moddedmcprep" src="https://github.com/Moo-Ack-Productions/MCprep/assets/54049252/9ca2bca8-0d33-43d2-9405-0c636d4c4866">
<img width="405" alt="Screenshot 2023-07-25 at 11 45 52 AM" src="https://github.com/Moo-Ack-Productions/MCprep/assets/54049252/14c78d18-61fd-4969-b3d7-663af89af727">

### Quality checks

- [X] I have used only vanilla textures (or created images that are derived from vanilla textures)
- [X] This is a vanilla minecraft asset/mob/effect
- [X] I have verified my asset has the correct scale (1 Blender unit = 1 meter), and the scale has been applied (control+a, scale)
- [X] If my asset includes a rig, I certified there is a "root" or "main" bone, and the rig object name ends with the name ".arma" or ".rig" to work well with MCprep
- [X] I am the sole creator of this rig, other than any textures by Mojang. Any custom textures I created myself.

### Describe any usage details

Not a rig, I hope it's still okay for me to submit this though.

### What name should we use to credit you in the README.md file?

FigureOfStickman

### What credit url should we use to reference you in the README.md file?

youtube.com/@figureofstickman6718

### Confirmation

[...]

Issue #451, opened 2023-07-25
Why are individually spawned blocks 1.008 meters? And is there a way to change this?


FigureOfStickman:
<img width="245" alt="Screenshot 2023-07-24 at 7 06 08 PM" src="https://github.com/Moo-Ack-Productions/MCprep/assets/54049252/bd044b44-cd3b-4f21-89b1-11cc99cefec7">

StandingPadAnimations:
Closing since this doesn't use the bug report template. That being said, this is already a known issue, but lies in Blender's `bmesh` module (what we use to generate blocks) and thus is beyond fixing

Issue #429, opened 2023-06-05
Block Spawner Dimensions Error


mguggenheim:
### Did the issue happen after restarting Blender?

- [X] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [X] I have checked known issues, and my problem is different

### What happened?

When I spawn blocks via the block spawner, they are coming in at a size of 1.008 instead of 1.

### What steps are needed to produce the bug?

Spawn any block from the block spawner.

### What version of MCprep are you using?

3.4.3

### What version of Blender are you using?

3.5.1

StandingPadAnimations:
> When I spawn blocks via the block spawner, they are coming in at a size of 1.008 instead of 1.

This is a known issue, but it seems to comes from Blender's side.

Issue #437, opened 2023-07-02
Mcprep spawning blocks with misplaced textures


shadowkat11:
### Did the issue happen after restarting Blender?

- [X] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [X] I have checked known issues, and my problem is different

### What happened?

So I wanted to start getting into Minecraft animation, so I got mcprep. It started off fine when I was just posing my skin, but when I wanted to start animating scenes, and placed blocks like a crafting table or a grass block (which all have multiple textures instead of just a one texture block like a wood plank), they came out misplaced. I've tried lots of things, including redownloading blender and re-installing mcprep, but couldn't get it fixed.  
Any way to fix this?
![little texture bug](https://github.com/TheDuckCow/MCprep/assets/138401185/23be9a7d-28c9-479e-92af-4d177036199e)

### What steps are needed to produce the bug?

All that is needed to make it (for me) is to find a block with multiple textures (ex. grass, it has the sides, top, and bottom) and to place it, as the textures will be in the incorrect places.

### What version of MCprep are you using?

Mcprep 3.4.3

### What version of Blender are you using?

Blender 3.6

StandingPadAnimations:
This is a known issue, we're not sure what causes it

shadowkat11:
alr, thanks for the answer anyway.

PR #453, opened 2023-07-31
Model Json tweak

zNightlord:
The 1.008 dimension issue as mentioned in #429  #451 #452 
Change the behavior of model spawner spawning in just one vertex, this should be a warning of invalid json file instead.


zNightlord:
a fix for #437

MCprep_addon/spawner/mcmodel.py
```diff
@@ -70,6 +70,16 @@ def rotate_around(
 		(new_pos[1] - offset[1]) * scale[1]
 	))
 
+def getPlanarAxis(verts: List[VectorType]) -> int:
+	"""Return the axis of the face"""
```
StandingPadAnimations, line 74:
```suggestion
"""
Given a set of verticies, return the axis of the face

Returns:
    An integer representing the axis
"""
```

StandingPadAnimations (changes requested):
I know this is still a WIP, but I've reviewed the code and commented the things that need to be looked at. Most of them deal with comments and error messages.

Comments shouldn't be vague. Phrases such as "For some reason" don't help other developers understand why something is the way it is. Comments should be clear, not vague.

zNightlord:
I just put random things as notes for myself. 
The "west" face gets appended twice, how matter I check the loop it just appends the last one to the first index so. Have no idea what cause it.
The cake is a json file issue, they just don't add UV coord inside the file.
The face planar check is still WIP for optimizing the model and reuse material is still not there yet.
For model like bed and banner ... in #267 using meshswap instead is there, but it is a bit messy. 
And also adding those models into meshswap blend file

TheDuckCow:
Just noticing this is in draft @zNightlord - please flip to "ready for review" if and when you'd like me to live test!

TheDuckCow:
Oops, sorry @zNightlord  for some reason merging the milestone branch didn't just promote this PR to be based on dev, it closed :/ and it won't let me reopen. You'll need to re-open a new PR just targeting `dev` this time. Sorry about that!

PR #445, opened 2023-07-11
i18n: add zh_CN for test

chisaato:
First step of #439

Issue #439, opened 2023-07-06
The Beginning of i18n Support


StandingPadAnimations:
Currently MCprep only supports English as a language option, but Minecraft animation is a much much more diverse field. As such, we need add some system to MCprep to allow i18n support. i18n support was proposed back in #165, and now it's time we start thinking about it, especially as MCprep's new architecture now makes it easier to add. Our goal is the following:
> "Let no Minecraft animator be confused by MCprep's UI simply because of the language they speak"

~~As such, this will be marked as high priority for MCprep 3.5.1, and the lack of basic i18n support should be considered a blocker for the MCprep 3.5.1 release.~~
EDIT: as per #481, I've pushed this to MCprep 3.6, so this is no longer a 3.5.1 blocker 

Ideally we'd want to use the system proposed [here](https://github.com/TheDuckCow/MCprep/issues/165#issuecomment-1481324418), to make it as easy as possible for users to create translations.

To begin i18n support, we need the following:
- [x] Ability to load JSON files at startup
- [x] Migrate MCprep's panels to use the setup [as defined in this guide](https://b3d.interplanety.org/en/localization-of-blender-add-ons-with-api/)
- [x] Create a system for users to add their own translations in a way that makes sure Google Translate and other AI translators aren't used (especially in non-European languages, AI translators tend to do a poor job at translating)
- [ ] Document the JSON format

StandingPadAnimations:
Whoops, sorry for the blank issue, let me add some description

StandingPadAnimations:
Added a basic mechanism of loading language files. This also adds the option of not just supporting full languages, but also dialects.
https://github.com/TheDuckCow/MCprep/commit/bc852c4e8dd324ca1153dbbac2382cefb53cfb01

https://github.com/TheDuckCow/MCprep/blob/bc852c4e8dd324ca1153dbbac2382cefb53cfb01/MCprep_addon/conf.py#L68-L87

StandingPadAnimations:
Alright I'm considering [this format](https://b3d.interplanety.org/en/localization-of-blender-add-ons/) since it seems to be a little easier to understand, but it doesn't update everything on the fly (some things require a restart).

StandingPadAnimations:
> First, it's strongly recommend to have the identitiy key for every field. Here is the thing, when we start translate, we will change every original text, so keep an identitiy key can help us know the association between original text and translated text.

[...]

PR #458, opened 2023-08-17
i18n Support: PR 1

StandingPadAnimations:
> \- Goal defined in #439 

PR 1 will be merged for MCprep 3.5.1. While not explicitly defined as a blocker for the 3.5.1 release, it will be treated as such unless decided otherwise. 

As of making PR 1, a good chunk of the MCprep N panel is translatable, with a Simplified Chinese translation on the way (#445). Since I have to juggle between this and https://github.com/Moo-Ack-Productions/bpy-build, a second developer would be appreciated. 

Internally, every string is represented by an ID. At runtime, MCprep calls `MCprepEnv.translate_str(ID)`, which takes an enum value defined in `MCprep_addon/translate_enum.py` and maps it to the string defined in JSON. If no such string exists, English is used as a fallback.


StandingPadAnimations:
As per #481, I'll be moving this to MCprep 3.6. Also gives us more time to refine the initial architecture

MCprep_addon/translate_enum.py
TheDuckCow, line 1: See comment here for additional keys we'd want to have, not quite exhaustive but gets a number of them

https://github.com/Moo-Ack-Productions/MCprep/pull/445#issuecomment-1806179667

TheDuckCow (changes requested):
Was otherwise about to approve, but could we rename `env.translate_str()` calls to be `env.tr()`? Something I love from godot is how succinct this translate string function is, since it'll be basically everywhere. That's the only reason for my request for changes here.

With my demo [over here](https://github.com/Moo-Ack-Productions/MCprep/pull/445), I can see it working well and I see no reason to push the first languages through, even if incomplete. I made a comment below about some missing keys, but we could also do this in a future branch.

StandingPadAnimations:
`env.tr()` is a bit too short, I think `env.tr_str()` or `env.tr_s()` would be fine. However, at the moment, I'm not working on this PR, so someone else would need to do that final push to the end.

StandingPadAnimations:
Also, this PR focuses on the N panel and basic infrastructure, so I'm not too concerned about operators rigth now

TheDuckCow:
I'm fine to take over for you, thanks for getting it this far! And sure - I can settle on `tr_s`.

Understood it focussed on the N panel, I was seeing the operator names poke through. I might expand on it a bit more.

StandingPadAnimations:
I do have one concern with the current JSON format, that being split lines.

At the moment, we have separate keys for split lines, based on how many lines it requires in English to fit a long sentence into the MCprep UI, such as the following:
https://github.com/Moo-Ack-Productions/MCprep/blob/606084f0d791febb690e9ece815204ff54335149/MCprep_addon/MCprep_resources/Languages/en_US.json#L26-L27 

However, other languages may require more or less, so I think we should change it to something like this:
```py
	"world_add_world" : [
		"No time controller,",
		"add dynamic MC world."
	]
``` 

And use a for loop in the UI for IDs like these. It's more work, but should help with preventing the UI from looking bad with languages other than or similar to English.

StandingPadAnimations:
> I'm fine to take over for you, thanks for getting it this far!

Thanks, I'll move this back into the goals for MCprep 3.5.1 then

StandingPadAnimations:
Ok, I just discovered that the Python standard library has a module for i18n that allows the use of standard POT files (10 times easier to work with in this case)
- https://docs.python.org/3/library/gettext.html
- https://lokalise.com/blog/beginners-guide-to-python-i18n/

In other words, the past 13 commits are no longer needed, and we should probably redo the entire PR to use standard convention

*sigh*

StandingPadAnimations:
Moving this to MCprep 3.6. This will give some extra time to work on this

StandingPadAnimations:
^ How I feel now

The way I see it, we have 2 options:
1. Discard this branch and PR, and make a new branch using `gettext` from Python's standard library
    - Would be time consuming (especially as I enter spring semester 🥲), but [looking at this comment](https://github.com/Moo-Ack-Productions/MCprep/issues/439#issuecomment-1627327879), we can try to salvage what we have, so it's not all for nothing
    - In the end, this route would make it easier for translators, and we could use something like [Weblate](https://weblate.org/en/) too (though that's a problem for future us to work out) 
2. Continue with what we have and move on

StandingPadAnimations:
I'll opt with option one, closing this PR and starting a new one

Issue #457, opened 2023-08-15
Importing mods models


BoBikDied:
### Did the issue happen after restarting Blender?

- [X] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [X] I have checked known issues, and my problem is different

### What happened?

When attempting to import a JSON model from mods (specifically from the mod "Create"), the following error was displayed:
https://drive.google.com/file/d/1lkH1f8-m7b3FOLLsRjINtaaXeNKfC89C/view?usp=sharing

### What steps are needed to produce the bug?

1) File - Import - .json
2)Import .json model from mod

### What version of MCprep are you using?

MCprep 3.4.3

### What version of Blender are you using?

I tried to do this on different versions of Blender(3.3 - 3.6)

StandingPadAnimations:
From the error, it seems like the texture couldn't be found in the default resource pack (all vanilla textures). Modded blocks aren't supported at the moment, unless you add their textures with the proper name into the MCprep resource pack folder.

zNightlord:
Beside the mod stuffs, this also addressed something else.
I think it shouldn't raise that texture exception, since this is from import from json *2 operators shared the same code also the same add_material* before #398 it didn't raise anything *as far as i remember it* and let the user manually assign the textures, unless we treat this unfindable/ empty texture as a bug rather a feature.

StandingPadAnimations:
The exception is raised because MCprep checks textures based on block name, which in turn is derived from material name (there's also a JSON file for material names to blocks):
https://github.com/Moo-Ack-Productions/MCprep/blob/master/MCprep_addon/materials/generate.py#L116-L123

As far as I can tell, JSON models should define textures anyway, so no need to mess with all global textures (which is overwritten during updates). Modded blocks however will always be a weird area because there's simply too many mods, so I think it's best if we don't support modded blocks explicitly (any support MCprep appears to have will entirely be unintentional)

zNightlord:
> The exception is raised because MCprep checks textures based on block name, which in turn is derived from material name (there's also a JSON file for material names to blocks):
> 
> https://github.com/Moo-Ack-Productions/MCprep/blob/master/MCprep_addon/materials/generate.py#L116-L123
> 
[...]

Issue #267, opened 2022-01-02
Block/model spawning issues


TheDuckCow:
With v3.3.0 just released, there are a few known issues already with the model spawner. Rather than wait until every single obscure block is functioning, the update instead comes a bit sooner but with some known issues. There may be more blocks than these that don't immediately work. *There are likely more than the below, these are just the known ones so far*

**Blocks that spawn apparently without any geometry**
  - beds
  - banner
  - barrel
  - shulker box
  - moving piston
  - chest
  - ender chest

**Blocks that spawn with at least some geo, but have other artifacts:**
  - Bell between walls (just the bar, no bell)
  - Cake kiinda works.. but weird, 'exploded' view
  - Brewing stand technically works.. it's just split into multiple pieces :/
  - Enchanting table (does spawn, but obv. no book)
  - Hopper: spawns, but something seems off about textures.
  - Cauldron (similar issue to hopper on top it seems)
  - Scaffolding (similar issue to hopper, but the "legs")
  - Sunflower just has z-fighting, likely true for any flower style model. 
  - End_portal
  - Composter has messed up UVs

**Blocks that result in errors on spawn:**
 - cube_all (likely should be removed anyways, as a meta-constructor)

**Materials could be auto-animated where relevant:**
 - Seagrass (animate textures doesn't work after the fact evidently)

Because the generation is entirely procedural from json files, these may take some time to fix

Zachadactyl:
Yes, the cauldron definitely has issues, I use BSS Cycles Edit for the cauldron instead.  I tried UV mapping but got lost whoops.

TheDuckCow:
There are some major improvements needed for this operator. I was tempted to move the feature into an experimental flag, but since it's been out there long enough and enough people are using it, I'm going to keep it as is. Some extra try catching has been added, but major issues like UV issues still remain. I'm going to move this bug report into a later milestone, but noting a large number of reported bugs will now be resolved thanks to the extra try/catch and list check before referencing code added previously.

I'm also hoping to add a way to capture what kinds of blocks people are attempting to generate to also get a sense where issues occur, possibly a mini survey asking if the block was generated correctly or not. I will not hold up releases on this however, and would prefer to hear from users on this thread (yes, anyone can reply!) with what they tried to spawn, attaching json files if appropriate (ie if not coming from the MCprep built in ones)

TheDuckCow:
[...]

Issue #148, opened 2020-08-28
Entities (Custom/Other Models) spawner category


TheDuckCow:
This would be a place in MCprep, likely under the spawner panel, which would be used to spawn additional models/rigs that don't fit into the existing mob / item / block categories. This could be implemented as just a "Models" category under mob spawner, but this would not be that intuitive given these are not mobs. It could also be another section alongside Mobs, (Meshswap) Blocks, and Items. 

Ideas of things to include (assumption is all of these would be custom, 3D models / rigs):

- [x] Rigged bows and crossbows
- [x] Boats with animatable oars
- [x] Minecarts, maybe with animatable wheels and toggles for Furnace, TNT, Hopper, and Chest
- [ ] Rigged book, with page turning and enchanted material toggle
- [x] Signposts with text ready for editing
- [ ] Item stand (possibly rigged)
- [ ] Rigged piston (this probably would be better suited as a meshswap block though)
- [x] Arrows
- [ ] Bowls
- [ ] Potions
- [ ] Trident
- [ ] Elytra
- [ ] Fishing rod rig

_See [more here](https://minecraft.fandom.com/wiki/Entity)_

KittenCat1080ti:
From what I understand, In order to do this, I need to add a few classes to Mcprep_ui.py something similar to the Item spawn class? My first goal is to make the ui appear

KittenCat1080ti:
![image](https://user-images.githubusercontent.com/71938947/96373266-5746d500-1139-11eb-93e8-c5fc92c444a6.png)
I Have added the draw class for the new button (I reused the Item code a bit) this code unfortunately, broke the whole MCprep panel & now it doesn't even appear, probably because of a bug in my code

KittenCat1080ti:
Ok, I fixed that

KittenCat1080ti:
Now It it almost works

TheDuckCow:
Thanks for working on this! One thing to ask - should we call this "models" or "entities"? I think entities might make more sense, to make it more obviously distinct from what is included in the meshswap file. ie, entities will be things that do not appear as blocks in world imports.

And to map out the key next steps here (not sure where you're at now):

- The UI is one piece, looks like you are on your way with this one!
[...]

PR #469, opened 2023-08-25
Model JSON tweaks

zNightlord:
Previously #453 

This should take care of that issue in #457, I hope. 
~~Need to find a nice way to check the namespace "minecraft" vanilla, no longer since there are cases can't really find minecraft vanilla texture~~
Use default no diffuse texture setup instead.


zNightlord:
Should be ready for review now.

zNightlord:
Should we move enum into one file because those are scattering everywhere now?

MCprep_addon/materials/generate.py
```diff
@@ -29,6 +30,18 @@
 
 AnimatedTex = Dict[str, int]
 
+class PackFormat(Enum):
+	def _generate_next_value_(name, start, count, last_values):
+		return name
+	
+	SIMPLE = auto()
+	SEUS = auto()
+	SPECULAR = auto()
```
StandingPadAnimations, line 39: Do not ever use `auto` for enums, use integers.

```suggestion
	SIMPLE = 1
	SEUS = 1
	SPECULAR = 2
```

MCprep_addon/materials/generate.py
```diff
@@ -1112,6 +1125,55 @@ def texgen_seus(mat: Material, passes: Dict[str, Image], nodeInputs: List, use_r
...
+			# The offset and link diffuse is for default no texture setup
+			links = mat.node_tree.links 
+			principled = None
+			for n in nodes:
+				if n.bl_idname == 'ShaderNodeBsdfPrincipled':
+					principled = n 
+					break
+			
+			links.new(node_diff.outputs[0], principled.inputs[0])
+			links.new(node_diff.outputs[1], principled.inputs["Alpha"])
```
StandingPadAnimations, line 1160: The variable here is redundant in this case, linking can be done in the actual for loop itself, and the variable is only used for links anyway.

 ```suggestion
			for n in nodes:
				if n.bl_idname == 'ShaderNodeBsdfPrincipled':			
			        links.new(node_diff.outputs[0], principled.inputs[0])
			        links.new(node_diff.outputs[1], principled.inputs["Alpha"])
			        break
```

File: MCprep_addon/materials/prep.py [hunk 3]
```diff
@@ -247,13 +247,13 @@ def execute(self, context):
...
-					use_principled=self.usePrincipledShader,
-					only_solid=self.makeSolid,
-					pack_format=self.packFormat,
-					use_emission_nodes=self.useEmission,
-					use_emission=False  # This is for an option set in matprep_cycles
+					passes, 
+					self.useReflections, 
+					self.usePrincipledShader, 
+					self.makeSolid, 
+					generate.PackFormat.from_str(self.packFormat.upper()),
```
StandingPadAnimations, line 254: `from_str` is unnecessary, just do `generate.PackFormat[self.packformat.upper()]`. It's already a documented feature of Python.

MCprep_addon/materials/generate.py
```diff
@@ -29,6 +30,18 @@
...
+	def _generate_next_value_(name, start, count, last_values):
+		return name
+	
+	SIMPLE = auto()
+	SEUS = auto()
+	SPECULAR = auto()
+  
+	@staticmethod
+	def from_str(enum_name: str):
+			return PackFormat[enum_name]
```
StandingPadAnimations, line 43: This isn't needed, the behavior of `Enum[string]` is already documented or could be inferred.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -69,52 +69,71 @@ def rotate_around(
...
+				True, 
+				False, 
+				generate.PackFormat.SIMPLE, 
+				False, 
+				False # This is for an option set in matprep_cycles
+			)
+			res = generate.matprep_cycles(
+				mat=mat,
+				options=options
+			)
```
StandingPadAnimations, line 133: That's… a lot of uncommented code with special behavior. I'm honestly confused at what I'm looking at here. Please comment this.

TheDuckCow:
I need to do more of a review, but seeing we're getting into the nitty details of code style makes me think we're pretty close on this, and as I understand would greatly increase the reliability of being able to use the model importer (our top source of errors from users). As such, I've marked this for the 3.5.0 milestones, in the hopes we can wrap this up in the next week or two max to merge.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -238,13 +258,39 @@ def read_model(context: Context, model_filepath: Path) -> Tuple[Element, Texture
...
 	"""Primary function for generating a model from json file."""
-	mesh = bpy.data.meshes.new(obj_name)  # add a new mesh
-	obj = bpy.data.objects.new(obj_name, mesh)  # add a new object using the mesh
-
 	collection = bpy.context.collection
 	view_layer = bpy.context.view_layer
+	
+	# Forcing usage of "fixed" block to meshswap spawner instead
+	# Theses blocks certainly are already in meshswap
+	blocks = ["chest", "banner", "lantern", "campfire"]
```
StandingPadAnimations, line 268: If the user imports a JSON model of one of these blocks over using meshswap, then we **should not** decide to use the meshswap model over the JSON model. I shouldn't have to say this, but here we are.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -310,8 +346,14 @@ def add_model(model_filepath: Path, obj_name: str="MinecraftModel") -> bpy.types
 
 			uv_coords = d_face.get("uv")  # in the format [x1, y1, x2, y2]
 			if uv_coords is None:
-				uv_coords = [0, 0, 16, 16]
-
+				uv_coords = [0, 0, 16, 16] 
+				# Cake, cake slices don't have UV key in the dict so use the default one which is incorrect.
```
StandingPadAnimations, line 350:
```suggestion
				# Cake and cake slices don't store the UV keys
				# in the JSON model, which causes issues. This
				# workaround this fixes those texture issues
```

MCprep_addon/spawner/mcmodel.py
```diff
@@ -322,21 +364,30 @@ def add_model(model_filepath: Path, obj_name: str="MinecraftModel") -> bpy.types
...
 				# will be 2 then 3, 0, 1.
 				face.loops[j][uv_layer].uv = uvs[(j + uv_idx) % len(uvs)]
 
-			face_mat = d_face.get("texture")
+			# Assign the material on face
 			if face_mat is not None and face_mat in materials:
 				face.material_index = materials.index(face_mat)
+			is_first = False
 
+	# Quick way to clean the model, hopefully it doesn't cause any UV issue
```
StandingPadAnimations, line 384:
```suggestion
	# Quick way to clean the model, hopefully it doesn't cause any UV issues
```

MCprep_addon/spawner/mcmodel.py
```diff
@@ -484,7 +535,11 @@ def execute(self, context):
 			return {'CANCELLED'}
 
 		try:
-			obj = add_model(os.path.normpath(self.filepath), name)
+			r, obj = add_model(os.path.normpath(self.filepath), name)
+			if r:
+				self.report(
+					{"ERROR"}, "The JSON defined is not valid for a Minecraft Java Edition model")
```
StandingPadAnimations, line 541:
```suggestion
					{"ERROR"}, "The JSON model is not avalid JSON model as per Minecraft Java Edition standards")
```

MCprep_addon/spawner/mcmodel.py
```diff
@@ -522,7 +577,11 @@ def execute(self, context):
 			return {'CANCELLED'}
 
 		try:
-			obj = add_model(os.path.normpath(self.filepath), filename)
+			r, obj = add_model(os.path.normpath(self.filepath), filename)
+			if r:
+				self.report(
+					{"ERROR"}, "The JSON defined is not valid for a Minecraft Java Edition model")
```
StandingPadAnimations, line 583:
```suggestion
					{"ERROR"}, "The JSON model is not avalid JSON model as per Minecraft Java Edition standards")
```

StandingPadAnimations (changes requested):
Couple of issues I noticed with enums and comments. Otherwise, mostly good.

I did see an instance where MCprep will force the user to use meshswap models over JSON models. That is unacceptable, the precedent is that we do not decide on the user's behalf anything related to imports unless it fixes an actual issue (Non-standard color spaces for OBJs, feature compatibility, etc.). Nowhere does that mean, “let's ignore the user's wish to use a JSON model because we already have a meshswap model”.

zNightlord:
> Couple of issues I noticed with enums and comments. Otherwise, mostly good.
> 
> I did see an instance where MCprep will force the user to use meshswap models over JSON models. That is unacceptable, the precedent is that we do not decide on the user's behalf anything related to imports unless it fixes an actual issue (Non-standard color spaces for OBJs, feature compatibility, etc.). Nowhere does that mean, “let's ignore the user's wish to use a JSON model because we already have a meshswap model”.

The issue is certain Minecraft "block" isn't really a normal block like most of the blocks in the Block model. Most of the blocks in #267 would be called as an entity block/tile and some cases like double block flower, tall grass... would it better to be a one mesh, a

zNightlord:
I originally I would make fixed json version of those entity blocks but blend file would be better suitable for it, I'm expecting "why this shulker block, chest in block model can't do that open intersction and meshswap has another version of them that can open, interact"
*accident press the close button*

MCprep_addon/spawner/mcmodel.py
zNightlord: How about "The JSON model is not valid for Minecraft Java Edition"
Or something like "The JSON model does not contain any geometry elements" since it only return that because it doesn't have any geometry information.

StandingPadAnimations:
> I originally I would make fixed json version of those entity blocks but blend file would be better suitable for it, I'm expecting "why this shulker block, chest in block model can't do that open intersction and meshswap has another version of them that can open, interact" _accident press the close button_

If a user imports a JSON model and expects to have all entity interactions, that's an issue with the user not knowing how to use MCprep, not an issue where MCprep has to spoonfeed users the quote on quote "correct" model (Especially because we _**already have**_ a meshswap spawner). For Steve's sake, it's called, "Import JSON Model", not, "Import JSON Model But We Force Meshswap Models for Certain Blocks". The latter sounds ridiculous, I know, but that's what's being implemented when we start assuming what the user wants. 

What about the users that want blocks _**without**_ any interaction features? (i.e. a regular block). Do we start telling them, "Sorry, but we don't want to give users the inconvinience of troubleshooting because of their mistake, so you'll have edit the MCprep code to properly import JSON models"? Again, sounds ridiculous, but again, that's what happens when we start to assume. 

(I would love to quote one of my former architecture teachers regarding assuming, but it involves some language not really fit on the GitHub repo for a Minecraft animation workflow addon)

The whole idea behind importing JSON models in MCprep is to import a JSON model as is, without any additional features. Forcing meshswap models in JSON import defeats the purpose of JSON imports (again, we already have a meshswap spawner).

zNightlord:
Why would they expecting an interaction while those blocks only the entity tile blocks in the game. How would they import a modded entity block? Since they don't even have any geometry data inside it would be **empty** because entity block is through hardcoded model renderer (now there is Geckolib, we don't even support Bedrock format so Geckolib is not too)
For the interaction. They don't want it they could delete the rig armature so... 

I know this is the hard one because of the terminology we use for this. #148 
MCprep - Minecraft - Generic 3D terms
Meshswap - Block or Entity Tile block - Can be a model or a rigged model
JSON Block Model  - Block - Model (no rig)

StandingPadAnimations:
If no geometry is defined in JSON models for entities, then it's simply more transparent to not include entities in the JSON importer. It's both disingenuous and frustrating to import meshswap models in the JSON importer. At this point, it's a decision between transparency vs. obscure behavior, which raises a red flag. If that means having to put a notice in the UI saying “Entities are under the Meshswap importer” (though I also see an entity spawner, so it should be obvious anyway), so be it. I'd rather have the JSON importer be transparent in what it can import than have to debug for hours later wondering why it's importing undesired meshswap models, only to find that it's forced behavior.

At this route, I'd say:
A. Remove entities from JSON models 
B. Check JSON models to see if they define geometry. If they don't, throw an error

I get it, we want to include every JSON model, but transparent behavior is preferred.

TheDuckCow:
So first want to acknowledge that this is a tricky subject, and indeed there's no singular correct answer. I do appreciate though the focus on what will make sense for users, and not obscuring functionality or extending lower level functions in ways that are not expected.

Copying some of my comments from discord, I could see us going a route that is more like creating a meta or uber operator that clearly indicators it's "doing multiple things", with lower level operations that can also be performed individually.

- Higher level operation:
   - Meshswap world -> that's the name today, and true it _is_ just "meshswap".
   - We could extend that or create a new operator which is to 3d-ify the world, using a combination of meshswap, json mode importing, etc. It could / should also be a bit more interactive (think: open gl, selecting boxes/voxels to convert, versus whole mesh selection) since rarely would someone intentionally try to swap and entire world at once for individual 3D models.. probably some area or cube selection would be relevant as there will always be performance overhead.
- Lower level operations:
   - Meshswap spawner as-is
   - Importing a JSON model (ie if there's no data in the json file, there's nothing to import / it should be skipped)
   - Something that does what @zNightlord is describing that somehow brings more animation/content to these modded files (we'd need to chat more about what these explicit things are, and how we can represent this)


These are just some high level thoughts. In terms of the scope of this PR: we are talking abotu Model JSON tweaks, so let's try to keep the scope tight and just focus on the core changes. It becomes problematic if we start making each PR try to cover many different things, as any conversation like this derails reviewing and merging of earlier changes that would otherwise be good to go.

So in that sense, I'd like to ask - is there anything else in progress here? Or are we good to review with the intention that approval would mean as-is merging? Asking @zNightlord thanks!

zNightlord:
Yeah it should be ready for another review
I am just going along with removing the entity "fixed" blocks out of the ui list to prevent the issue for now. 

For the modded block entity case, that is just an example for "Import JSON model" there are other json model structure that user would think they can import and have modded block geometry and rigged bones, animation.

mcprep_data_base.json
```diff
@@ -103,6 +103,25 @@
 	"Strider",
 	"Warden"
 ],
+"fixed_blocks": [
```
StandingPadAnimations, line 106: These aren't really fixed, so the name should reflect that. We'll say unspawnable for now, while we look at supporting Bedrock's JSON format in the future  (Minecraft is Minecraft, Java and Bedrock are equal to us)
```suggestion
"unspawnable_for_now": [
```

MCprep_addon/spawner/mcmodel.py
```diff
@@ -403,6 +437,10 @@ def update_model_list(context: Context):
 		# #fire or the likes in the file.
 		if "template" in name:
 			continue
+		# Filter the fixed_blocks 
+		# Either entity block or block that doesn't good for json
+		if name in env.json_data.get("fixed_blocks", blocks):
```
StandingPadAnimations, line 442:
```suggestion
		if name in env.json_data.get("unspawnable_for_now", blocks):
```

StandingPadAnimations (approved):
Just some name changes that are needed

StandingPadAnimations:
Mergind into dev

Issue #430, opened 2023-06-09
fixed BoxScape Studios villager rig


Roni-Raihan:
### Name of asset submission

rig villager
[villager.zip](https://github.com/TheDuckCow/MCprep/files/11705003/villager.zip)

### Quality checks

- [X] I have used only vanilla textures (or created images that are derived from vanilla textures)
- [X] This is a vanilla minecraft asset/mob/effect
- [X] I have verified my asset has the correct scale (1 Blender unit = 1 meter), and the scale has been applied (control+a, scale)
- [X] If my asset includes a rig, I certified there is a "root" or "main" bone, and the rig object name ends with the name ".arma" or ".rig" to work well with MCprep
- [X] I am the sole creator of this rig, other than any textures by Mojang. Any custom textures I created myself.

### Describe any usage details

This rig is a modified villager rig in mcprep from BoxScape Studios. I only try fix the hand. so I guess credit still has to go to BoxScape Studios

### What name should we use to credit you in the README.md file?

This rig is a modified villager rig in mcprep from BoxScape Studios. I only try fix the hand. so I guess credit still has to go to BoxScape Studios

### Confirmation

- [X] By submitting this asset, I affirm the asset is of my own creation. I understand this asset will be included with MCprep with a CC-BY (attribution) license agreement, and that users of MCprep should refer to [MCprep's readme](https://github.com/TheDuckCow/MCprep#credit) for credit or directly name me for credit when this rig is used.

TheDuckCow:
FYI I've created this updated version, which does some additional fixes:

- last saved version is blender 2.80, this is important for backwards compatibility
- Made two primary materials: one that has SSS, one that doesn't
- Fixed eyes so that if someone does prep materials, they go to the eye texture color instead of white
- Reset so the collection center point is at the base bone's center point (was offset slightly before)

[villager - Boxscape.blend.zip](https://github.com/Moo-Ack-Productions/MCprep/files/12472188/villager.-.Boxscape.blend.zip)

Might be good to sanity check that I didn't break anything with this update! But if you find this file ok now, we're good to call this submission confirmed.

Roni-Raihan:
Tried in versions 2.8 and 3.0. both work fine. And tried installing it on the mob spawner, it also worked well
![image](https://github.com/Moo-Ack-Productions/MCprep/assets/82071318/93c32fd5-8adc-483c-91e9-792bf7993ff8)

TheDuckCow:
[...]

Issue #420, opened 2023-05-05
Warden Rig Submission


DS9448:
### Name of asset submission

Warden Rig
[Warden Rig V2.zip](https://github.com/TheDuckCow/MCprep/files/11410633/Warden.Rig.V2.zip)

### Quality checks

- [X] I have used only vanilla textures (or created images that are derived from vanilla textures)
- [X] This is a vanilla minecraft asset/mob/effect
- [X] I have verified my asset has the correct scale (1 Blender unit = 1 meter), and the scale has been applied (control+a, scale)
- [X] If my asset includes a rig, I certified there is a "root" or "main" bone, and the rig object name ends with the name ".arma" or ".rig" to work well with MCprep
- [X] I am the sole creator of this rig, other than any textures by Mojang. Any custom textures I created myself.

### Describe any usage details

The gear icon next to the head has some properties that can be accessed through the 'item' tab in the side menu.

### What name should we use to credit you in the README.md file?

DS_9448

### What credit url should we use to reference you in the README.md file?

https://www.youtube.com/channel/UCSh4JBglkwdiMW__MvF9lPw

### Confirmation

- [X] By submitting this asset, I affirm the asset is of my own creation. I understand this asset will be included with MCprep with a CC-BY (attribution) license agreement, and that users of MCprep should refer to [MCprep's readme](https://github.com/TheDuckCow/MCprep#credit) for credit or directly name me for credit when this rig is used.

StandingPadAnimations:
@DS9448 can we receive a follow-up on this?

StandingPadAnimations:
Since there hasn't been a follow-up on this, and has passed through the whole 3.5.X release cycle, I'll be closing this. @DS9448 if you still want this to be part of MCprep, please create a new submission request

Issue #409, opened 2023-03-26
Wolf.


MrTerroChain:
### Name of asset submission

Well, I just changed your wolf and now I'm ready to share. It's Wolf. https://cdn.discordapp.com/attachments/784399899797487626/1089522099669966848/wolf.blend

### Quality checks

- [X] I have used only vanilla textures (or created images that are derived from vanilla textures)
- [X] This is a vanilla minecraft asset/mob/effect
- [X] I have verified my asset has the correct scale (1 Blender unit = 1 meter), and the scale has been applied (control+a, scale)
- [X] If my asset includes a rig, I certified there is a "root" or "main" bone, and the rig object name ends with the name ".arma" or ".rig" to work well with MCprep
- [X] I am the sole creator of this rig, other than any textures by Mojang. Any custom textures I created myself.

### Describe any usage details

None.

### What name should we use to credit you in the README.md file?

MrTerroChain.

### What credit url should we use to reference you in the README.md file?

https://www.youtube.com/@MrTerroChain.

### Confirmation

- [X] By submitting this asset, I affirm the asset is of my own creation. I understand this asset will be included with MCprep with a CC-BY (attribution) license agreement, and that users of MCprep should refer to [MCprep's readme](https://github.com/TheDuckCow/MCprep#credit) for credit or directly name me for credit when this rig is used.

TheDuckCow:
Sorry for the long delay to actually review this rig, we've been stretched thin. @MrTerroChain could you clarify what your version intends to provide as additional value to the existing rig?

At the moment, opening the rig in Blender 3.6 or 2.80 both seem to have fewer bone controls and the head doesn't rotate when rotating the intuitive head bone. I'm inclined to go with the current wolf rig we have in MCprep, but if there's specific reasons why your submission is an improvement, please let us know and then we can give tactical feedback to get it up to standard for inclusion.

TheDuckCow:
Following up here @MrTerroChain to see if you can help respond to the question above - would hate to discard your work, but want to make sure the value added is clear! Thanks

StandingPadAnimations:
Following up again, we'd hate to discard the rig, but we'll have to if no follow up is given

TheDuckCow:
Closing this for now, will happily re-open later if available!

PR #471, opened 2023-08-30
Defining new asset standards

### LINKED ISSUE #455: Defining new asset standards
With this newfound push for progress, I'm excited that we'll be having individuals like @MajesticalDiscomfort helping out with both asset contribution QA and even direct rig contributions. 

But in order to do this, we need to establish clear standards to follow (so that QA can be done objectively enough). My initial proposals, I'll refine this as we discuss further - with the intent to eventually create it's own markdown doc for this.

## Mob + Entity standard proposal

- The most important QA test: use "install rig" in MCprep, and then test importing
   - There should be no errors, and materials should work, and the armature and everything should be there with no unexpected parented objects or "extras" coming in.
   - Do this test in BOTH blender 2.80 and the latest (e.g. 3.6). 
- The blend file must be last saved in blender 2.80 OR have version partitioning across all versions of blender 2.80+
  - This is important as we claim to support 2.80+, so all rigs need to adhere to that
  - You can check the last version saved by opening the given blend file, going to the Script tab, and then typing this command: `bpy.data.version`, it should print out something like the screenshot below
  - Not sure what rig partitioning means? Take this as an example, and hopefully it clarifies:
    - `Warden - DigDanAnimates pre2.80.0.blend` Will be used by anything _before_ blender 2.80 (in other words, we'll be deleting this file soon!)
    - `Warden - DigDanAnimates pre3.0.0.blend` Will be used by anything _before_ blender 3.0 (e.g. 2.80, 2.93)
    - `Warden - DigDanAnimates.blend` Will be used by blender 3.0 and higher (due to no other "pre#.#.# being higher than 3.0.0)
  - We should aim to NOT partition rigs by version as much as possible, as it bloats download size and becomes even more to maintain. But sometimes, we can't avoid it.
- The rig must be self contained. That means textures are packed
   - This is to negate the issue of relative vs absolute paths, which just become a pain
   - Best practice is still to run "relpaths" beforehand, to avoid having any user's paths hard coded into the file, which is more of a privacy thing than anything.
- Extra unused data should be purged, to minimize bloat
- All rigs should be defined in their own top level collection. All other collections in the file not related to the rig must be removed
  - If a rig itself has multiple nested collections, that's ok - the top level collection just needs to have "MCprep" in the name, e.g. "Warden MCprep" will be what gets listed in the UI list, and it won't list any sub collections or other collections in the file. The warden rig is again an example of this. But prefer to just have a single collection where possible.
- Custom bone shapes are ok and in fact encouraged, just don't put them in the rig scene. If you want to put them into their own collection, either give it the name MCskip OR make sure the primary collection (s) have the text MCprep in it, so that the 
- The root-most bone of the rig must:
  1. Exist. Ie one central bone that everything is parented to, so if you move this everything, including control-only panels, also move
  2. Be named one of: `["main", "root", "base", "master"]`; case does not matter, though convention is to primarily use ROOT in all uppercase.
- _All bones_ should be named, no name.001's or anything like that.
  - Furthermore, left/right bones should be named as per the blender guidelines. [See here](https://docs.blender.org/manual/en/latest/animation/armatures/bones/editing/naming.html), which helps ensure animation tools will work like flip pose left/right.
- _All meshes_ should be named as well. No cube.001's, give them representation names

- The rig must be a reasonably faithful representation of a Vanilla mob or entity
  - No custom mobs that aren't in game (yet)
  - No highly detailed or stylized looks that would differ from the rest of the library, e.g. generally there shouldn't be beveling
  - Some extra details that accentuate and make high fidelity animations possible are OK. Examples include: Knees and elbows that bend (but it should still look vanilla if they don't bend), extra face controls or facial details (in the case of player rigs)
- Drivers are ok, but shouldn't be excessive - and minimize the use of python expression drivers where possible
  - Drivers add a lot of overhead, especially if you have many of one mob type in a scene. 
  - Simple drivers, such as "SUM position" used to translate a control panel to bone transforms are fine - but they should NOT be python expressions, rather they should be SUM of [variable transform channels], as these run much quicker.
- A fun note: The default position of the rig should show off its flaire! Ie, the saved pose _should_ be posted.
  - MCprep has controls for clearing the post on import. Meanwhile, having a posed starting point helps show off what the rig can do.
- If relevant, the rig can have a single default animation. There should _not_ be more than one animation though, since it wouldn't be obvious and would clutter the file during subsequent imports
  - The bat is an example of this, with a simple flying pattern. The Enchanting table is also an example (although that is a "block" and not an entity or rig)
- Technically, rigs can have a single python script named after the same blend file that are used to draw custom interfaces
  - But this is strongly not encouraged, as we cannot guarantee maintaining these ad hoc scripts. Much preferred to have drivers in the system itself, so that we don't have to rely on the user finding a custom panel or anything like that
  - However, this feature of having python script loading will be kept, as it is very useful for users who have their own custom rigs they like to install and use.

## Other asset contributions

QA will be done on a case by case basis. This would apply to meshswap mesh contributions, geometry node weather effects, and anything else not mentioned.

TheDuckCow:
Defining a new readme for guidance on rig contributions, and a resource we can point too when getting new submissions. Also updates the 

Closes #455

### DIFF
--- a/.github/ISSUE_TEMPLATE/Asset-Submission.yaml
+++ b/.github/ISSUE_TEMPLATE/Asset-Submission.yaml
@@ -23,6 +23,8 @@ body:
       options:
         - label: I have read, understood, and applied all the conventions of [this page](https://github.com/Moo-Ack-Productions/MCprep/blob/master/docs/asset_standards.md)
           required: true
+        - label: This file was last saved in blender 2.80 (for backwards compatibility)
+          required: true
         - label: This is a vanilla minecraft asset/mob/effect, and uses only vanilla textures
           required: true
         - label: I am the sole creator of this rig, other than any textures by Mojang. Any custom textures I created myself.
--- a/.github/ISSUE_TEMPLATE/Asset-Submission.yaml
+++ b/.github/ISSUE_TEMPLATE/Asset-Submission.yaml
@@ -1,6 +1,6 @@
 name: Asset contribution
 description: Submit an asset to include with MCprep
-labels: ["enhancement"]
+labels: ["asset-submission"]
 body:
   - type: markdown
     attributes:
@@ -21,13 +21,9 @@ body:
       label: Quality checks
       description: Please ensure all of the following are true for your asset.
       options:
-        - label: I have used only vanilla textures (or created images that are derived from vanilla textures)
+        - label: I have read, understood, and applied all the conventions of [this page](https://github.com/Moo-Ack-Productions/MCprep/blob/master/docs/asset_standards.md)
           required: true
-        - label: This is a vanilla minecraft asset/mob/effect
-          required: true
-        - label: I have verified my asset has the correct scale (1 Blender unit = 1 meter), and the scale has been applied (control+a, scale)
-          required: true
-        - label: If my asset includes a rig, I certified there is a "root" or "main" bone, and the rig object name ends with the name ".arma" or ".rig" to work well with MCprep
+        - label: This is a vanilla minecraft asset/mob/effect, and uses only vanilla textures
           required: true
         - label: I am the sole creator of this rig, other than any textures by Mojang. Any custom textures I created myself.
           required: true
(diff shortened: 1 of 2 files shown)


TheDuckCow:
Will be using these submissions as a validation point for the instructions:

https://github.com/Moo-Ack-Productions/MCprep/issues/409
https://github.com/Moo-Ack-Productions/MCprep/issues/420
https://github.com/Moo-Ack-Productions/MCprep/issues/430

TheDuckCow:
I'm going to merge this in the meantime, @MajesticalDiscomfort you are still free to comment here if you have suggestions, we can make more changes in the future as we have suggestions!

PR #480, opened 2023-09-18
Fixed syntax error on line 253 in prep.py

StandingPadAnimations:
Exactly what the title says, I can't push to `dev` for some reason

### DIFF
--- a/MCprep_addon/materials/prep.py
+++ b/MCprep_addon/materials/prep.py
@@ -634,7 +634,7 @@ class MCPREP_OT_load_material(bpy.types.Operator, McprepMaterialProps):
 
 		if self.animateTextures:
 			sequences.animate_single_material(
-				mat, context.scene.render.engine, export_location="original")
+				mat, context.scene.render.engine, export_location=sequences.ExportLocation["original"])
 
 		return success, None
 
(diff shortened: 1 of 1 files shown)


zNightlord (approved):
You might need to add that to the load material one too since I missed it.
https://github.com/Moo-Ack-Productions/MCprep/blob/ef78447d9c8d432a593b0c8b5fd03bb420458717/MCprep_addon/materials/prep.py#L622

zNightlord (approved):

StandingPadAnimations:
Alright, I'll merge this, but I've also noticed some other issues with incorrect types, such as like 637 (where a literal is used instead of `MCprep_addon.sequences.ExportLocation`).

@zNightlord, just curious, does your Python setup catch avoidable type errors like this? If not, I would highly suggest looking into revamping it. While I can try my best to catch these errors during the review process, things will definitely pass through as seen here. I know I sound harsh, but it's easier to catch and fix errors during development rather then fixing them during the last stretch before release (especially now that we can't commit directly to `dev` anymore)

TheDuckCow (approved):
Approving. 

As per the other thread, I think it's important to not directly push to dev - slightly annoying to always have to use a PR, but mistakes easily happen with "quick fixes" right before releases, I've been there before. Thanks for making the change here, and we'll surely fix other things as we go along.

PR #490, opened 2023-10-02
Improve release automation with automated test running

TheDuckCow:
Useful to ensure there is consistency during releases. This still requires local, manual running in the meantime, but would be plausible to further automated with github actions once we are comfortable with it.

### DIFF
--- a/push_latest.sh
+++ b/push_latest.sh
@@ -1,0 +1,103 @@
+source venv/bin/activate
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m' # No Color
+
+# -----------------------------------------------------------------------------
+# Validate and build the release.
+# -----------------------------------------------------------------------------
+
+# Affirm on the master branch
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+
+if [ "$BRANCH" != "master" ]; then
+    echo -e "${RED}Not on master branch, run:${NC}"
+    echo "git checkout master"
+    exit
+fi
+
+# TODO(TheDuckCow): Perform check when we have an automated way to set prod env.
+# if [[ `git status --porcelain` ]]; then
+#   echo "There are uncommited changes, ending"
+#   exit
+# fi
+
+git pull --quiet
+echo ""
+echo "Current status (should be empty!)"
+git status
+
+echo "Running tests"
+python run_tests.py -a
+if [ $? -eq 0 ]; then
+    echo "Tests passed, moving on"
+else
+    echo "Tests failed, exiting"
+    exit
+fi
+
+# -----------------------------------------------------------------------------
+# Update the data mapping to the latest values
+# -----------------------------------------------------------------------------
+
+# Manual override: we must recheckout some files after test runs finished,
+# due to some side effects of testing with local code references.
+git checkout MCprep_addon/MCprep_resources/mcprep_data_update.json
+git checkout test_files/test_data/jmc2obj_test_1_15_2.mtl
+git checkout test_files/test_data/mineways_test_combined_1_15_2.mtl
+git checkout test_files/test_data/mineways_test_separated_1_15_2.mtl
+
+python mcprep_data_refresh.py -auto
+
+git status
+
+ANY_DIFF=$(git diff MCprep_addon/MCprep_resources/mcprep_data_update.json)
+if [ -z "$ANY_DIFF" ]
+then
+      echo ""
+else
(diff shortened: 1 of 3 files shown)


TheDuckCow:
For reference, this was the exact script used to generate the initial draft of the MCprep v3.5.0 release (including the local mods of the test runner, so that we could extract an exit code).

push_latest.sh
```diff
@@ -0,0 +1,103 @@
+source venv/bin/activate
```
StandingPadAnimations, line 1: Add a shebang so we know what shell this is compatible with

StandingPadAnimations, line 1: Add a shebang so we know what shell this is compatible with

push_latest.sh
```diff
@@ -0,0 +1,103 @@
...
+# fi
+
+git pull --quiet
+echo ""
+echo "Current status (should be empty!)"
+git status
+
+echo "Running tests"
+python run_tests.py -a
+if [ $? -eq 0 ]; then
```
StandingPadAnimations, line 33: Assuming this is Bash, you can check the command exit code directly 
```suggestion
if "python run_tests.py -a"; then
```

push_latest.sh
```diff
@@ -0,0 +1,103 @@
...
+
+git status
+
+ANY_DIFF=$(git diff MCprep_addon/MCprep_resources/mcprep_data_update.json)
+if [ -z "$ANY_DIFF" ]
+then
+      echo ""
+else
+      echo "Commit new updates to mapping file before release:"
+      echo $ANY_DIFF
```
StandingPadAnimations, line 61: So I recently learned that Zsh (what macOS uses by default) does allow variables to not have quotes, which is nice. However, if we're assuming this can be run in Bash, quotes are required around variables to prevent issues (cause Bash is weird)
```suggestion
      echo "$ANY_DIFF"
```

push_latest.sh
```diff
@@ -0,0 +1,103 @@
...
+# -----------------------------------------------------------------------------
+
+
+echo "Building prod addon..."
+bpy-addon-build # No --during-build dev to make it prod.
+ls build/MCprep_addon.zip
+
+echo ""
+echo "Current live tags online:"
+echo $(git tag -l)
```
StandingPadAnimations, line 77: `echo` isn't needed here, the output of `git tag -l` will appear without it

push_latest.sh
```diff
@@ -0,0 +1,103 @@
...
+
+echo ""
+# Extract the numbers between parentheses, replace comma and space with period
+BASE_VER=$(grep "\"version\":" MCprep_addon/__init__.py | awk -F"[()]" '{print $2}' | tr ',' '.' | tr -d ' ')
+NEW_TAG="${BASE_VER}"
+
+echo -e "Current __init__ version: ${GREEN}${NEW_TAG}${NC}"
+read -p "Continue (Y/N): " confirm && [[ $confirm == [yY] || $confirm == [yY][eE][sS] ]] || exit 1
```
StandingPadAnimations, line 85: `read` should be passed with `-r` to prevent backslashes from being mangled

push_latest.sh
```diff
@@ -0,0 +1,103 @@
...
+
+NEW_NAME="MCprep_addon_$NEW_TAG.zip"
+mv build/MCprep_addon.zip "build/$NEW_NAME"
+
+# Make the tags
+echo ""
+echo "Generating tags and DRAFT release on github"
+# use --generate-notes to auto generate release notes (edit for public changelog)
+gh release create $NEW_TAG \
```
StandingPadAnimations, line 94: Again `$NEW_TAG` should have quotes due to Bash being Bash

StandingPadAnimations (commented):
Again, since no one wants to take charge with reviewing this, I've reviewed this PR (at least the shell script). I'm basing a lot of my comments off of `shellcheck`'s output (assuming this is also Bash).

There are a couple of things in this script that would be issues in Bash, but not in Zsh (the shell macOS uses as its default). While I assume this is a script meant to be ran with Bash, it would be interesting to make this `POSIX` compatible (i.e., no Bash, no Zsh, just standard POSIX), but that's not a huge requirement.

Since I'm on break, I won't approve or reject this PR; that should come from someone else on the team :sweat_smile:

push_latest.sh
TheDuckCow: Unforrunately when I tried this on my machine, the script errors with:

`./push_latest.sh: line 32: python run_tests.py -a: command not found`

Issue #491, opened 2023-10-07
Fetch new install ID if bad id found


TheDuckCow:
We had an oopsie with the launch of MCprep v3.5.0 regarding analytics: we shipped the addon with a pre-existing json file containing an install id! This meant that anyone installing MCprep between the launch of MCprep v3.5.0 on Sept 30th 12:30pm and Oct 6 ~8:30pm was being treated as the same user and no installs were detected.

As a general reminder for anyone reading this: [See here](https://theduckcow.com/dev/blender/mcprep/mcprep-stats/) for details about the optout MCprep tracking, and callout you can optout in user preferences of MCprep.

The fix to make is twofold:
- Ensure that big warning bells go off if any tracker file is made with a build (likely means some change in bpy-addon-build)
- Patch the `tracking.py` file to detect this known bad ID, so that the next time any of these users install an update, they'll get a fresh generated ID. 

I am in progress working on these updates, which should go in the next release.

TheDuckCow:
Marking as done.

PR #492, opened 2023-10-07
Ensure recover from bad tracker IDs

TheDuckCow:
This PR implements a few things:

- Detect if a "bad install id" is found in the addon folder. If so, fetch a new ID or recover from the parent folder's ID (if not a bad id itself)
- General tests for the tracker.py file to ensure it's working as intended for the top-levle public methods
- Test to validate this new recovery mechanism works exactly as intended
- A critical test to avoid ever shipping a new build with a prepackaged json id, which was how this whole problem started.

Closes https://github.com/Moo-Ack-Productions/MCprep/issues/491

All tests pass:

```
(4.0.0)   	all_tests	64	2	0	No errors
tests took 219s to run
```

### DIFF
--- a/MCprep_addon/tracking.py
+++ b/MCprep_addon/tracking.py
@@ -16,6 +16,11 @@
 #
 # ##### END GPL LICENSE BLOCK #####
 
+# critical to at least load these
+import os
+import bpy
+
+
 # safe importing, due to rogue python libraries being missing or invalid
 # in bad installs cases (if the rest of the addon works, don't make it
 # fail to work because of this module)
@@ -24,13 +29,9 @@ VALID_IMPORT = True
 # Request timeout (total request time may still be longer)
 TIMEOUT = 60
 
-# critical to at least load these
-import os
-import bpy
 
 # remaining, wrap in safe-importing
 try:
-	from . import conf
 	from . import util
 	from .addon_updater import Updater as updater
 	import re
@@ -46,7 +47,7 @@ try:
 	from .conf import env
 except Exception as err:
 	print("[MCPREP Error] Failed tracker module load, invalid import module:")
-	print('\t'+str(err))
+	print(f"\t{err}")
 	VALID_IMPORT = False
 
 
@@ -58,9 +59,10 @@ IDNAME = "mcprep"
 
 # max data/string lengths to match server-side validation,
 # if exceeded, request will return denied (no data written)
-SHORT_FIELD_MAX = 64-1
-USER_COMMENT_LENGTH = 512-1
-ERROR_STRING_LENGTH = 1024-1
+SHORT_FIELD_MAX = 64 - 1
+USER_COMMENT_LENGTH = 512 - 1
+ERROR_STRING_LENGTH = 1024 - 1
+
 
 # -----------------------------------------------------------------------------
 # primary class implementation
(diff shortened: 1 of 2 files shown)


MCprep_addon/tracking.py
```diff
@@ -260,10 +279,13 @@ def initialize(self, appurl, version, language=None, blender_version=None):
...
-							os.pardir,self._addon+"_trackerid.json")
-		self._tracker_json = os.path.join(os.path.dirname(__file__),
-							self._addon+"_tracker.json")
+		self._tracker_idbackup = os.path.join(
+			os.path.dirname(__file__),
+			os.pardir,
+			self._addon + "_trackerid.json")
+		self._tracker_json = os.path.join(
+			os.path.dirname(__file__),
+			self._addon + "_tracker.json")
```
StandingPadAnimations, line 288: I think it'll be worthwhile to see how this would look with `pathlib`, since `pathlib` tends to be clearer

MCprep_addon/tracking.py
```diff
@@ -386,35 +407,68 @@ def set_tracker_json(self):
...
+		valid_tracker = False
 		if os.path.isfile(self._tracker_json) is True:
 			with open(self._tracker_json) as data_file:
-				self.json = json.load(data_file)
+				jdata = json.load(data_file)
+
+			_json = jdata
+			if self._verbose:
+				print(f"{self._addon}: Read in json settings from tracker file")
+			if jdata.get("install_id") in SKIP_IDS:
```
StandingPadAnimations, line 427: I would add a comment explaining what happened (with a link to #491) that lead to this check. I'd also say that `SKIP_IDS` should be renamed to `BAD_IDS` or something along those lines.

MCprep_addon/tracking.py
```diff
@@ -386,35 +407,68 @@ def set_tracker_json(self):
...
-		else:
-			# set data structure
-			self.json = {
-				"install_date":None,
-				"install_id":None,
-				"enable_tracking":False,
-				"status":None,
-				"metadata":{}
-			}
+					print(f"{self._addon}: Skip ID detected, treat as new")
```
StandingPadAnimations, line 432:
```suggestion
					print(f"{self._addon}: ID in skip list detected, treating as new")
```

StandingPadAnimations (commented):
Since no one else seems to want to take charge of reviewing this, I'll just add my comments.

Seems good to me, albeit with some minor nitpicks I have in terms of style, but nothing seems off in the actual logic itself.

I won't approve or reject this PR though. I'll leave that to someone else (looking at y'all @zNightlord and @MajesticalDiscomfort)

MCprep_addon/tracking.py
TheDuckCow: Good call to add the git issue link. I think doing so at the top makes more sense. On the rename, I decided to go with INVALID_IDS since it's a bit more telling how we treat them, and something could be invalid for any number of reasons (whereas 'bad' in my mind is somehow more negative implication, albeit the most likely case)

MCprep_addon/tracking.py
TheDuckCow: With the new "Invalid ID" naming, I think this reads better now: `Invalid ID detected, treat as new`

MCprep_addon/tracking.py
TheDuckCow: idk I'm still a bit torn on pathlib always being better. It means we're more likely to stick with long lines, whereas right now I sort of like the multi line input arguments as clearly each segment of the path.

is this how you'd write that first line? (noting this goes over 80 charcters)

```
		self._tracker_idbackup = Path(
			os.path.dirname(__file__) / os.pardir / self._addon + "_trackerid.json")
```

In retrospect, the lines where we are using Pathlib are the ones most exceeding line lengths now in files like conf.py (even if we push to a newline after the `(` in `Path(`, still exceeds line length for the definition of `self.json_path_update`)

I'll probably stick with this and merge this in the meantime, but feel free to comment after merging if I'm overlooking something.

TheDuckCow:
Changes pushed, going to merge shortly.

```
tests took 17s to run
```

Issue #493, opened 2023-10-11
Use emissions tick box does not show up in the settings of swap texture pack


Jaitie018:
### Did the issue happen after restarting Blender?

- [X] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [X] I have checked known issues, and my problem is different

### What happened?

No enable or disable emissions tick box
![image](https://github.com/Moo-Ack-Productions/MCprep/assets/73315227/2aabe5c0-1b49-414c-8265-b9fda6d3b7d2)

### What steps are needed to produce the bug?

Click on mcprep's Swap texture pack

### What version of MCprep are you using?

3.5.0

### What version of Blender are you using?

3.6.4

zNightlord:
I can confirm the the UI inconsistency issue. 
The code for drawing prep material UI and the one in swap texture aren't shared at the moment.

Jaitie018:
omg hi znight

StandingPadAnimations:
@zNightlord do you think we can get this done soon? I've put it in the 3.6.1 milestone, but I'd rather for this to be fixed in time for 3.6 so we don't forget. If needed, I can create a PR for this as soon as possible

TheDuckCow:
Wow, the LOC to days of ticket open is quite a number to behold, thanks for getting to this one. Merged into dev now, will be in the next release.

StandingPadAnimations:
> Wow, the LOC to days of ticket open is quite a number to behold, thanks for getting to this one. Merged into dev now, will be in the next release.

Yeah, I sense we got a good handful of bug reports like that. 3.6.1 would be a good opportunity to go through some old bug reports and work on fixing those issues (assuming they're not already)

PR #495, opened 2023-10-15
Base class for Prep Material Operators

zNightlord:
Use base class for prep material operators for sharing the properties, drawing and prep material function
Use the same UI Layout draw as the main prep material #493 
Additionally, change some of the properties, added DocEnum for PackFormat

### DIFF
--- a/MCprep_addon/materials/generate.py
+++ b/MCprep_addon/materials/generate.py
@@ -30,10 +30,18 @@ from ..conf import env, Form
 
 AnimatedTex = Dict[str, int]
 
-class PackFormat(Enum):
-	SIMPLE = 0
-	SEUS = 1
-	SPECULAR = 2
+class DocEnum(Enum):
+	def __new__(cls, value, doc=""):
+		self = object.__new__(cls)  # calling super().__new__(value) here would fail
+		self._value_ = value
+		if doc != "":
+			self.__doc__ = doc
+		return self
+    
+class PackFormat(DocEnum):
+	SIMPLE = 0, "Use a simple shader setup with no PBR or emission falloff."
+	SEUS = 1, "Sets the pack format to Specular."
+	SPECULAR = 2, "Sets the pack format to SEUS."
 
 # -----------------------------------------------------------------------------
 # Material prep and generation functions (no registration)
--- a/MCprep_addon/materials/prep.py
+++ b/MCprep_addon/materials/prep.py
@@ -494,7 +494,7 @@ class MCPREP_OT_swap_texture_pack(
 			generate.set_saturation_material(mat)
 
 		if self.prepMaterials:
-			self.prep_mat()
+			self.prep_mat(context)
 
 		if invalid_uv:
 			self.report({'ERROR'}, (
--- a/MCprep_addon/materials/prep.py
+++ b/MCprep_addon/materials/prep.py
@@ -21,7 +21,8 @@ import os
 
 import bpy
 from bpy_extras.io_utils import ImportHelper
-from bpy.types import Context
+from bpy.types import Operator, Context
+from bpy.props import BoolProperty, FloatProperty, StringProperty, EnumProperty, IntProperty
 
 from . import generate
 from . import sequences
@@ -35,8 +36,8 @@ from ..conf import env
 # -----------------------------------------------------------------------------
 
 
-class McprepMaterialProps():
-	"""Class to inheret reused MCprep settings.
+class McprepPrepMaterial():
+	"""Base class for Prep Material
 
 	Benefit is to also enforce the same options, required where exec funcitons
 	pass through from one operator to the other (e.g. prep mats on swap packs)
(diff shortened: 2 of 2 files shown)


MCprep_addon/materials/prep.py
```diff
@@ -35,171 +36,158 @@
...
-		itms = []
-		itms.append((
-			"simple", "Simple (no PBR)",
-			"Use a simple shader setup with no PBR or emission falloff."))
-		itms.append(("specular", "Specular", "Sets the pack format to Specular."))
-		itms.append(("seus", "SEUS", "Sets the pack format to SEUS."))
-		return itms
-
-	animateTextures: bpy.props.BoolProperty(
+		return [(i.name.lower(), i.name.capitalize(), i.__doc__) for i in generate.PackFormat]
```
StandingPadAnimations, line 48: This is one of the worst things I've ever seen in Python, and I've seen a lot.

You've taken a perfectly easy to understand function, and compressed it down to this godforsaken mess for what, a really long single line return statement? This barely fits GitHub's file diff, let alone people's editors, and I had to reread this line 10 times to understand what it was doing.

Just no, I'm not allowing this change

MCprep_addon/materials/generate.py
```diff
@@ -30,10 +30,18 @@
...
+		self = object.__new__(cls)  # calling super().__new__(value) here would fail
+		self._value_ = value
+		if doc != "":
+			self.__doc__ = doc
+		return self
+    
+class PackFormat(DocEnum):
+	SIMPLE = 0, "Use a simple shader setup with no PBR or emission falloff."
+	SEUS = 1, "Sets the pack format to Specular."
+	SPECULAR = 2, "Sets the pack format to SEUS."
```
StandingPadAnimations, line 44: Why? 

You're taking a simple-to-understand concept (enums) and overcomplicating them with something related to docstrings. I can't even understand what's going on here, how would a new developer figure out what on God's earth this means?

MCprep_addon/materials/prep.py
```diff
@@ -35,171 +36,158 @@
...
-	track_param = None
-	track_exporter = None
-	@tracking.report_error
-	def execute(self, context):
-
+	def draw_before(self, context: Context) -> None:
+		pass
+	
+	def draw_after(self, context: Context) -> None:
+		pass
```
StandingPadAnimations, line 127: Don't leave functions that serve no purpose.

MCprep_addon/materials/prep.py
```diff
@@ -308,10 +296,35 @@ def execute(self, context):
...
+			self, width=300 * util.ui_scale())
+
+	track_function = "materials"
+	track_param = None
+	track_exporter = None
+	@tracking.report_error
+	def execute(self, context):
+		result = self.prep_mat(context)
+		if result:
+			return {'CANCELLED'}
```
StandingPadAnimations, line 323: I don't see the point in taking a perfectly good operator and move the body of `execute` to a separate method only to call the method in `execute`.

MCprep_addon/materials/prep.py
```diff
@@ -35,171 +36,158 @@
...
+		row.prop(self, "optimizeScene")
+		row.prop(self, "useEmission")
+		
+		self.draw_after(context)
+		
+	def prep_mat(self, context: Context):
+		"""A method for prepping material
+		
+		Returns: 0 Success, 1 Error
+		"""
```
StandingPadAnimations, line 177: Don't use integers for success or failure, use booleans. This is Python, not C.

StandingPadAnimations (changes requested):
(I'm not going to sugarcoat my comments here, this is truly how I feel about this PR)

This PR has so many issues with the changes made that I'm just going to close it (even though I wouldn't normally do an approval or denial on my break). Many of the changes here don't do any improvement in the code, and only serve to overcomplicate things. No good reason has, all that was stated is:

> Use base class for prep material operators for sharing the properties, drawing and prep material function
Use the same UI Layout draw as the main prep material #493 enum for PackFormat

That doesn't give any reason for me to even consider merging this, that just states the changes. You don't need to restate the changes you make because the commits already tell me that. Since I only have commits to work off of, and the changes made in the commits are terrible, I'm just going to reject this PR completely.

TheDuckCow:
Hey @StandingPadAnimations I think there's still merit in the problems being solved at least, this is after all in repsonse to a legitamite issue and I don't want to steer away others from contributing. I'm happy to take over reviewing with a coaching mentality on these things if you need to more formally step away - it might take me a minute to get back to PRs by myself or others (as you've seen), but I will get to it. 

@zNightlord - thanks for taking the time to put some changes together, let me get back to you soon on a recommendation!

StandingPadAnimations:
After reflecting for a bit the past day or so, I really have to apologise for my comments, I went too harsh on this.

With that, I'll reconsider the basic changes, but you'll need to open a new PR with the following changes:
- Make a new branch and redo the commits. Keep in mind we have a [preferred conventioned](https://github.com/Moo-Ack-Productions/MCprep/blob/master/CONTRIBUTING.md#commit-messages) defined in `CONTRIBUTING.md`, since commit messages like "initial" aren't good explainations of what the changes are (and more importantly the why behind the changes)
- Write a proper PR explaining why the changes are necesary, don't just restate the changes (that's what commits are for)

MCprep_addon/materials/generate.py
TheDuckCow, line 44: I do agree that this obfuscates the code a bit too much - fun to know this is possible, but I think we can get far enough by just defining the shared settings class.

MCprep_addon/materials/prep.py
TheDuckCow: I do agree with this point to make sure we don't leave leftover code, presuming this was WIP or something.

MCprep_addon/materials/prep.py
TheDuckCow: I'm less concerned about integers vs otherwise, but if we do want to understand the result of the operation, I think an annotated function return of `) -> bool:` would do nicely.

MCprep_addon/materials/prep.py
TheDuckCow: I actually think the intent here is good: we don't want to be calling duplicate code from different places, and if we can reuse the self.prep_mat function as-is in 3 places vs rewriting the same boilerplate, then it's probably a good thing. I'm not convinced as of yet it's a great idea to force the same popup UI for each place we are including prep materials options. I'd have to try it live first to see how it's looking and feeling. There's some pros and cons here since we're also clobbering together any warning messages or popups that happen in the normal course of prep materials, but then again maybe those would be still relevant anyways.

So, for the re-opened PR, feel free to keep this same original structure as a starting point, and we can go from there.

MCprep_addon/materials/prep.py
```diff
@@ -489,19 +494,7 @@ def execute(self, context):
...
-				improveUiSettings=self.improveUiSettings,
-				usePrincipledShader=self.usePrincipledShader,
-				useReflections=self.useReflections,
-				useExtraMaps=self.useExtraMaps,
-				normalIntensity=self.normalIntensity,
-				makeSolid=self.makeSolid,
-				syncMaterials=self.syncMaterials,
-				packFormat=self.packFormat,
-				skipUsage=True)
+			self.prep_mat(context)
```
TheDuckCow, line 497: This line is definitely a big pro of this change. With the old way, it is tech debt of constantly having to keep the same args in sync in different places.

TheDuckCow (commented):
I'm supportive of re-opening this PR, with my own commentary below for reference.

MCprep_addon/materials/generate.py
StandingPadAnimations: I don't really think it even makes sense to abstract it in the first place. `bpy.props.EnumProperty` is tuple based, so we'll have to do conversions anyway. I think the current way it's done on `master` is already a good compromise between readability and verbosity; any other route is more significantly more verbose for little (if any) readability gains.

The only possible route I can think of that somewhat makes sense is setting each enum in `PackFormat` to be equal to their `bpy.props.EnumProperty` counterpart, but that's not a good idea (creation of unnecessary objects in if statements would be an issue)

PR #498, opened 2023-11-01
Change Skin Swap's behavior for more nuanced swapping

StandingPadAnimations:
Skin swap originally would change all image textures on all materials of the selected object, regardless of whether the user wanted to or not. This is not ideal for rigs that need multiple separate textures. For example, the ardoni rigs in Black Plasma Studio's (now Squared Media) Songs of War series used multiple textures to handle stuff like emission. In addition, speaking from personal experience, I have several textures I use on my rigs these days, and skin swap's current behavior makes it unusable in my case. 

To combat this, skin swap will now use a more nuanced method of swapping skins based on node name, which is an already established precedent from the MCprep Cycles Optimizer ([README](<https://github.com/Moo-Ack-Productions/MCprep?tab=readme-ov-file#cycles-optimizer-node-settings>)). MCprep will now search for a node called `MCPREP_SKIN_SWAP`, and treat it as a diffuse node. The old behavior will still remain as a legacy option, but will be deprecated in 3.5.2/3.6 (depending on the circumstances) and will be removed completely in 3.6/3.7 (again, depending on circumstances)


StandingPadAnimations:
This does quality as a breaking change, hence why we have the option for legacy behavior

zNightlord (commented):
Looks good now. 

This part doesn't impact much but i think it should be dicussed.

Using `label` might be better than `name`. Name has that need to be unique 
For example: having 2 nodes for swap would be `MCPREP_SKIN_SWAP_A` `MCPREP_SKIN_SWAP_B` while label would be `MCPREP_SKIN_SWAP` for both

Optionally having an extra `MCPREP_SKIN_SWAP' property in the material properties and used that string for checking is also good for someone have their own naming convention, don't have to follow a hardcode one if it exists. Without it it is fine.

zNightlord (approved):

TheDuckCow (changes requested):
I think this is an ok change, but some requests from me:

- I think we should rename the "legacy" option to "Swap all textures".
- Make this "Swap all textures" be default-on, and we should not intend to remove it in the future
- And finally, add a warning message if "Swap all textures" (ie this new mode you've created) is False _AND_ no texture was modified. This message should say: `No image node "MCPREP_SKIN_SWAP" found, nothing swapped`.

**Rationale:**

For any users who have _not_ prepped materials on their rig (though this is the default in MCprep on spawn), skin swap will stop working. Favoring the advanced use case will confuse new users. If the skin swap operation requires you to go into a material to edit a node to make it work after bringing in a rig you downloaded somewhere (and likely doesn't have this label), it's defeating much of the purpose of the operator in the first place since it's for people who don't want to bother (or know how) to modify shader nodes.

So, I'd rather this be an advanced feature that can be toggled _on_. You'd only want this behavior if you already knew you had previously set up the rig material with this label, and therefore you would know to expect un-tick the box in the redo last menu. For everyone else, they'd be unaware it was an option but they'd still get skin swapping as expected.

TheDuckCow:
I'm suggesting we move this to v3.5.2, I'd be happy to make the changes I recommended above to ensure that it _is_ backwards compatible. The more nuanced swapping should be a user preference anyways to opt in, already better to have it rolled out as an option.

We have loose plans for a wider UI overhaul which would include a more overall scene settings configuration, this could eventually live there in addition to user default preferences.

TheDuckCow:
FYI I'm powering through the changes here, to wrap up the release.

TheDuckCow:
Alright, I'm pretty happy with the current state of this PR - @StandingPadAnimations not sure if you want to do a quick sanity check, ironically reviewing your own PR with my coauthored commit, but I realized there were some mistakes with some use of the positional arguments (and a weird indentation with one function).

All tests pass locally for me:

```
(4.0.2)   	all_tests	64	2	0	No errors
tests took 196s to run, ending with code 0
```

TheDuckCow (approved):
Meant to press approve for the record yesterday

TheDuckCow:
Presuming this is working fine-enough, I'm going to go ahead and merge so we can cut a release soon.

PR #501, opened 2023-11-07
Fix block model overlay geometry

zNightlord:
Small fix of this issue: [Image](https://media.discordapp.net/attachments/737872746700079235/1163524282618490900/image.png?ex=655b92ee&is=65491dee&hm=a97dea5690e99be313a6388ad21807b47484e1ca6c1629f0f8583bc4334e8d24&)

[Grass block json](https://github.com/InventivetalentDev/minecraft-assets/blob/1.20.1/assets/minecraft/models/block/grass_block.json)

Block like grass block has a small overlay on them ~~should be treated as a threat causing z fighting and clean it up~~ but let this one pass.

### DIFF
--- a/MCprep_addon/spawner/mcmodel.py
+++ b/MCprep_addon/spawner/mcmodel.py
@@ -371,7 +371,7 @@ def add_model(
 			# Give slight offset by normal for overlay geometry
 			if face_mat == "#overlay":
 				bmesh.ops.translate(bm, verts=face.verts,
-						    vec=0.05 * face.normal)
+						    vec=0.02 * face.normal)
 
 			for j in range(len(face.loops)):
 				# uv coords order is determened by the rotation of the uv,
@@ -386,7 +386,7 @@ def add_model(
 	# Quick way to clean the model, hopefully it doesn't cause any UV issues
 	# Ignore model has overlay geometry, causing issue
 	if not textures.get("overlay"):
-		bmesh.ops.remove_doubles(bm, verts=bm.verts, dist=0.01)
+		bmesh.ops.remove_doubles(bm, verts=bm.verts, dist=0.005)
 
 	# make the bmesh the object's mesh
 	bm.to_mesh(mesh)
--- a/MCprep_addon/spawner/mcmodel.py
+++ b/MCprep_addon/spawner/mcmodel.py
@@ -378,7 +378,8 @@ def add_model(
 				face.material_index = materials.index(face_mat)
 
 	# Quick way to clean the model, hopefully it doesn't cause any UV issues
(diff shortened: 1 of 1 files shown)


TheDuckCow (approved):
Thanks for the fix!

TheDuckCow:
At face value, seems this could still cause some z-fighting issues. Seems to render fine in Eevee, but there's still a problem in cycles on render (in blender 3.6):

Screen Shot 2023-11-10 at 9 13 38 AM

My approval of the PR still stands because this is technically still better than before (we can just remove the overlay geo), but what about just the faces of "overlay" geo, or expanding it slightly to have a slight offset @zNightlord?

MCprep_addon/spawner/mcmodel.py
```diff
@@ -367,6 +367,12 @@ def add_model(
 			)
 
 			face.normal_update()
+			
+			# Give slight offset by normal for overlay geometry
+			if face_mat == "#overlay":
+				bmesh.ops.translate(bm, verts=face.verts,
+						    vec=0.05 * face.normal)
```
TheDuckCow, line 374: Seems a little strong from my standpoint at 0.05, 

```suggestion
						    vec=0.02 * face.normal)
```

Not sure if that gets in the way of hte remove doubles distance, maybe drop that as well to half?

MCprep_addon/spawner/mcmodel.py
```diff
@@ -378,7 +384,9 @@ def add_model(
 				face.material_index = materials.index(face_mat)
 
 	# Quick way to clean the model, hopefully it doesn't cause any UV issues
-	bmesh.ops.remove_doubles(bm, verts=bm.verts, dist=0.01)
+	# Ignore model has overlay geometry, causing issue
+	if not textures.get("overlay"):
+		bmesh.ops.remove_doubles(bm, verts=bm.verts, dist=0.01)
```
TheDuckCow, line 389: Per above suggestion, to ensure enough margin

```suggestion
		bmesh.ops.remove_doubles(bm, verts=bm.verts, dist=0.005)
```

TheDuckCow (approved):
I think this is good now, I'm going to add these two suggestions and merge.

FireDasher:
this adds a depth effect that looks kinda cool but it makes holes in the mesh

StandingPadAnimations:
@FireDasher please make a [new bug report](https://github.com/Moo-Ack-Productions/MCprep/issues/new?template=Bug-Report.yml) as opposed to commenting on a pull request merged 2 years ago (otherwise assistance becomes harder)

FireDasher:
ok

Issue #508, opened 2023-11-24
Capping upper supported Blender version for MCprep 


StandingPadAnimations:
# Rationale
With MCprep 3.5, we tested with a version of the Blender 4.0 alpha. Finding that MCprep 3.5 at the time worked with 4.0, we released 3.5 thinking it had 4.0 support. Fast-forward after release, and Principled V2 was merged into Blender 4.0 (at the time, still in alpha). Suddenly, MCprep no longer had support for MCprep 4.0, due to breakage in prep materials. While this is really more of an unfortunate coincidence, Blender development has recently started to up the pace of changes made. Blender 4.1 for instance no longer has the Musgrave texture node, and has merged it with the Noise texture node.

I think now it's necessary that we add a hard cap to MCprep's upper supported Blender version, and change how we determine support during development

# Determining Support
I propose that for testing and developing MCprep, we simply ignore the alpha release, and focus on the latest stable or beta (depending on if there is a beta or not). Alphas are volatile, and bound to break MCprep with a single PR, whereas betas pretty much don't break unless absolutely necessary.

For example, with the next MCprep release, for now we should only focus on Blender 4.0, not 4.1. Once Blender 4.1 enters beta, however, then we can start developing around it. This should allow for a smoother transition for MCprep with regard to new versions.

Alternatively, we could also only do the latest LTS release as the upper limit (starting with 4.3), and jump to the next LTS release when it enters beta. This may sound like a weird suggestion, but LTS releases are much more stable in terms of API changes, so we can spend more time focusing on the actual addon itself, rather than dealing with compatibility issues.

# Implementing a Hard Cap
As it stands, it's easy for users to expect MCprep to be compatible with the latest release of Blender if it pops up in the UI. MCprep currently has no indicator with regard to lack of support for a version beyond what it has been developed around. Thus, I propose we add a hard cap (again, based on the latest stable or beta release) that is checked by the addon, and alert the user when they're using a version too new for MCprep.

Will this be annoying? Yes.
Is it necessary with modern Blender development? Also Yes.

While it's annoying that we have to consider a hard cap now, hopefully this should improve user experience in the end.

StandingPadAnimations:
This felt quite dystopian to write, idk why

TheDuckCow:
[...]

Issue #268, opened 2022-01-03
Helpful Horse - suggestions panel


TheDuckCow:
Create a "helpful horse" panel. It would be the topmost panel in MCprep, and would provide meta help to MCprep users. Not to be thought of as  *too* similar to Microsoft Clippy, this would be a passive feature that early only would be hidden behind an experimental mode setting, and later potentially minimized by default. The goal is to be strictly helpful, out of the way, but is there if you need it. In that sense the goal is to "be effective, be succinct, be gone". If it never gets used, then it means things are going well.

As of the current moment, the idea is the panel we have two macro capabilities:
1) Proactive suggestions: Based on the state of the scene, suggest specific things to do. E.g. if it detects there is a tiled image loaded in, but it hasn't been animated yet (and thus looks squashed), or if there are missing textures, suggest using the `replace missing` operator. It could also be used to showcase new features, and provide quick access to "take feedback survey" in blender itself.
2) Generic question box: Essentially a search bar within MCprep. User queries are parsed, and results are listed out in little sub panels, providing quick actions. Behind the scenes, there could be some minimal set of hard coded replies for inputs (such as model lists or operator names). But later, this could be connected to a cloud database doing best guess responses for a given input query. This cloud feature would not be part of the MVP.

Below are some examples of both of these modes, but would expand over time.

- Proactive suggestions:
  - Update to new version of MCprep (right now, this dynamically shows in other panels, which feels random; here is a better home)
  - Show Animate textures operator if non animated, tiled images loaded
  - Suggest "Replace missing textures" operator if missing textures found
  - Feature checklist: an optional element the user can look at to see if they've used all features of MCprep (can be used as a feature discovery aide, as well as some form of a built-in tutorial mechanism).
  - Calls for Community challenges, potentially with built in blender submissions (stretch goal to say the least).
  - In-blender way to submit renders to the MCprep community, this could be paired with a discord bot (though this would need a lot of thought on its own, to ensure accountability, safety, moderation etc. Early on could be a review queue if submissions are few).
- Generic question box result types:
  - Quick spawn. If your text matches any block, mob, item, show that spawner button directly. This is probably the # 1 reason to implement this panel, since it can be unclear to the user where to find a specific spawnable "thing" (is it an item? Entity?)
  - Quick skin change (for those already local, or from user names).
  - Tutorials, link out to youtube videos or help pages for specific operators if matching the input name.
[...]

PR #509, opened 2023-11-26
Added warning for using MCprep in a version beyond what was tested

StandingPadAnimations:
This will now warn the user when using MCprep in a version of Blender above the upper cap. This is intended to prevent users from thinking that MCprep won't have issues in new Blender versions.

Currently we warn that we won't provide support for MCprep users using a version above the current upper limit, set at 4.0 for now. 

How it looks:
image

This is to be set with the convention laid out in #508 (subject to change), which defines how we should choose the upper version cap for MCprep

### DIFF
--- a/MCprep_addon/conf.py
+++ b/MCprep_addon/conf.py
@@ -30,6 +30,8 @@ from bpy.utils.previews import ImagePreviewCollection
 # TYPING UTILITIES
 # -----------------------------------------------------------------------------
 
+# Upper supported Blender version
+UPPER_BV_CAP = (4, 0, 0)
 
 class Form(enum.Enum):
 	"""Texture or world import interpretation, for mapping or other needs."""
@@ -75,6 +77,7 @@ class MCprepEnv:
 		self.dev_file: Path = Path(os.path.dirname(__file__), "mcprep_dev.txt")
 
 		self.last_check_for_updated = 0
+		self.valid_environment: bool = True
 
 		# Check to see if there's a text file for a dev build. If so,
 		if self.dev_file.exists():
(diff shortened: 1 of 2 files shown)


TheDuckCow:
Commented [over here](https://github.com/Moo-Ack-Productions/MCprep/issues/508#issuecomment-1826496719), let's discuss on the issue thread and decide what to do here after. Though I expect to at least suggest we keep the code for the notification / alert, even if it's off by default.

StandingPadAnimations:
With some of the concerned raised in #508, I've now modified the error message to be more encouraging of testing and bug reports
image

TheDuckCow:
Ok, ha, interesting it's now aggressive on the side of pushing for feedback. It feels a bit much that it appears on every panel and takes up so much space, and is not dismissible (unlike update notifications). 

What about this proposal: What if we make a head start to implementing the ["Helpful Horse" panel](https://github.com/Moo-Ack-Productions/MCprep/issues/268), which could be the only place both this notification as well as the update notice panel appear?

And, if any notice is present (either this new one OR the update ready notice), then a warning icon appears in the panel header. To be clear, this would thus be a panel separate to the others, but appearing first in order ideally so it's at the top.

Already a plan for v3.6.0, we eventually want things like a "cross-spawner search" and other tips and things to show up there. So, this is perhaps a nice stepping stone.

Finally, on the version, I still don't think we should be that aggressively pushing each time blender updates. It just means that most of the time this notification will always be there (which also means over time, people will start to just ignore it entirely, thus diluting if we ever have _other_ notifications). How about we set it to show _two_ minor versions of blender in the future from some reference point? Let me know your thoughts @StandingPadAnimations

StandingPadAnimations:
I think we could use LTS releases as reference then, since the second number (as of LTS 2.93) will always be divisible by 3 (and thus easy to detect), as well as any changes to the first number (while Blender doesn't officially follow semantic versioning, the first number does often represent a major breaking release) 

Helpful Horse integration would also probably be a better idea, especially as we could also merge the in-Blender report pop-ups as well (which should make it easier for users to use). Bonus if we could have a button to autofill the GitHub bug report form too (like Blender does)

StandingPadAnimations:
Closing this to allow us to start Helpful Horse on a clean slate

Issue #505, opened 2023-11-16
bpy-addon-build keeps including tracker json


TheDuckCow:
For some reason, the addon builder keeps including this `mcprep_addon_tracker.json`. Builds should _**never**_ have this file included ever ever, it's already messed out our analytics a number of times, and is currently in prod for the second time. At this moment, I can't see where this file is getting picked up from in my local git repo. Need to address asap and update prod version of 3.5.1.

Technically no user-facing impact, but we'll lose an understanding of how many people are having success using blender 4.0 and so forth until this is fixed.

TheDuckCow:
Relates to https://github.com/Moo-Ack-Productions/MCprep/issues/491, where I thought I had put in sufficient checks and balances to prevent this from being a problem. Not the case... :(

TheDuckCow:
It seems a tracker json did end up in my local git folder. The issue is that the test runner code appears to generate the file and backup file during testing. It's not even due to just the tracker tests, other tests seem to somehow generate this. 

I have no idea how `test_no_local_idfile` [wasn't failing before](https://github.com/Moo-Ack-Productions/MCprep/pull/504#issuecomment-1811815223), even though it _is_ correctly failing now.

I'm going to create create a branch to merge into master which does a force removal of this file after tests run.

PR #513, opened 2023-12-19
Upgrade to BpyBuild 0.3.0

StandingPadAnimations:
This PR upgrades the MCprep build system (BpyBuild) to 0.3.0. BpyBuild 0.3.0 itself is almost done, just currently in the final stages of testing (https://github.com/Moo-Ack-Productions/bpy-build/pull/6, separate PR due to some current issues with repository settings).

To install BpyBuild 0.3.0:
```sh
pip install git+https://github.com/Moo-Ack-Productions/bpy-build@v0.3.0
```

# Changes for Developers
*sales pitch time*

All BpyBuild commands have been shortend, including the BpyBuild command itself. Here are some examples of what's changed:
- `bpy-addon-build --during-build dev` -> `bab -b dev`
- `bpy-addon-build -v 3.0` -> `bab -v 3.0`

BpyBuild also now uses argparse for the CLI, which makes the arguments far more reliable in design

The config has also changed. Here's an example of what changes
## pre-0.3.0 syntax
```yaml
# bpy-build.yaml
during_build:
  dev:
    - copy_file("some_file -> destination in the addon folder")
    - [some_command with-args | and-pipes, another_command]
```

## 0.3.0 syntax
```yaml
# bpy-build.yaml
build_actions:
  dev: "dev.py"
```
```py
# dev.py
def main() -> None:
    with open("mcprep_dev.txt", "w") as f:
        f.write("hi guys c:")

if __name__ == "__main__":
    main()
```

All build actions are now Python based, massively expanding what's possible. This means BpyBuild can also be expanded in its feature set in almost any way you want. Some examples of what's possible now with BpyBuild's new build actions:
- Ignore filters by deleting files with an action
- Downloading and inserting external assets
- A "production" build option for including large files
- Using custom analysis scripts 
- etc

If it can be done in Python, it can now be done in BpyBuild, a massive improvement from BpyBuild's old clunky system.

Versions are no longer strings, which means the following syntax is now unsupported:
```yaml
install_versions:
  - '~/some/path/to/blender/3.5`
```

In the future, a new system will be introduced to handle custom config paths


StandingPadAnimations:
In time for Christmas too!

bpy-build.yaml
```diff
@@ -2,22 +2,18 @@ addon_folder: MCprep_addon
...
+  - 3.5
+  - 3.4
+  - 3.3
+  - 3.2
+  - 3.1
+  - 3.0
+  - 2.93
+  - 2.9
+  - 2.83
+  - 2.8
```
TheDuckCow, line 16: Does this still work for 2.80 with this naming convention? If so, no complaints

TheDuckCow (approved):
One comment but you're approved so you're free to merge when you see fit

StandingPadAnimations:
@TheDuckCow when you get the chance, can you test this on your machine? Especially with more complex arguments like `-v` and whatnot. Want to make sure BpyBuild'a bugs get ironed out first before merging

StandingPadAnimations:
@TheDuckCow remember the annoying issue in #505? That should be resolved now with the new ignore filters action :D

I've also made a separate folder to store all the action scripts so that everything's organized (default scripts have their own subfolder too). Let me know if any changes need to be made.

StandingPadAnimations:
Updated `run_tests.py` to use the new build command, should we pass the `-dbg` flag as well (for BpyBuild) during testing to get debug information, or should we just keep it as it?

TheDuckCow:
Marking myself to re-requesting review so I can try it with the latest changes of the package itself

StandingPadAnimations:
Doesn't work atm with the changes I made earlier today (but that's a BpyBuild side issue)

StandingPadAnimations:
Alright, issues from yesterday seem to have resolved themselves somehow (cosmic bit flip perhaps?), so it should be ready to try

TheDuckCow:
I'll take a look in the next couple days. Slightly wary of rolling out this for a minor release, but I can see how confident I am with testing in the end. I at least really do know what to look for now.

StandingPadAnimations:
FYI I pushed a change to BpyBuild today that changes `filters` to `ignore_filters` (to make it more explicit what it means). In addition, I initially moved the Python based ignore filter (for `mcprep_addon_tracker.json`) to BpyBuild's ignore filters, but I decided to keep the Python based one since we can give a warning if we find a file we don't want to end up in a final build:
image

Perhaps we should rename this to "Blacklisted Files" and maybe even stop the build if we find a blacklisted file, though I think this is sufficient for now.

StandingPadAnimations:
I think we're good with the migration, just need to get BpyBuild's PR merged before we can merge this

bpy-build.yaml
TheDuckCow: Per [this comment](https://github.com/Moo-Ack-Productions/bpy-build/pull/6#pullrequestreview-1859500659), seems to not work for 2.8 (or 2.80) with the current branch of v0.3.0 of bpy-build

bpy-build.yaml
```diff
@@ -2,22 +2,23 @@ addon_folder: MCprep_addon
...
-  dev:
-    - create_file("mcprep_dev.txt")
-  
+build_actions:
+  default: 
+    script: "action-scripts/default.py"
+  dev: 
+    script: "action-scripts/dev.py"
+    ignore_filters: 
+      - "*.blend"
```
TheDuckCow, line 24: On further review, I don't think this filter works very well because it just results in the blend files being missing in the final archive right?

I was most of the way through implementing a "fast" build option which set the ignore filter to be the mcprep_resources folder, and that seemed to be working fine until I realize we fully delete the existing install always via the line:

```
if addon_path.exists():
    shutil.rmtree(addon_path)
```

Not having any of the source files means test fail, so not very useful for dev building. I'd remove this until we have a proper fast flag that can leave some files in place.

TheDuckCow (approved):
Approving, but let's remove the ignore filter usage from the dev mode (it's not even that much faster unless you skipp the entire resource folder, because much of the time is coming from zipping and unzipping all of the texture files).

StandingPadAnimations:
Assuming there's no more issues, I'll merge this. BpyBuild v0.3.0 is also now availible on PyPI

Issue #514, opened 2023-12-24
Optimize scene "Transmission" not found in blender 4.0


TheDuckCow:
The most common error report we have in the current version of MCprep is below, only affected blender 4.0+ users (90 different users have submitted reports). This appears to be _only_ present in v3.5.1 of the addon

Raw report:

```
File "<addon_path>/prep.py", line 292, in execute\n bpy.ops.mcprep.optimize_scene()\n File "<addon_path>/ops.py", line 109, in __call__\n ret = _op_call(self.idname_py(), kw)\nRuntimeError: Error: Python: Traceback (most recent call last):\n File "<addon_path>/optimize_scene.py", line 325, in execute\n self.is_pricipled(context, mat_type, node)\n File "<addon_path>/optimize_scene.py", line 183, in is_pricipled\n transmission_socket = node.inputs["Transmission"]\nKeyError: 'bpy_prop_collection[key]: key "Transmission" not found'\nLocation: C:\Program Files\Blender Foundation\Blender 4.0\4.0\scripts\modules\bpy\ops.py:109\n\n
```

Different associated user messages with this report (quotes):

- Prepping JMC materials
- Hola quiero comunicar este error. Este me paso por que intente hacer una animacion, le di a la opcion el renderizado de siclos de ecenas. y me salto este error
  - My Spanish is not that great, and google translate gets hung up on 'siclos', but I gather it's the same root as above with an issue on optimizing the scene. 
- prepping materials
- i was trying to optimize my scene in cycles for faster renderng

StandingPadAnimations:
> Hola quiero comunicar este error. Este me paso por que intente hacer una animacion, le di a la opcion el renderizado de siclos de ecenas. y me salto este error 

Using my best Spanish translating ability (Texas forces us to take 2 years of learning a language other then English for graduation), though this will be more of a general translation, and will be far from perfect:
> Hello, I want to communicate this error. I came across this when I try to make an animation, with the render option "Optimize Cycles Scene", and I get this error

("Siclos" I assume means "Cycles" in this context, from the phrase "siclos de ecenas", which I assume means "Cycles scene") 

That out the way, I'm not exactly sure what's going on, but I have been wanting to remove the Cycles optimizer for a while now, for several reasons:
- Hasn't caught up with latest Cycles developments
[...]

PR #517, opened 2023-12-24
Mark optimizer as deprecated

StandingPadAnimations:
This marks the Cycles optimizer as deprecated in the UI, and alerts the user about the change
image

In addition, `optimize_scene.py` has a note on the deprecation, with the reasons why (mostly a copy+paste from #514)


TheDuckCow (approved):
Looks fair to me, thanks for proactively working on this

PR #518, opened 2023-12-26
Prevent collection-effect obj particle src from being cleaned up.

### LINKED ISSUE #515: Blender cleans up object particles sources after collection spawning
A problem I ran into while working on this tutorial, if you save a collection effect whereby the source of a particle system used is an object (as opposed to a collection), it will work right away as an instance in a scene, but after saving and reloading this scene, the source particle gets cleaned up and thrown away.

Using manual inspection, I found we can prevent this by marking the objects used as particle sources as fake users at the object level in the outliner. We just need to do this in code. Honestly, seems like a blender bug and I've run into it before, but I think we only addressed the collection use case before.

The workaround in the meantime is to only author collection effects where any included particle systems only reference collections as the source. The consequence is potentially lost work after coming back to this file - I myself even ran into this after rendering an entire scene

TheDuckCow:
Before this change, if a collection effect included a particle system where the source of the particles was an object (as opposed to an entire collection), then on save and reload of that file, the particle object would have been cleaned up, as it was not linked to the scene and for some reason, since forever, blender does not count usage of an object as a particle emitter as an object use reference when it comes to its own internal cleanup.

Works now, tested by saving and reloading a file after spawning in a version of the firework tutorial where I didn't use a collection as the firework source.

All tests passing; edit: manually tested in blender 4.0 and 2.80

```
(3.6.2)   	all_tests	63	2	0	No errors
(4.0.2)   	all_tests	63	2	0	No errors
(3.5.1)   	all_tests	63	2	0	No errors
(3.4.0)   	all_tests	63	2	0	No errors
(3.3.1)   	all_tests	63	2	0	No errors
(3.2.1)   	all_tests	63	2	0	No errors
(3.1.0)   	all_tests	63	2	0	No errors
(3.0.0)   	all_tests	63	2	0	No errors
(2.93.0)   	all_tests	62	3	0	No errors
(2.90.1)   	all_tests	62	3	0	No errors
(2.80.75)   	all_tests	61	4	0	No errors
tests took 218s to run, ending with code 0
```
Closes #515

### DIFF
--- a/MCprep_addon/spawner/effects.py
+++ b/MCprep_addon/spawner/effects.py
@@ -151,14 +151,7 @@ def add_area_particle_effect(context: Context, effect: ListEffectsAssets, locati
 	post_systems = list(bpy.data.particles)
 	imported_particles = list(set(post_systems) - set(pre_systems))[0]
 
-	# Assign particles as fake, to avoid being purged after file reload.
-	if hasattr(imported_particles, "instance_object"):
-		# 2.8+
-		if imported_particles.instance_object:
-			imported_particles.instance_object.use_fake_user = True
-	elif imported_particles.dupli_object:
-		# the 2.7x way.
-		imported_particles.dupli_object.use_fake_user = True
+	mark_particles_fake_user(imported_particles)
 
 	# Assign the active object and selection state.
 	for sel_obj in bpy.context.selected_objects:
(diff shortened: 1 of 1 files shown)


StandingPadAnimations (approved):
Looks good to me

PR #522, opened 2023-12-31
Made notice regarding owning a legal copy of Minecraft more visible

StandingPadAnimations:
This PR makes the notice regarding us assuming the user has a valid copy of Minecraft more visible

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ MCprep is an addon dedicated to speeding up the workflow of Minecraft animators 
 ## Installing MCprep
 ### Click the link below and download the .zip file (re-zip if auto-unzipped into a folder necessary), install into blender (2.80 through 3.5 supported)
 
-**Note: It is assumed you have a valid, legal copy of Minecraft. Moo-Ack! Productions reserves the right to deny support if one is found to use a pirated copy of Minecraft**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,9 +11,11 @@ MCprep is an addon dedicated to speeding up the workflow of Minecraft animators
 ## Installing MCprep
 ### Click the link below and download the .zip file (re-zip if auto-unzipped into a folder necessary), install into blender (2.80 through 3.5 supported)
 
+**Note: MCprep is developed with the assumption that our users have a valid, legal copy of Minecraft. As such, Moo-Ack! Productions reserves the right to deny support if one is found to use a pirated copy of Minecraft**
```
TheDuckCow, line 14: If the purpose is to be more legally airtight, then we should rephrase "we assume" to "you must" as I do below. Otherwise, it almost further emphasizes the license as optional by saying the downside is just no user-support.

```suggestion
**Note: To use MCprep, you must have a valid, legal copy of Minecraft to comply with Mojang's terms of service. No support is provided for anyone using MCprep without a valid Minecraft license**
```

TheDuckCow (approved):
Approving this, thanks for making the change! I'll direct-commit the suggestion below

PR #523, opened 2024-01-01
i18n Support: PR 1 - Second Attempt

StandingPadAnimations:
This is a second attempt at the first PR for i18n support, using Python's [`gettext` module](https://docs.python.org/3/library/gettext.html) instead of a custom JSON protocol. This ended up being easier since the GNU developers have tons of utilities for automatic generation (such as `xgettext`)

As of making PR 1, a good chunk of the MCprep N panel is translatable, taking advantage of GNU Gettext (which is the standard for translation). Translators may use utilities like [Poedit](https://poedit.net) to aid in translation. In the future, we may also look towards [Weblate](https://weblate.org/), though that's not a priority right now.

For developers, what changes is that strings should now use `text=env._(str)` instead of `text=str`, in order to make it translatable. In addition, for every release, we need to refresh the base POT file. This is simple though:
```sh
xgettext -d base -o MCprep_addon/MCprep_resources/Languages/mcprep.pot "MCPREP_FILE" --keyword="env._
```

(Now here's the copy paste from #458)
> \- Goal defined in #439 

PR 1 will be merged for MCprep 3.6. While not explicitly defined as a blocker for the 3.6 release, it will be treated as such unless decided otherwise.


StandingPadAnimations:
So far, with a basic test (using English), everything seems fine. We'll have to port the Chinese translation we've received in #458 and see how well that works though. No errors though (thank god)

StandingPadAnimations:
As I try to salvage what we had on the old branch, I think I need to do some of the previous ID stuff again, cause this is getting out of hand
image

StandingPadAnimations:
The only changes that should be required though is swapping the text for the IDs in all the `env._` calls, then using PO files to map all of that to actual text. The mapping could be automated, but setting the IDs will require some work.

.github/ISSUE_TEMPLATE/Bug-Report.yml
```diff
@@ -5,7 +5,7 @@ body:
   - type: markdown
     attributes:
       value: |
-        Thanks for taking the time to fill out this bug report for MCprep! We just need some info. Check here for more info: https://github.com/TheDuckCow/MCprep/issues/322. Please note that if you are looking for general help with Blender itself or a related problem like Mineways, you are better off joining the official [MCprep discord](https://discord.gg/mb8hBUC). **Note that support will not be provided if you're found to have an illegal copy of Minecraft!**
+        Thanks for taking the time to fill out this bug report for MCprep! We just need some info. Check here for more info: https://github.com/TheDuckCow/MCprep/issues/322. Please note that if you are looking for general help with Blender itself or a related problem like Mineways, you are better off joining the official [MCprep discord](https://discord.gg/mb8hBUC).
```
TheDuckCow, line 8: Is there a reason for removing this recent addition here? (or just a mixup on syncing with dev?)

MCprep_addon/MCprep_resources/Languages/en/LC_MESSAGES/mcprep.po
```diff
@@ -0,0 +1,523 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
```
TheDuckCow, line 4: We should probably fill these out/replace with the actual GPL header, or remove and have it just use the package's parent license inherently.

MCprep_addon/MCprep_resources/mcprep_data_update.json
TheDuckCow, line 1: I tend to prefer to not edit this file during unrelated code changes, and rather make it its own change closer to the time of a release (even if it shows up as being modified locally). Unless is this just showing as a diff due to merging in dev? We could try changing the base branch and then changing it back to reset, weird trick that fixes the diffs

MCprep_addon/conf.py
```diff
@@ -123,6 +125,13 @@ def __init__(self):
 		# that no reading has occurred. If lib not found, will update to [].
 		# If ever changing the resource pack, should also reset to None.
 		self.material_sync_cache = []
+		
+		# i18n using Python's gettext module
+		self.en_i18 = gettext.translation("mcprep", self.languages_folder, fallback=True, languages=['en'])
+
+		# We've made this a part of the env class 
+		# since it's probably better in the long run
+		self._ = self.en_i18.gettext
```
TheDuckCow, line 134: This feels very odd to me, what is this line accomplishing?

MCprep_addon/conf.py
```diff
@@ -210,6 +219,183 @@ def deprecation_warning(self):
 
 env = MCprepEnv()
 
+
+# ! Deprecated as of MCprep 3.5
+def init():
```
TheDuckCow, line 224: Let's check diffs/make sure dev is merged in - I'll maybe hold off on further review/comments below until that's done

TheDuckCow (commented):
Leaving just an initial comment status, as I think we need to refresh (either by flicking the base branch) or otherwise merging in the latest changes from dev. But still adding a few comments in the meantime. I like the idea of using the standardize python library, good to see that forming

.github/ISSUE_TEMPLATE/Bug-Report.yml
StandingPadAnimations: It's a mix up with dev syncing, I'm not sure how that happened

MCprep_addon/conf.py
StandingPadAnimations: It creates a functions called `_` in the env class. Generally `_` is used for localizing strings (in Python as well as C++) as it's pretty easy to type and recognizable by localization tools. However, I've put it in env since `_` is also syntax for a variable that's unused (like in a loop), and I'm not sure if that would be affected by making `_` global

.github/ISSUE_TEMPLATE/Bug-Report.yml
TheDuckCow: Cool - once resolve, just re-request review for me. Sorry it took ages to get to this in the first place

MCprep_addon/conf.py
TheDuckCow: Do you have some reference or examples in third party packages using _ for localization? More for my own learning. Definitely a fan of following convention, though want to make sure it remains clear as well. But it does make sense to use such short functions for things like that, since it'll be all over.

StandingPadAnimations: The Python docs for [`gettext`](https://docs.python.org/3/library/gettext.html) use `_`:
```py
import gettext
gettext.bindtextdomain('myapplication', '/path/to/my/language/directory')
gettext.textdomain('myapplication')
_ = gettext.gettext
# ...
print(_('This is a translatable string.'))
```
This is also standard in the GNU gextext library for C (which is what the Python gettext is based on), and is the default keyword `xgettext` looks for:
https://en.m.wikipedia.org/wiki/Gettext#Operation

StandingPadAnimations:
We're not taking translations at the moment, but submissions will be open when this is merged (hopefully by MCprep 3.6)

StandingPadAnimations:
Ok I'm trying to refresh this branch, but GitHub is being a bit weird about it. Should we create a new branch and cherry-pick the changes on top?

TheDuckCow:
If you need to force push, it's fine with me. Or a fresh branch, whatever you prefer to avoid less headache. As long as the changes are carried through that's the main thing (just be aware using a new branch won't correlate back to this PR)

StandingPadAnimations:
Moved to #538

PR #526, opened 2024-01-02
Added clarification to README

StandingPadAnimations:
Added clarification that the new rule only applies to assets (applying to all of MCprep is a violation of GPL as I later realized)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ MCprep is an addon dedicated to speeding up the workflow of Minecraft animators 
 ## Installing MCprep
 ### Click the link below and download the .zip file (re-zip if auto-unzipped into a folder necessary), install into blender (2.80 through 3.5 supported)
 
-**Note: To use MCprep, you must have a valid, legal copy of Minecraft to comply with Mojang's terms of service. No support is provided for anyone using MCprep without a valid Minecraft license**
+**Note: To use MCprep with assets, you must have a valid, legal copy of Minecraft to comply with Mojang's terms of service. No support is provided for anyone using MCprep without a valid Minecraft license** Versions of MCprep without the assets are not bound to this requirement
 
 [![Install MCprep](/visuals/mcprep_download.png)](https://theduckcow.com/dev/blender/mcprep-download/)
 


StandingPadAnimations:
Before we merge this, I'll email the FSF regarding the topic of JSON models, since (unless I'm mistaken and the JSON models aren't from the vanilla resource pack) JSON models could actually count as code under GPL

StandingPadAnimations:
Yeah, looks like whatever we do, it's going to be a grey area:
```
Thank you for contacting us and for your willingness to abide by the GPL. Let me start by explaining that the Free Software Foundation (FSF) is working to secure freedom for computer users by promoting the development and use of free software. Please note that we at the FSF do not and cannot help developing proprietary
software, even indirectly. Not only we find it morally wrong, we are also a small organization with limited resources that cannot afford to lose sight of our goal, which is free software. Also, we do not provide legal advice.

As far as I understand your project is an effort to improve rendering animations from a nonfree game. The game is released under a nonfree EULA and it's copyright holder attempts to control modding efforts and reuse of the game's data files by a policy, which can be changed at any time. Your project encourages users to use the nonfree game and results in them being unable to exercise the freedom.

Projects based on proprietary software, such as yours, knowingly live in a gray area. These projects are often one update of a proprietary EULA, or one cease-and-desist letter away from oblivion. So even your best efforts, which are absolutely commendable, to do the right thing in regards with the GPL are overshadowed by this unfortunate state of affairs.

Even though your project is released under the GPL and even though it is an improvement to the free software program Blender, the freedom that the GPL grants cannot be reconciled with the restrictions on the game files imposed by the copyright holder to the game.
```

The key part is this:
> Even though your project is released under the GPL and even though it is an improvement to the free software program Blender, the **freedom that the GPL grants cannot be reconciled with the restrictions on the game files imposed by the copyright holder to the game.**

I think for now, we'll just add the clarification, and then work towards not distributing Minecraft assets (instead asking users to provide a copy of the Minecraft JAR file for assets). I have sent a follow-up regarding requiring users to supply the Minecraft assets themselves, but I think it's fine if we merge this for now.

TheDuckCow:
Let me know if we're still waiting for some kind of response here

TheDuckCow (approved):

PR #528, opened 2024-01-04
Readme and resources update

TheDuckCow:
No functional changes, routine auto updates and text improvements.


TheDuckCow:
Welp, so much for being a safe update. Apparently minecraft renamed `grass` ie upright grass texture) to `colormap/grass` and on top of that, for some reason the texture isn't getting saturated/color tinted like it should. Being such a central texture, used for meshswapping.

It means we need to decide whether to rename the actual texture / object name within meshswap for grass, or whether to create an inserted fake mapping for that object in our mapping file. I need to contemplate this.

PR #530, opened 2024-01-07
Added error object for better user experience

StandingPadAnimations:
In the past, MCprep used many methods to determine if an error occurred somewhere outside an operator, such as checking for `None`, using `1` to signify an exception, etc. These methods have the following issues though:
- Not descriptive for the caller; the caller knows an error occurred somewhere, but doesn't know where. This makes it difficult to create a well crafted error message, which makes errors harder to troubleshoot
- Inconsistent, which makes MCprep's code harder to understand.

To mitigate these issues in new code, a new error class is needed. This PR introduces the `MCprepError` class, which aims to give error messages enough information one needs to debug. With `MCprepError`, the following can be used as general rules:
- The function that returns an object is responsible for providing the following:
	- Error type 
	- File and line number (this is where the `inspect` module comes in handy)
	- Error message (optional, as some functions might be general purpose)
- The function that receives the error is responsible for the following:
	- Graceful handling of the error
	- Alerting the user of the error, *with all the information provided by the error object*

To bring this out to devs, this PR doesn't perform any actual migration of old code, hence why it's pretty lackluster. The goal is to migrate as we go.

### DIFF
--- a/MCprep_addon/conf.py
+++ b/MCprep_addon/conf.py
@@ -236,7 +236,6 @@ class MCprepError(object):
 		was created in. The preferred way 
 		to get this is __file__
 	"""
-	msg: Optional[str]
 	err_type: BaseException
 	line: int 
 	file: str
--- a/MCprep_addon/conf.py
+++ b/MCprep_addon/conf.py
@@ -18,7 +18,8 @@
 
 from mathutils import Vector
 from pathlib import Path
-from typing import Union, Tuple, List, Dict
+from typing import Optional, Union, Tuple, List, Dict
+from dataclasses import dataclass
 import enum
 import os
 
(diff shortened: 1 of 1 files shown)


MCprep_addon/conf.py
```diff
@@ -207,6 +208,54 @@ def deprecation_warning(self):
...
+class ErrorType(enum.Enum):
+	"""
+	All error types
+
+	Attributes
+	------------
+	FILE_NOT_FOUND:
+		File was not found
+	"""
+	FILE_NOT_FOUND = enum.auto()
```
TheDuckCow, line 220: I feel like error type is unnecessary, we already have a whole slew of exception classes that python provides. Why not make that the datatype field, a _type_ assignment, e.g. something like this in the dataclass: `error_type = Exception` (not sure how typing would be for that actually, since it shouldn't be an _instance_ of an exception)

TheDuckCow (approved):
At face value I'm fine with this, curious to see an example of where you would like to use it.

I'll also say, some of this feels it could later become duplicative to what the `traceback` provides, which we utilize in the tracking.py file. Just something to consider

MCprep_addon/conf.py
StandingPadAnimations: I think it should be fine to use exception classes directly, since they're just objects. As long as they're not being raised, it shouldn't cause any problems

TheDuckCow: Yup exactly, let's do that - will avoid a lot of re-definitions of things, I think

StandingPadAnimations:
Merging this as done

PR #536, opened 2024-01-20
Refactored colorspace setup to be more flexible

StandingPadAnimations:
In the past, we'd set this to a hard coded value. However, that proved to be annoying to users using non-standard OCIO configs like ACES or early versions of AgX. MCprep already fixes MTL files for ACES compatibility, so we're expanding this to prep materials.

In `mcprep_data.json`, there will now be a section called "non_color_options", which is a list of different options for Non-Color Data/Generic Data. If a user is using a non-standard setup, they can simply add the correct option in the JSON file and prep materials will function properly.

The matching goes in order from first to last, and MCprep will use the first value matched at runtime.

### DIFF
--- a/MCprep_addon/conf.py
+++ b/MCprep_addon/conf.py
@@ -69,7 +69,7 @@ except:
 class MCprepEnv:
 	def __init__(self):
 		self.data = None
-		self.json_data = None
+		self.json_data: Optional[Dict] = None
 		self.json_path: Path = Path(os.path.dirname(__file__), "MCprep_resources", "mcprep_data.json")
 		self.json_path_update: Path = Path(os.path.dirname(__file__), "MCprep_resources", "mcprep_data_update.json")
 
@@ -207,6 +207,38 @@ class MCprepEnv:
 			import traceback
 			self.log("Deprecation Warning: This will be removed in MCprep 3.5.1!")
 			traceback.print_stack()
+
+	def current_line_and_file(self) -> Tuple[int, str]:
+		"""
+		Wrapper to get the current line number and file path for
+		MCprepError.
+
+		This function can not return an MCprepError value as doing
+		so would be more complicated for the caller. As such, if 
+		this fails, we return values -1 and "UNKNOWN LOCATION" to 
+		indicate that we do not know the line number or file path 
+		the error occured on.
+
+		Returns:
+			- If success: Tuple[int, str] representing the current 
+			line and file path 
+
+			- If fail: (-1, "UNKNOWN LOCATION")
+		"""
+		from inspect import currentframe, getframeinfo
+		cur_frame = currentframe()
+		# currentframe can return a None value
+		# in certain cases
+		if cur_frame:
+			# Get the previous frame since the
+			# current frame is made for this function,
+			# not the function/code that called 
+			# this function
+			prev_frame = cur_frame.f_back
+			if prev_frame:
+				frame_info = getframeinfo(prev_frame)
+				return frame_info.lineno, frame_info.filename
+		return -1, "UNKNOWN LOCATION"
 
 @dataclass
 class MCprepError(object):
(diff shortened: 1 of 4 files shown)


MCprep_addon/conf.py
```diff
@@ -69,7 +69,7 @@ class Engine(enum.Enum):
 class MCprepEnv:
 	def __init__(self):
 		self.data = None
-		self.json_data = None
+		self.json_data: Optional[Dict] = None
```
TheDuckCow, line 72: My POV is that if we add typing, we should do so for the entire definition (making data optional = None above as well, and also hint what the non-optional form is). Good improvement either way 👍

MCprep_addon/conf.py
```diff
@@ -208,6 +208,38 @@ def deprecation_warning(self):
...
+		indicate that we do not know the line number or file path 
+		the error occured on.
+
+		Returns:
+			- If success: Tuple[int, str] representing the current 
+			line and file path 
+
+			- If fail: (-1, "UNKNOWN LOCATION")
+		"""
+		from inspect import currentframe, getframeinfo
```
TheDuckCow, line 228: Would prefer importing at the top of the file

MCprep_addon/conf.py
```diff
@@ -208,6 +208,38 @@ def deprecation_warning(self):
...
+		# in certain cases
+		if cur_frame:
+			# Get the previous frame since the
+			# current frame is made for this function,
+			# not the function/code that called 
+			# this function
+			prev_frame = cur_frame.f_back
+			if prev_frame:
+				frame_info = getframeinfo(prev_frame)
+				return frame_info.lineno, frame_info.filename
```
TheDuckCow, line 240: This gets to be pretty deep nesting, would prefer early exits. Just wrote this with out checking in the git PR comment so could be off, but something like this could be good:

```suggestion
		if not cur_frame:
		        continue
	        # Get the previous frame since the
	        # current frame is made for this function,
	        # not the function/code that called 
	        # this function
	        prev_frame = cur_frame.f_back
	        if not prev_frame:
	                continue
		frame_info = getframeinfo(prev_frame)
		return frame_info.lineno, frame_info.filename
```

MCprep_addon/materials/generate.py
```diff
@@ -436,7 +439,9 @@ def set_cycles_texture(
 				node.image = new_img
 				node.mute = False
 				node.hide = False
-				util.apply_colorspace(node, 'Non-Color')
+				res = util.apply_noncolor_data(node)
+				if res is not None:
+					env.log(f"TypeError on {res.line} in {res.file}: {res.err_type}")
```
TheDuckCow, line 444: Since we use this in several places, maybe conf.py could provide a "static" const template string? Maybe or even a utility function called `log_error_lineno()` which takes this object as an input.

MCprep_addon/util.py
```diff
@@ -48,31 +48,45 @@
...
-		node.color_space = 'NONE'  # for better interpretation of specmaps
-	elif hasattr(node.image, "colorspace_settings"):  # later 2.8 versions
-		# try default 'Non-color', fall back to best guess 'Non-Colour Data'
-		if color_enum == 'Non-color' and noncolor_override is not None:
-			node.image.colorspace_settings.name = noncolor_override
-		else:
+	
+	# While earlier versions of 2.8 used a different 
+	# attribute name for colorspace, MCprep officially 
+	# supports the full 2.8 release at a minimum (as of 3.5)
```
TheDuckCow, line 71: I think it's enough to just say this should work for 2.80+, don't need to keep comments around for what things "used" to do (that's what changelogs are for)

TheDuckCow (changes requested):
Minor change requests, but I think this largely looks good. I'll just want to live test before I fully approve (and once these little updates are made)

Running tests locally, everything passes which is good:

```
(4.0.2)   	all_tests	64	2	0	No errors
```

TheDuckCow (approved):
Looks good, thanks for making this improvement!

PR #538, opened 2024-01-28
i18n Support: PR 1 - Third and Final Attempt

StandingPadAnimations:
(This PR is to resolve some issues related to the second PR)

This is a third attempt at the first PR for i18n support, using Python's [`gettext` module](https://docs.python.org/3/library/gettext.html) instead of a custom JSON protocol. This ended up being easier since the GNU developers have tons of utilities for automatic generation (such as `xgettext`)

As of making PR 1, a good chunk of the MCprep N panel is translatable, taking advantage of GNU Gettext (which is the standard for translation). Translators may use utilities like [Poedit](https://poedit.net) to aid in translation. In the future, we may also look towards [Weblate](https://weblate.org/), though that's not a priority right now.

For developers, what changes is that strings should now use `text=env._(str)` instead of `text=str`, in order to make it translatable. In addition, for every release, we need to refresh the base POT file. This is simple though:
```sh
xgettext -d base -o MCprep_addon/MCprep_resources/Languages/mcprep.pot "MCPREP_FILE" --keyword="env._
```

(Now here's the copy paste from #458)
> \- Goal defined in #439 

PR 1 will be merged for MCprep 3.6. While not explicitly defined as a blocker for the 3.6 release, it will be treated as such unless decided otherwise.


TheDuckCow:
Just marking myself as a reviewer so I get back to it, unless you think there will be further structural changes. I'll review this and the bpy-build together (I've held off on that as I had some issue sin the past getting things working again, but will be good to get that merged in too ahead of our developer meeting)

.gitignore
```diff
@@ -12,7 +12,7 @@ MCprep_addon/MCprep_resources/resourcepacks/mcprep_default/materials.blend1
 mcprep_venv_*
 .cache
 .python-version
-venv/
+.venv/
```
TheDuckCow, line 15: Maybe we should just do both, since our docs show quick start where venv isn't hidden? (though I do agree, hidden is more standard)

```suggestion
venv/
.venv/
```

MCprep_addon/MCprep_resources/Languages/mcprep.pot
```diff
@@ -0,0 +1,523 @@
+# SOME DESCRIPTIVE TITLE.
```
TheDuckCow, line 1: Calling out this is not a descriptive title :)

MCprep_addon/MCprep_resources/Languages/mcprep.pot
```diff
@@ -0,0 +1,523 @@
...
+"POT-Creation-Date: 2024-01-01 10:18-0600\n"
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"Language: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=CHARSET\n"
+"Content-Transfer-Encoding: 8bit\n"
+
+#: MCprep_addon/mcprep_ui.py:98
```
TheDuckCow, line 20: I will admit, I'm both a bit concerned and confused by this file.

The concern: if we add one new line at the start of each file, does that already invalidate every single line in every single po file, since the lineno is no longer accurate?

The confusion: Are these line numbers somehow automatically generated, or did you have to hand author these line numbers? If hand authored, then I have no idea how we'll keep this up to date. Seems like we should be able to get away with the .po key translation on its own for any given string, shouldn't we? I recall the built in way of blender handling translations was by doing something like auto recognizing a given phrase is used for translation, and then automatically swapping it for us so long as that exact phrase was present in the translations dictionary.

I'm all for using more standardized cross-project ways of managing translations, this feels like it's going to be a huge amount of work and much more cumbersome than what we had in prior iterations with just key-value pair lookups.

MCprep_addon/conf.py
```diff
@@ -124,6 +126,13 @@ def __init__(self):
 		# that no reading has occurred. If lib not found, will update to [].
 		# If ever changing the resource pack, should also reset to None.
 		self.material_sync_cache = []
+		
+		# i18n using Python's gettext module
+		self.en_i18 = gettext.translation("mcprep", self.languages_folder, fallback=True, languages=['en'])
```
TheDuckCow, line 131: Is this going to feed in with Blender's native locale setting? Or do users have to somehow configure that separately. It should be a clear requirement that MCprep's translation follows suite of blender itself.

MCprep_addon/mcprep_ui.py
```diff
@@ -95,9 +95,9 @@ def restart_layout(layout):
 	alert_row.alert = True
 	alert_row.operator(
 		"wm.quit_blender",
-		text="Restart blender",
+		text=env._("Restart blender"),
```
TheDuckCow, line 98: Related to my point above, being confused about these.

Is this "Restart blender" used as a key lookup somewhere? (Are these env._() calls used to generate the po files... for then manual editing..?)

TheDuckCow (commented):
Left a few commnets, I think I need a bit of clarity on how translations are going to be authored in practice. 

Either way with the PR, it would be good to add a readme about how to make new translations. Could you start a simple markdown that has two sections?
- A simple 1,2,3 step guide of how to contribute a first translation. ie what files/commands must be run, whether they are starting with a template etc. What they need installed, and how to check its working (change blender settings etc)
- Another simple 1,2,3 guide for repo owners (us) on how to ensure translations are working during testing/releases (ie, do we need to re-generate some files on releases or check for update line numbers)

Perhaps once I am reading that, I'll better understand and be able to comment further. Definitely want to have this in for 3.6, but want to make sure we're doing it in the best way for both maintainability, ease of third party contribution, and (last after the first two) following pythonic conventions.

MCprep_addon/MCprep_resources/Languages/mcprep.pot
StandingPadAnimations: The line numbers are autogenerated by `xgettext`. They don't have actual meaning, beyond just giving an idea of where in the source code the text comes from. The actual import parts are lines like these:
```po
msgid "Restart blender"
msgstr ""
```

Which are identical to our old system of key-value pairs, just in a more standard format, with parsing handled by Python's `gettext` module. There's no really "built-in" way of doing translations in Blender, but Blender itself does use PO files as well, so I think we're on the right track.

MCprep_addon/conf.py
StandingPadAnimations: The goal is to, by default, use the user's set locale for Blender, but we might also have our own menu if we receive translations for languages not in Blender's locale settings.

MCprep_addon/mcprep_ui.py
StandingPadAnimations: `env._` is what does the actual mapping of a string to its corresponding translation, like how `translate_str` in the first attempt of i18n maps strings to translations. In addition, it's also used by `xgettext` to generate POT files (PO templates). 

Normally this would be `_` in the global scope, but to avoid unintentional breakage I've put it in `env`.

StandingPadAnimations:
@TheDuckCow finished the port from our old system to the new system: 
image

This changes on the fly and falls back to English if the user is using a language we don't support

MCprep_addon/MCprep_resources/Languages/mcprep.pot
StandingPadAnimations: (just realized this now) POT files are templates, so this is meant to be a placeholder for translators

TheDuckCow:
Ok thank you for the clarity and making the readme! I understand it quite a bit better, and alleviates my concerns for the most part. Give me a bit of time as I want to play around with it myself to see how a few steps would occur.

One thing I recommend is that we also put into our release preparation script or even the compile script, the command which re-compiles the po files to mo. That way there's no chance they are out of date, and while contributors *should* build and validate their translations live, in case they didn't check it in we'd know pretty quick.

StandingPadAnimations:
Would be interesting, but for now I think we can just manually compile it ourselves when merging a PR related to translation. 

As an aside, this PR is almost done as far as I'm aware, I just need to add some notes for maintainers and verify that we've gotten the entire N panel (obviously we haven't gotten everything, but the N panel is a good start for MCprep 3.6)

StandingPadAnimations:
I'd say this is complete now. @TheDuckCow, can you double-check and see if there's anything that's missing?

There's currently the following limitations I hope to address in future PRs:
- Operator names are not translated. This should be addressed in PRs 2 and 3 (roadmap for these is to be decided at a later date, but I think 3.7 for PR2 and 3.8 for PR 3 would make for some good pacing)
- Headers are not updated immediately and require a restart. From what I can tell, `bl_label` is set at registration, and isn't dynamically updated. From what I've heard, [`bpy.app.translations`](https://docs.blender.org/api/current/bpy.app.translations.html) could aid in this, but it's also extremely verbose for translators, and would result in a large Python file that would be hard to maintain in the end (hence why I didn't go this route). However, for just operators and panel headers, we could manually (or better yet, automatically it at build time) construct a small dictionary ourselves for just panel and operator labels.
- Dynamic menus are not translated. I did say we might do a 4th PR for these, but dynamic menus add enough complication that I think we should just skip these in the end.

StandingPadAnimations:
Added conversion script to go from PO files to a Python dictionary that is loaded into `bpy.app.translations` (with a fallback to directly sourcing the PO and MO files themselves). This does add one small dependency (`polib`, since `gettext` only uses MO files, not the source PO files), but overall it works surprisingly well. I'm curious if we should however keep it as a default script, or move it to its own action. 

Sadly even with `bpy.app.translations`, headers don't seem to update automatically. Will have to debug further.

StandingPadAnimations:
Determined the issue regarding headers; turns out MCprep was silently falling back to "direct i18n" (i.e. using `gettext` over `bpy.app.translations`), but I didn't have a dev build enabled so there were no logs.

Once that issue is fixed, everything changes in real time. We should be good to go to merge this.

StandingPadAnimations:
@TheDuckCow do you think we're good to merge by the end of this week?

StandingPadAnimations:
Just realized we can also use `polib` to build MO files at build time, so I've added that to the default case. For organization reasons, I've removed the MO files from the repo.

StandingPadAnimations:
Looking further, `polib` is actually a single file Python library, so we could just add the file directly in `action-scripts` and call it a day.

TheDuckCow:
I definitely support the idea of some utility or auto-compile options with bpy-addon-build, this PR is already quite large so maybe worth doing in a follow up PR.

TheDuckCow:
Regarding being PR-ready @StandingPadAnimations, is everything in order..? I just tried the branch out as-is (not commits or re-running any commands), and found that that some of the simplified chinese lines which are defined in Po don't seem to be translated in the application itself. Do we still need to do the dictionary registration step / load via po per bpy docs? I'd say that would be critical for this PR, otherwise the translations system isn't quite useable

Screen Shot 2024-02-19 at 7 10 24 PM

Would be good to show some screenshots of what you are seeing from your end

StandingPadAnimations:
You need to add `translate` to the list of actions when building (since we rebuild all MO files and generate the translation dictionary, which adds to build times). `bab -b dev translate` will make a dev build with i18n support. I just forgot to change the developer docs to account for this

TheDuckCow:
Yup, that has it working for me! Thanks for clarifying that, and updating the docs. Should probably update the `push_latest.sh` script as well, but I could do that in a follow up PR after this one. Will code review now, let's lock this in and do incremental improvements :+1:

MCprep_addon/mcprep_ui.py
TheDuckCow: Resolving this, knowing these calls are necessary to generate the template files.

action-scripts/default.py
```diff
@@ -1,3 +1,7 @@
 from default_scripts import ignore_filters
 
-ignore_filters.main()
+def main():
```
TheDuckCow, line 3: when do you want to roll out the new bpy-addon-build, since the args will change for that (needing to pass in context). Could do in a small follow up PR

action-scripts/libs/polib.py
TheDuckCow, line 1: I gather this is not something you wrote yourself, is that right? Is this something that should actually be pip installed instead? I'm wearing of checking in 1800 lines of code. If it's been manually modified, is there a way we could manage the modifications separately from the source (sort of like how you can mock/replace functions in unittest)?

docs/i18n/developer_notes.md
TheDuckCow, line 1: Thanks for adding the details here 👍

docs/i18n/translating.md
```diff
@@ -0,0 +1,72 @@
+# Translation Guidelines
+To make sure translations are up to par with the standards we hold on the MCprep repo, all translators must follow these guidelines.
+1. No AI for translations, such as Google Translate, ChatGPT, etc. AI translations are unreliable, differ massively in terms of quality between different languages, and don't take context properly into account. 
+2. All translations must be confirmed by a regular user who knows the language in question. This is so we can verify that the translation is up to par.
+
+Failure to follow these guidelines may prevent one from being able to contribute translations for MCprep in the future.
```
TheDuckCow, line 6: I think we can remove this line, as it sets a bit of a negative tone. As maintainers, we'll always have the right to refuse contributions when it comes down to it. If we do face it, let's approach it from a teach and learn perspective.

poetry.lock
```diff
@@ -1,31 +1,81 @@
...
-python-versions = "*"
+python-versions = ">=3.7"
 files = [
-    {file = "docopt-0.6.2.tar.gz", hash = "sha256:49b3a825280bd66b3aa83585ef59c4a8c82f2c8a522dbe754a8bc8d08c85c491"},
+    {file = "exceptiongroup-1.2.0-py3-none-any.whl", hash = "sha256:4bfd3996ac73b41e9b9628b04e079f193850720ea5945fc96a08633c66912f14"},
+    {file = "exceptiongroup-1.2.0.tar.gz", hash = "sha256:91f5c769735f051a4290d52edd0858999b57e5876e9f85937691bd4c9fa3ed68"},
 ]
 
+[package.extras]
+test = ["pytest (>=6)"]
```
TheDuckCow, line 77: Where is pytest being used? Is this something pulled in by one of the dependent libraries? MCprep itself is using the built in unittest library.

requirements.txt
```diff
@@ -1,3 +1,2 @@
 bpy-addon-build>=0.3.0
 fake-bpy-module-2.80==20230117
-flake8==5.0.4
```
TheDuckCow, line 3: In contrast, I do think we should set flake8 as an expectation. I want to set it up as a github runner. even if we don't have the quota to reliably run unittests remotely, would be nice to perform automated flake8 syntax and style checks as one good method to safeguard future changes.

TheDuckCow (commented):
Only reason for not directly approving is due to my comment on the `action-scripts/libs/polib.py` file and whether that should be added to requirements.txt, other comments are more clarifications or minor things. We're close! And to reaffirm this, translations are working well for me now.

TheDuckCow:
(You can re-request my review one you have replies, so I don't miss them)

action-scripts/libs/polib.py
StandingPadAnimations: `polib` is a library that can be pip installed, but it's also single file. For convenience purposes, I've copied the file to our local tree.

poetry.lock
StandingPadAnimations: Might be from when I initially had `polib` as one of the dependencies defined in `pyproject.toml`. I'll regenerate the lock file

StandingPadAnimations: Ok I've regenerated the file and it still has it. My best guess is that it comes from `flake8` (which uses pytest). I think it should be fine to keep it though, as it shouldn't interfere with what we have.

action-scripts/default.py
StandingPadAnimations: I think I'll roll it out after this PR. I don't want to drag this PR on for too long, and BpyBuild still needs a little more time before the API is released

poetry.lock
TheDuckCow: Gotcha, thanks for checking - I'm fine to leave it as-is then

requirements.txt
TheDuckCow: Nevermind this comment - I see you have flake8 in the main config.

action-scripts/libs/polib.py
TheDuckCow: Can we just do pip install, so that way we can get updates (and are never tempted to directly modify its contents)? It's already a requirement that we are using a virtual env given we have the blender addon builder.

TheDuckCow (changes requested):
Ok, hopefully a simple change, [per discussion](https://github.com/Moo-Ack-Productions/MCprep/pull/538#discussion_r1495249646) request is to move to pip install and not check in the po lib. Once we have that, I'll be good to approve

StandingPadAnimations:
Done

action-scripts/libs/polib.py
TheDuckCow: Thank you!

TheDuckCow (approved):
Alright, that's an approval! Big step forward here, fantastic work and thanks for your perseverance trying all these different approaches. Will nicely enable future work such as use of po editors, excited for us to democratize that soon.

StandingPadAnimations:
Alright, updated with `milestone-3-6-0` and updated the POT file, I'll merge this in now

PR #544, opened 2024-02-20
Updated contributing guide readme

TheDuckCow:
Updated per upcoming changes being merged in with the translations branch, to ensure developers install requirements for the project (if they are using poetry, I gather they would already be set, just trying to make sure our most simple getting started instructions make it as straight forward as possible)


StandingPadAnimations (approved):
Looks good

PR #549, opened 2024-03-01
Error Migration and Refactoring for 3.6.0

StandingPadAnimations:
This PR is mainly to migrate errors in MCprep to use the new `MCprepError` class. At some point, we realized the need for being able to return objects with messages, so we also added an optional `msg` field to `MCprepError`.

In this PR, we also perform any general refactoring that may crop up as needed. Most of these are related to error handling, but there are a few that for the sake of getting out of the way I've done, which are:
- Removal of Blender Internal and Blender 2.7X code in `world_tools.py`
- Removal of 2.7X link and append operations in `bAppendLink`, as well as deprecate the `active_layer` argument

The following functions have been refactored to use the new error type:
- `convert_mtl`: `Optional[bool]` -> `Optional[MCprepError]`
- `enable_obj_importer`: `Optional[Bool]` -> `Union[OBJImportCode, MCprepError]`
- `bAppendLink`: `bool` -> `Optional[MCprepError]` 
- `open_program`: `Union[int, str]` -> `Optional[MCprepError]`


MCprep_addon/materials/default_materials.py
```diff
@@ -1,219 +0,0 @@
...
-				err = sync_default_material(context, mat, material_name, self.engine.upper()) # no linking
-				if err:
-					env.log(err)
-			except Exception as e:
-				print(e)
-
-		return {'FINISHED'}
-
-
-class MCPREP_OT_create_default_material(bpy.types.Operator):
```
TheDuckCow, line 146: Just sanity checking - I see the code is commented out in the UI, was this ever exposed such that users could have been using these material operators from this file? (outside of manually stumbling onto the code).

I just want o ensure it's safe to delete - I'm all for cleaning up dead code, just want to make sure I understand if there's any loss of functionality we need to communicate

MCprep_addon/util.py
```diff
@@ -228,7 +227,11 @@ def min_bv(version: Tuple, *, inclusive: bool = True) -> bool:
...
 
 def bv28() -> bool:
-	"""Check if blender 2.8, for layouts, UI, and properties. """
+	"""
+	Check if blender 2.8, for layouts, UI, and properties. 
+	
+	Deprecated in MCprep 3.5, but kept to avoid breakage for now...
+	"""
 	env.deprecation_warning()
 	return min_bv((2, 80))
```
TheDuckCow, line 236: All of these references have been removed now. If not on this 3.6.0 branch, then check in dev (maybe merging dev into the milestone branch), in which case either way I think this function can just be removed.

MCprep_addon/world_tools.py
```diff
@@ -756,41 +781,7 @@ def prep_world_eevee(self, context: Context) -> None:
 
 		# Renders faster at a (minor?) cost of the image output
 		# TODO: given the output change, consider make a bool toggle for this
-		bpy.context.scene.render.use_simplify = True
-
-	def prep_world_internal(self, context):
```
TheDuckCow, line 761: Oof, definitely good to delete this code now! Nice cleanup

TheDuckCow (commented):
Nice changes! I see it's still in draft, so just leaving as comments for now. please re-request review once ready for more input

StandingPadAnimations:
I think I'll wait till #555 is merged before continuing to work on this, since both #555 and this PR have changes in `world_tools.py`. One #555 is merged, I'll update this branch with `minestone-3-6-0` and continue working on this

StandingPadAnimations:
I think we should merge this into the CommonMCOBJ branch so we can knock out 2 birds with one stone

StandingPadAnimations:
Alright, merging into the CommonMCOBJ branch

PR #555, opened 2024-03-24
Implementing CommonMCOBJ Support

### LINKED ISSUE #553: Moving Towards CommonMCOBJ
As we get closer to finalizing [CommonMCOBJ V1](https://github.com/CommonMCOBJ/CommonMCOBJ), I think it's time we start planing out how we should introduce support. Since the initial spec is simple, I think we'll be able to add support for 3.6. On the CommonMCOBJ side, there is now an OBJ exporter (forked from jmc2OBJ) called [cmc2OBJ](https://github.com/CommonMCOBJ/cmc2obj), which acts as the reference implementation for the CommonMCOBJ spec, so verifying compatibility with CommonMCOBJ should be trivial. 

While CommonMCOBJ would make the current Mineways/jmc2OBJ options obsolete, I don't think it would be a huge maintaince burden to at keep them in MCprep, however they will become redundent for most use cases as time passes. As sucb, here's what I think we should do for those options:
- Add some form of analytics tracking CommonMCOBJ usage, to get an idea of how often its being used
- Add a CommonMCOBJ option where we have Mineways and jmc2OBJ options (replacing `(none)`) in 3.6
- In MCprep 4.0 (assuming we're not skipping 3.7-3.9), if adoption is going well, then create a legacy option in user preferences (for exposing distinct OBJ exporter options), and completely eliminate user facing OBJ exporter selection by default
    - This would not eliminate special handling for individual exporters, as CommonMCOBJ gives that information anyway

Overall, I think we're ready to start adopting support for it in MCprep

StandingPadAnimations:
This implements parsing and handling for CommonMCOBJ headers in OBJs, and reworks exporter checking code to be more exporter agnostic.

[CommonMCOBJ](https://github.com/CommonMCOBJ/CommonMCOBJ) is a new exporter standard that Mineways and jmc2OBJ are currently working on supporting. CommonMCOBJ provides more information for MCprep to utilize for OBJs, and is beneficial overall in the long run.

When an OBJ with a CommonMCOBJ header is imported, MCprep will parse the header and add all parsed options as custom properties to the OBJ. In addition, as the plan is to phase out global exporter options in MCprep 4.0, OBJs without the CommonMCOBJ header will gain an additional `MCPREP_OBJ_EXPORTER` property, for backwards compatibility in MCprep 4.0. OBJs without either the CommonMCOBJ properties or the `MCPREP_OBJ_EXPORTER` will fall back to the globally set exporter (though globally set exporters will become a feature behind a legacy flag in MCprep 4.0)

When it comes to exporters, CommonMCOBJ is exporter agnostic, and provides a key for the exporter name. To handle that, all exporter checks in MCprep will now be done with an enum rather than pure strings. The options in the enum are:
- Supported exporters with CommonMCOBJ (`Mineways` and `Jmc2OBJ`)
- [`Cmc2OBJ`](https://github.com/CommonMCOBJ/cmc2obj) (which for the most part is treated as unsupported, but exists as a distinct option for testing purposes)
- `ClassicMW` (Mineways before CommonMCOBJ) and `ClassicJmc` (jmc2OBJ before CommonMCOBJ), which uses either the globally set exporter in the N panel or the `MCPREP_OBJ_EXPORTER` attribute on the active object
- Fallback `Unknown` for all other exporters

EDIT 30/03/2024: `commonmcobj_parser.py` is, unlike the rest of MCprep, licensed under the BSD 3-Clause license to allow developers to reuse without worrying about GPL. As such, the README has been updated to account for that, and `LICENSE-3RD-PARTY.txt` has also been added.

Closes #553


TheDuckCow:
Great to have this support! Small but important request from my side: can you implement at least one parsing test? I suggest you do so by adding manually created, simple obj file with and without a header which is saved into the testing directories of the addon (so it's not shipped). This is the number 1 way we can ensure it continues to function as intended and to spec, though I'm ok if the test isn't exactly exhaustive over all scenarios.

Let me know if you have trouble running the tests, they should be nicely cross platform by now though.

MCprep_addon/commonmcobj_parser.py
```diff
@@ -0,0 +1,181 @@
...
+    the header exists.
+    
+    f: TextIO
+        File object
+    
+    Returns:
+        - CommonMCOBJ object if header exists
+        - None otherwise
+    """
+    lines = f.readlines()
```
TheDuckCow, line 161: This is pretty memory intensive and almost always will load in more data than we need, since the header is at the top. This would load the entire obj into python memory (doublely bloated) while using an iterator would only have individual lines at a time up until we exit out.

Let's do a with-clause iterator where we read one line at a time, stopping after we reach a the `# COMMON_MC_OBJ_END` or maybe after reading 10 lines with no comments (`#`).

With each line, we can append to a list which we later .join() into a string (if relevant).

MCprep_addon/materials/prep.py
```diff
@@ -305,9 +306,10 @@ def execute(self, context):
 				"Nothing modified, be sure you selected objects with existing materials!"
 			)
 
-		addon_prefs = util.get_user_preferences(context)
 		self.track_param = context.scene.render.engine
-		self.track_exporter = addon_prefs.MCprep_exporter_type
+		# TODO: Rework exporter tracking
+		# addon_prefs = util.get_user_preferences(context)
+		# self.track_exporter = addon_prefs.MCprep_exporter_type
```
TheDuckCow, line 312: I think the tracker is still worthwhile, as it helps us understand the ecosystem of what's being used and how to best support everyone.

Main question is: do we have the header auto assign a value of e.g. jmc2obj to the exporter type, or should we add two more types which are  jmc2obj_cmobj (or something like that) and the same for Mineways? I'm a bit torn. It could be quite nice to know how well adopted the common format is, to know what feature work and whether it applies to a good % of the MCprep userbase.

MCprep_addon/world_tools.py
```diff
@@ -110,51 +116,127 @@ def texture_type(self):
...
-					"# File type: Export tiles for textures to directory textures",
-					"# File type: Export individual textures to directory tex"
-				)
-				print('"{}"'.format(header))
-				if header in atlas:  # If a texture atlas is used
-					obj_header.set_atlas()
-				elif header in tiles:  # If the OBJ uses individual textures
-					obj_header.set_seperated()
-				return
+			cmc_header = parse_header(obj_fd)
```
TheDuckCow, line 207: Also based on the implementation above, wouldn't the readlines() line within parse_header clear out the remainder of readlines below? Would be good to double check all tests still pass as you're developing here.

MCprep_addon/commonmcobj_parser.py
```diff
@@ -0,0 +1,181 @@
+from dataclasses import dataclass
+from enum import Enum
+from typing import List, Optional, Tuple, TextIO
+
+class CommonMCOBJTextureType(Enum):
+    ATLAS = "ATLAS"
+    INDIVIDUAL_TILES = "INDIVIDUAL_TILES"
+
+@dataclass
+class CommonMCOBJ:
```
TheDuckCow, line 10: Great to have this general dataclass representation. Should we also (or instead) have a blender property group version too?

I imagine the most useful thing would be to ensure the entire spec can be saved to the the root of a world import, so it can be opened later and the user could e.g. extend the world import or re-trigger opening mineways/jmc2obj with the same settings.

We could save the content as either pre-parsed details, or maybe it would be safer in a way to save the whole header as a text block, so that the MCprep addon always attemtps to parse, but in case the user has an outdated mcprep instance compared to the spec, the full spec details are still saved in case the user updates MCprep later and then goes back to reopen this old file. If we do that, much less of a need for custom property groups.

(if this is already happening, sorry for missing it - maybe link me)

TheDuckCow (commented):
Just leaving some initial comments

MCprep_addon/materials/prep.py
StandingPadAnimations: I think we could use the output of `get_exporter` (something like `get_exporter(context).value` to retrive the underlying string from the enum, assuming it's not `None`) as it would be a 1-1 mapping, although we'd then have to concert the Mineways and jmc2OBJ references in the tracker to `mineways-c` (Mineways before CommonMCOBJ) and `jmc2obj-c` (jmc2OBJ before CommonMCOBJ). It would also allow tracking the use of unknown exporters as well

MCprep_addon/commonmcobj_parser.py
StandingPadAnimations: The dataclass is used to create properties after importing the OBJ, with all values used (mostly as backward/forward compatibility in case we introduce new features that depend on more of the header)

MCprep_addon/commonmcobj_parser.py
StandingPadAnimations: Resolved in [`ab2e163`](https://github.com/Moo-Ack-Productions/MCprep/pull/555/commits/ab2e1635fb5d5752e98971dd833ccdb00ff54014)

StandingPadAnimations:
@TheDuckCow by the way, should we implement the "legacy" flag for OBJ imports now and have it enabled by default (then disable it later on) or should we wait until MCprep 4.0 (which if I recall correctly will have UI refactorimg anyway)

MCprep_addon/materials/prep.py
StandingPadAnimations: Thinking abut this further, I think we should use the string in the `exporter` key of the header. If we use the enum, we won't be able to know if a new OBJ exporter is gaining traction (and thus could be better supported in MCprep), but the exporter key would make it more clear what specific exporters are being used

StandingPadAnimations:
Alright, normally I'm not a fan of dual licensing, but I've decided to put the CommonMCOBJ parser under the more permissive BSD 3-Clause license (compatible with GPL), mainly to make it easier for developers to reuse in their own projects without worrying about GPL. This should be fine as the parser doesn't interact with BPY and BSD 3-Clause is compatible with GPL.

If wanted, I'm fine with changing it back to GPL, but I'd rather have it under a more permissive license (especially as it doesn't depend on Blender and could be reused anywhere)

TheDuckCow:
Yup I'm supportive of this, thanks for thinking of it. Will review when I'm back home later On Mar 30, 2024 12:48 PM, Mahid Sheikh ***@***.***> wrote:
Alright, normally I'm not a fan of dual licensing, but I've decided to put the CommonMCOBJ parser under the more permissive BSD 3-Clause license (compatible with GPL), mainly to make it easier for developers to reuse in their own projects without worrying about GPL. This should be fine as the parser doesn't interact with BPY and BSD 3-Clause is compatible with GPL.
If wanted, I'm fine with changing it back to GPL, but I'd rather have it under a more permissive license (especially as it doesn't depend on Blender and could be reused anywhere)

StandingPadAnimations:
Alright, I've added `LICENSE-3RD-PARTY.txt` (not the best name as `commonmcobj_parser.py` isn't 3rd party, but oh well) to indicate the license for `commonmcobj_parser.py` and added clarification in the README regarding it.

StandingPadAnimations:
I've added some basic forward compatibility by adding the original header to the object

MCprep_addon/materials/prep.py
TheDuckCow: Remind me, what would be the values we potentially get in the exporter? that's the mineways-defined or jmc2obj-defined names for themselves, right?

I think having something like jmc2obj_cmobj would be fantastic or at least _somehow__ knowing which imports comes form the commonmcobj spec in the analytics data. A canned suffix added to the `exporter` string could do. Knowing the depth of adoption of the new standard helps understand how many people can make use of a tool we build that depends on that (if e.g. less than 10% of obj imports are using the spec, that could be good rationale to not yet spend ages working on the bridge tool just yet for instance).

TheDuckCow (changes requested):
I havne't been able to live test (I keep getting stuck with partially updated bpy-build updates that mess with my test script/compiling rn), but in addition to the comments below, this is what I'd love to see before merging:

- let's get the tracker assignments updates so we still get the exporter names (ideally with a `_cmcobj` suffix or similar, so we can di sql grouping/breakdowns later). For legacy imports, we should still get the legacy names as we have before, so we can compare equally.
- Having a single source which has the responsibility of what represents the "world", so that all the raw data isn't duplicated everywhere. I think that would mean having the data stored in the collection that gets generated, or optionally in a top level parent (empty?) object. If someone moves things around outside the collection, they probably aren't expecting it to stay connected anymore anyways, as that's sort of like breaking the link.
  - Another option here could also be, having the collection of the world retain the raw text data of the commobj spec, saved as a plaintext property. Then, I'd be _more_ ok with individual objects having their own properties applied as they could be reloading/updated/read from the parent collection/object.
  - But even still, that just results in a lot of duplicated data for each mesh (like paths)
- At least one unit test covering the commoncobj _parser.py file. We have increasingly decent test code coverage with MCprep. I think it would be great if we added at least one that tried to make some coverage of the common obj implementation. I'd be ok even if it was a super simple one, one that just calls the main functions, passing in onw or two known scenarios, and checking we get the known output assignments in the class too.
- A step further would also be adding a test for the world importer operator, with known world exports (tiny in size, like literally just a single block or two) from Mineways + jmc2obj in the new format, and checking those in test objs to run against and verify we get the correct params set (not checking in textures since they shouldn't be necessary to the test code).

MCprep_addon/materials/prep.py
StandingPadAnimations: > Remind me, what would be the values we potentially get in the exporter? that's the mineways-defined or jmc2obj-defined names for themselves, right?

It would be the name of the exporter, so `jmc2obj`, `mineways`, `helpicantexitvim`, etc.

StandingPadAnimations:
> * let's get the tracker assignments updates so we still get the exporter names (ideally with a `_cmcobj` suffix or similar, so we can di sql grouping/breakdowns later). For legacy imports, we should still get the legacy names as we have before, so we can compare equally.

Alright, sounds good

> * Having a single source which has the responsibility of what represents the "world", so that all the raw data isn't duplicated everywhere. I think that would mean having the data stored in the collection that gets generated, or optionally in a top level parent (empty?) object. If someone moves things around outside the collection, they probably aren't expecting it to stay connected anymore anyways, as that's sort of like breaking the link.

Are custom properties for collections supported in older versions of Blender? I'm fine with moving from adding properties to all objects to just collections, but that's my main concern.


> * At least one unit test covering the commoncobj _parser.py file. We have increasingly decent test code coverage with MCprep. I think it would be great if we added at least one that tried to make some coverage of the common obj implementation. I'd be ok even if it was a super simple one, one that just calls the main functions, passing in onw or two known scenarios, and checking we get the known output assignments in the class too.
> * A step further would also be adding a test for the world importer operator, with known world exports (tiny in size, like literally just a single block or two) from Mineways + jmc2obj in the new format, and checking those in test objs to run against and verify we get the correct params set (not checking in textures since they shouldn't be necessary to the test code).

[`cmc2OBJ`](https://github.com/CommonMCOBJ/cmc2obj) is the reference implementation of CommonMCOBJ, so ideally we should use that for CommonMCOBJ tests

TheDuckCow:
> Are custom properties for collections supported in older versions of Blender? I'm fine with moving from adding properties to all objects to just collections, but that's my main concern.

I'm not sure, would be worth checking into. I'm thinking a couple steps ahead to the import bridge, where different chunks of the world would be loaded together (probably larger than mc chunks though). I'd expect they would all live within one "world" collection, but each individual chunk should have its own "door". That could be a sub collection or event an empty parent. What do you think, would it be very annoying if all world imports had a collection and also a root empty parent? When that root is selected we show special drawing features for instance potentially. Although the same could be also accomplished through looking at the active objects parent collection/object too. 



> [`cmc2OBJ`](https://github.com/CommonMCOBJ/cmc2obj) is the reference implementation of CommonMCOBJ, so ideally we should use that for CommonMCOBJ tests (it has some known issues though with material exports, but it would only be used for CommonMCOBJ tests anyway)

Totally fine and onboard with having the core tests stay over there, that's fair. I'd like *some* kind of integration test in MCprep just to make sure that everything is glued together ok. Could be an extension of a world import operator unit test, hence the idea of having a small checked in obj with the spec. 

Thanks for working on these!

StandingPadAnimations:
> I'm not sure, would be worth checking into. I'm thinking a couple steps ahead to the import bridge, where different chunks of the world would be loaded together (probably larger than mc chunks though). I'd expect they would all live within one "world" collection, but each individual chunk should have its own "door". That could be a sub collection or event an empty parent. What do you think, would it be very annoying if all world imports had a collection and also a root empty parent? When that root is selected we show special drawing features for instance potentially. Although the same could be also accomplished through looking at the active objects parent collection/object too.

I think an empty acting as the parent for chunks is a good idea, it also makes moving OBJs easier for users. 

One issue I realized last night about collection properties it that if the user has a non-OBJ object in said collection (like a plane or rig), it would be interpreted as part of the OBJ, so an empty is a better idea IMO.

> Totally fine and onboard with having the core tests stay over there, that's fair. I'd like some kind of integration test in MCprep just to make sure that everything is glued together ok. Could be an extension of a world import operator unit test, hence the idea of having a small checked in obj with the spec.

I'm fine with integration tests in MCprep for CommnMCOBJ, just that we should use OBJs exported from `cmc2OBJ` for tests as `cmc2OBJ` is the reference implementation of CommonMCOBJ, so ideally exporters should replicate their common header output to match `cmc2OBJ`. 

Btw, on the topic of the tracker, is there any extra work that needs to be done tracker side to support new exporter strings, or is it fine as is?

TheDuckCow:
Hey @StandingPadAnimations getting back to this. One question, since we're nearing to a release. Do we want to merge the milestone 360 branch into dev before or after this common mc obj branch? I'm in favor of merging sooner so we can get everything in order, and avoid us having to replicate changes across multiple branches.

As for the changes above, I'd say: let's go ahead with empty parents for chunks. 

> integration tests in MCprep for CommnMCOBJ, just that we should use OBJs exported from cmc2OBJ for tests as cmc2OBJ is the reference implementation of CommonMCOBJ, so ideally exporters should replicate their common header output to match cmc2OBJ.

Works with me 👍 

> Btw, on the topic of the tracker, is there any extra work that needs to be done tracker side to support new exporter strings, or is it fine as is?

Nope, it accepts anything you feed into it. Which is why I'm picky in the the review stage as it'd be a pain to override later :)

TheDuckCow:
Just remember to re-request my review once it's at a stage for me to hands on with the empty approach.

StandingPadAnimations:
> Hey @StandingPadAnimations getting back to this. One question, since we're nearing to a release. Do we want to merge the milestone 360 branch into dev before or after this common mc obj branch? I'm in favor of merging sooner so we can get everything in order, and avoid us having to replicate changes across multiple branches.
> 
> As for the changes above, I'd say: let's go ahead with empty parents for chunks.
> 
> > integration tests in MCprep for CommnMCOBJ, just that we should use OBJs exported from cmc2OBJ for tests as cmc2OBJ is the reference implementation of CommonMCOBJ, so ideally exporters should replicate their common header output to match cmc2OBJ.
> 
> Works with me 👍
> 
> > Btw, on the topic of the tracker, is there any extra work that needs to be done tracker side to support new exporter strings, or is it fine as is?
> 
> Nope, it accepts anything you feed into it. Which is why I'm picky in the the review stage as it'd be a pain to override later :)

I'm fine with merging `milestone-3-6-0` before this one. I'll just change the base branch to `dev` and we can go ahead with the merge. An approximate ETA for this would be the end of next week, so it's best to start preparing.

With the merge to dev, I think we should get at least 1 or 2 Release Candidates (RCs) before release, considering MCprep 3.5+ ignores RCs in the updater now (allowing us to make public releases). MCprep 3.6 includes a lot of major changes, and we haven't really tested in production yet. If we merge today or tomorrow, I can get a RCs made tomorrow or Monday (for context, RCs for MCprep 3.5 were made on Sundays).

StandingPadAnimations:
Alright, changed the base to dev, we can go ahead with merging `milestone-3-6-0`

StandingPadAnimations:
In an effort to try and implement the empty at OBJ center feature, I've come across an issue with how OBJ's are handled in Blender,

Normally `export_bounds_*` should correspond to the OBJ geometry, and they do, but only as a local transform (when going into edit more on an OBJ and checking the transforms). However, when creating the empty, the empty's coordinates are in world space, which is completely different. Not sure how I can resolve this

TheDuckCow:
Hey @StandingPadAnimations if I'm understanding the problem correctly, maybe we can solve it by adding an extra property/meta data field on each chunk obj which is the chunk offset or index or something like that?

That should give us the information to do the practical translation from one coordinate system to the other. What do you think?

StandingPadAnimations:
I would rather avoid adding additional keys to the spec, mostly because V1 hasn't even been fully adopted yet (and V2 isn't due till 2025 anyway). Plus, I think this is more of a Blender issue then an OBJ one. 

The thing is, export bounds are correct in the exported OBJs, but only if you go in edit mode and look at local transforms (global transforms are where the issue lies). If we could somehow set the location of an empty based on those local transforms, it'd work fine as far as I know.

TheDuckCow:
I wouldn't think of this as part of the spec, but something specific to MCprep which is an internal nuance of how it handled the spec. So in terms of having a pure python presentation, this property would not live there. This would not be anything any other implementation of the spec outside MCprep would even make sense for, just some intermediate caching. The property could even potentially be save based on naming the chunk empty itself, so "1_1"or "0_-5" couldnbecome.the empty name, and by parsing that we could know the offset. 

The other thought could be: *do* we even need to worry about this offset? If we parent the meshes with initial transforms to the empty location, we can always negate it by checking the transform of any one single object to get back at the offset. So maybe we don't need to save any extra data at all? (That would break if anyone applied all transforms to make it global, whereas asking people to not rename the names of Empty chunks is probably a little bit more reasonable and understandable)

StandingPadAnimations:
Sounds good. To clarify, what exactly would chunk offset refer to?

TheDuckCow:
Sorry missed this comment. It's whatever you need to have it refer to in order to get around the change from one coordinate reference point to the other :) So whatever you choose that makes sense, I trust your judgement on that. Like I said in my mind, could either be chunk counts or even direct block unit offsets. Direct offsets are probably safer, in case we eventually implement set extensions which allow for variable size chunks or something like that

StandingPadAnimations:
So by direct offsets, you mean parenting the OBJ to the generated empty at say `(0,0,0)` so we could determine the actual offset if needed? That could work, but the origin of OBJs is also at `(0,0,0)`, so the offset would always return 0.

TheDuckCow:
@StandingPadAnimations 

> So by direct offsets, you mean parenting the OBJ to the generated empty at say `(0,0,0)` so we could determine the actual offset if needed? That could work, but the origin of OBJs is also at `(0,0,0)`, so the offset would always return 0.

Fair to make sure we're totally explicit here. So, I'm imagining something akin to this file setup:
Screen Shot 2024-05-08 at 9 24 03 PM

Does this solve the challenge you were facing in the end? In the end, it's just whatever value is needed to get back at the spec compared to whatever transform setup we establish. The offset could just mean whatever translation amount was needed to be applied to the children when parenting to the empty for instance. I did a manual object property here, but presuming that we'd have the rest of the commobobj format also saved to this empty object. For typical world imports, it'll be a single root empty. For cases where we support batch loading of chunks (hmm similar to what MiEx does I guess), then there will be multiple empties for each chunk. The user could even choose to move this chunks to other collections (think, disabling far collections as you focus on just the near).

StandingPadAnimations:
Makes a lot more sense now, thanks! I've also checked to see how coordinates match in Blender, and it seems like if we inverse the bound's `Z` coordinate, the empty would be in the correct location. At least, that's how `jmc2OBJ`/`cmc2OBJ` behave, I haven't tested with Mineways yet. So in essence, given a bound `(X, Y, Z)`, the empty would be at `(X, -Z, Y)` (`Z` swapped with `Y` since `Y` is up in Minecraft)

StandingPadAnimations:
(If Mineways doesn't show the same behavior, that might be something we could look into clarifying in V2, but I assume Mineways does)

StandingPadAnimations:
Alright, Mineways also shows the same behavior (although it should be noted both the initial `Z` and the `Z` offset need to be negated, but that's a given), so we're good. I'll still add a specification in V2 just in case.

TheDuckCow:
Fantastic! Will happily await once you're ready to re-request me so I can try out with a couple worlds. Thanks a ton again for driving this forward!

StandingPadAnimations:
Finally got this in (though it took a bit of pain as Blender's matrix transforms are weird)

StandingPadAnimations:
Also I just realized this PR is reported to be larger than the actual changes, probably due to that accidental squash merge

StandingPadAnimations:
Updated this branch with the remerge in `dev`. Sadly, we still have some extra commits, but it should be more manageable now

StandingPadAnimations:
@TheDuckCow we're good to go for review. Just a forewarning that there's bit more stuff to review now as #549 was merged into this branch 2 weeks ago.

TheDuckCow:
Hey @StandingPadAnimations did a very quick test and ran into this issue below while importing an old world I have.

```
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.1/scripts/addons/MCprep_addon/tracking.py", line 887, in wrapper
    res = function(self, context)
          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.1/scripts/addons/MCprep_addon/world_tools.py", line 772, in execute
    self.track_exporter = addon_prefs.MCprep_exporter_type  # Soft detect.
                          ^^^^^^^^^^^
NameError: name 'addon_prefs' is not defined
```

Then, tried the test runner. It seems we have 19 tests failing, so would try running through this again when you have a chance. Maybe it's an issue with handling OBJs that don't have any spec data?
```
(3.6.2)   	all_tests	65	2	19	test_prep_material_animated;test_prep_missing_pbr_passes;test_prep_pbr_seus;test_prep_pbr_specular;test_prep_simple;test_replace_missing_images_animated;test_replace_missing_images_fixed;test_replace_missing_images_moved_blend;test_replace_missing_images_name_incremented;test_swap_texture_pack [jmc2obj_noprep];test_swap_texture_pack [jmc2obj_withprep];test_swap_texture_pack [Mineways_withprep];test_meshswap_world_jmc;test_meshswap_world_mineways_combined;test_meshswap_world_mineways_separated;test_world_import_jmc_full;test_world_import_mineways_combined;test_world_import_mineways_separated;test_convert_mtl_simple
tests took 13s to run, ending with code 1
```

MCprep_addon/spawner/meshswap.py
```diff
@@ -517,9 +517,9 @@ def draw(self, context):
 	@tracking.report_error
 	def execute(self, context):
 		tprep = time.time()
-		# TODO: Rework exporter tracking
-		# addon_prefs = util.get_user_preferences(context)
-		# self.track_exporter = addon_prefs.MCprep_exporter_type
```
TheDuckCow: Didn't look deeply, maybe this is the one needing an update?

StandingPadAnimations:
> Hey @StandingPadAnimations did a very quick test and ran into this issue below while importing an old world I have.
> 
> ```
> Traceback (most recent call last):
>   File "/Users/patrickcrawford/Library/Application Support/Blender/4.1/scripts/addons/MCprep_addon/tracking.py", line 887, in wrapper
>     res = function(self, context)
>           ^^^^^^^^^^^^^^^^^^^^^^^
>   File "/Users/patrickcrawford/Library/Application Support/Blender/4.1/scripts/addons/MCprep_addon/world_tools.py", line 772, in execute
>     self.track_exporter = addon_prefs.MCprep_exporter_type  # Soft detect.
>                           ^^^^^^^^^^^
> NameError: name 'addon_prefs' is not defined
> ```
> 
> Then, tried the test runner. It seems we have 19 tests failing, so would try running through this again when you have a chance. Maybe it's an issue with handling OBJs that don't have any spec data?
> 
> ```
> (3.6.2)   	all_tests	65	2	19	test_prep_material_animated;test_prep_missing_pbr_passes;test_prep_pbr_seus;test_prep_pbr_specular;test_prep_simple;test_replace_missing_images_animated;test_replace_missing_images_fixed;test_replace_missing_images_moved_blend;test_replace_missing_images_name_incremented;test_swap_texture_pack [jmc2obj_noprep];test_swap_texture_pack [jmc2obj_withprep];test_swap_texture_pack [Mineways_withprep];test_meshswap_world_jmc;test_meshswap_world_mineways_combined;test_meshswap_world_mineways_separated;test_world_import_jmc_full;test_world_import_mineways_combined;test_world_import_mineways_separated;test_convert_mtl_simple
> tests took 13s to run, ending with code 1
> ```

Able to reproduce on my end, seems like an error when I reenabled tracking

StandingPadAnimations:
Fixed, it was due to some missing `addon_prefs = util.get_user_preferences(context)` lines. Not sure why past me removed those

TheDuckCow:
All good, I think it was due to saying we'd get back to some analytics changes. Reviewing shortly, thanks for the quick fix

MCprep_addon/util.py
```diff
@@ -228,7 +240,11 @@ def min_bv(version: Tuple, *, inclusive: bool = True) -> bool:
...
 
 def bv28() -> bool:
-	"""Check if blender 2.8, for layouts, UI, and properties. """
+	"""
+	Check if blender 2.8, for layouts, UI, and properties. 
+	
+	Deprecated in MCprep 3.5, but kept to avoid breakage for now...
+	"""
 	env.deprecation_warning()
 	return min_bv((2, 80))
```
TheDuckCow, line 249: I believe we have already removed all references to this, haven't we? I'd be in favor of just removing now.

MCprep_addon/materials/default_materials.py
TheDuckCow, line 1: Remind me, why is this file being removed, and was the functionality ever released to production? I see it seems similar to the sync materials logic, just trying to recall what we discussed before OK'ing removal.

MCprep_addon/world_tools.py
TheDuckCow, line 1: One thing we do need to fix is to make sure that the added parent empty is linked only into the same collection as the one created for the new import, see the screenshot here:

Screen Shot 2024-06-01 at 7 43 21 AM

TheDuckCow, line 1: Discussion question: Should we update the structure of legacy imports too, so that no matter what worlds you import, there's always a parent empty?

I feel users might find it confusing why sometimes their worlds have a parent empty, and sometimes they don't (since they won't be as keenly aware of the commonMCOBJ support under the hood).

Screen Shot 2024-06-01 at 7 55 12 AM

TheDuckCow, line 1: General comment that I do notice many unit tests are failing now though on this branch which aren't on dev, some function errors and some behavioral.

Failing (13/65):
```
test_prep_material_animated;test_replace_missing_images_animated;test_replace_missing_images_fixed [missing_resolved];test_replace_missing_images_moved_blend;test_replace_missing_images_name_incremented;test_swap_texture_pack [jmc2obj_noprep];test_swap_texture_pack [jmc2obj_withprep];test_swap_texture_pack [Mineways_withprep];test_meshswap_world_mineways_combined;test_meshswap_world_mineways_separated;test_convert_mtl_simple;test_world_import_mineways_combined;test_world_import_mineways_separated

```

e.g. this one is now failing:
```
FAIL: test_meshswap_world_mineways_separated (spawner_test.MeshswapTest)
  File "/Users/patrickcrawford/Documents/gits/mcprep/test_files/spawner_test.py", line 439, in test_meshswap_world_mineways_separated
    self.assertEqual(self.addon_prefs.MCprep_exporter_type, "Mineways")
AssertionError: 'jmc2obj' != 'Mineways'
- jmc2obj
+ Mineways
```

so there might be a problem or some flip with our header checking now. And another one, for the test `test_prep_material_animated`:

```
Failed to register scene update handler
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/tracking.py", line 887, in wrapper
    res = function(self, context)
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/materials/prep.py", line 594, in execute
    res, err = self.update_material(context, mat)
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/materials/prep.py", line 647, in update_material
    sequences.animate_single_material(
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/materials/sequences.py", line 104, in animate_single_material
    tile_path_dict, err = generate_material_sequence(
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/materials/sequences.py", line 172, in generate_material_sequence
    img_pass_dict = generate.find_additional_passes(image_path)
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/materials/generate.py", line 586, in find_additional_passes
    abs_img_file = bpy.path.abspath(image_file)
  File "/Applications/blender 3.6/Blender.app/Contents/Resources/3.6/scripts/modules/bpy/path.py", line 66, in abspath
    if path.startswith("//"):
AttributeError: 'PosixPath' object has no attribute 'startswith'
```

Remember that you can run individual tests at a time to reduce the noise of everything and get straight to the error: e.g. `python run_tests.py -t test_replace_missing_images_moved_blend` (for this test, I get the following error):

```
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/tracking.py", line 887, in wrapper
    res = function(self, context)
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/materials/material_manager.py", line 416, in execute
    res = generate.replace_missing_texture(passes[pass_name])
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/materials/generate.py", line 665, in replace_missing_texture
    image.filepath = image_path
TypeError: bpy_struct: item.attr = val: Image.filepath expected a string type, not PosixPath
```

TheDuckCow (changes requested):
Largely looking good! A few comments but we're quite close, sorry it took so long for me to review.

I tested this out using the recent jmc2obj release, and seems to load in the props nicely on this parent empty.

Request change sis mostly due to unit tests failing which are pointing out some problems, and the parent empty needing to be added to the same collection as the rest of the world imported (and that related discussion point on whether to do that for old style worlds as well).

MCprep_addon/materials/default_materials.py
StandingPadAnimations: This was removed due to it never being fully finished, and was never usable in production

TheDuckCow: Ok cool, just wanted to make sure 👍

MCprep_addon/materials/material_manager.py
```diff
@@ -431,6 +431,8 @@ def execute(self, context):
 
 		self.report({'INFO'}, f"Updated {count} materials")
 		self.track_param = context.scene.render.engine
+
+		# NOTE: This is temporary
```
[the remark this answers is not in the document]
StandingPadAnimations: Here we still base the exporter on what's set in MCprep (Mineways or jmc2OBJ), not the exporter property on an OBJ

MCprep_addon/world_tools.py
StandingPadAnimations: I was considering it, however it wouldn't match what we have currently for imports (which are legacy)

TheDuckCow: I would argue consistently across the same version of MCprep is more important that partial consistency with prior versions, we can call it out in the changelog that we changed structure. It would just cause confusion why it's sometimes one way vs another.

TheDuckCow:
I'm suggesting that we merge/rebase dev into/under this branch once https://github.com/Moo-Ack-Productions/MCprep/pull/587 and https://github.com/Moo-Ack-Productions/MCprep/pull/586 have been merged. 

Then, as mentioned, I'd still like to more consistently put everything as a child of an empty. Other benefit is it makes it easier to select one thing and reliably transform the entire world (yes, not that much harder to "select all objects in collection", but I'd say it's still a pro to give another option).

TheDuckCow:
I'm willing to help out to do any test fixing since you're still having trouble getting them to run locally. Once we have this branch in, I'd say we are well overdue to calling 3.6 complete.

TheDuckCow:
Fixed several of the issues, but some failures remaining for my future reference below. Also need to move the added empty parents into the same new world collection, and consider whether we should toggle the selectable/global visibility options (right now both are off; thinking we could leave selection on but let it be scene non visible, just in the hierarchy).

Several of them are indicating the wrong `MCprep_exporter_type` value assigned, so wonder if there's an actual issue there.

```
FAIL: test_swap_texture_pack (materials_test.MaterialsTest)
End to end test of swap texture pack.
  File "/Users/patrickcrawford/Documents/gits/mcprep/test_files/materials_test.py", line 892, in test_swap_texture_pack
    with self.assertRaises(RuntimeError):
AssertionError: RuntimeError not raised

FAIL: test_meshswap_world_mineways_combined (spawner_test.MeshswapTest)
  File "/Users/patrickcrawford/Documents/gits/mcprep/test_files/spawner_test.py", line 476, in test_meshswap_world_mineways_combined
    self.assertEqual(self.addon_prefs.MCprep_exporter_type, "Mineways")
AssertionError: 'jmc2obj' != 'Mineways'
- jmc2obj
+ Mineways


FAIL: test_meshswap_world_mineways_separated (spawner_test.MeshswapTest)
  File "/Users/patrickcrawford/Documents/gits/mcprep/test_files/spawner_test.py", line 439, in test_meshswap_world_mineways_separated
    self.assertEqual(self.addon_prefs.MCprep_exporter_type, "Mineways")
AssertionError: 'jmc2obj' != 'Mineways'
- jmc2obj
+ Mineways


FAIL: test_convert_mtl_simple (world_tools_test.WorldToolsTest)
Ensures that conversion of the mtl with other color space works.
  File "/Users/patrickcrawford/Documents/gits/mcprep/test_files/world_tools_test.py", line 324, in test_convert_mtl_simple
    self.assertIsNotNone(
AssertionError: unexpectedly None : Failed to mock color space and thus could not test convert_mtl

FAIL: test_world_import_mineways_combined (world_tools_test.WorldToolsTest)
  File "/Users/patrickcrawford/Documents/gits/mcprep/test_files/world_tools_test.py", line 219, in test_world_import_mineways_combined
    self.assertEqual(self.addon_prefs.MCprep_exporter_type, "Mineways")
AssertionError: 'jmc2obj' != 'Mineways'
- jmc2obj
+ Mineways


FAIL: test_world_import_mineways_separated (world_tools_test.WorldToolsTest)
  File "/Users/patrickcrawford/Documents/gits/mcprep/test_files/world_tools_test.py", line 197, in test_world_import_mineways_separated
    self.assertEqual(self.addon_prefs.MCprep_exporter_type, "Mineways")
AssertionError: 'jmc2obj' != 'Mineways'
- jmc2obj
+ Mineways


Ran 67 tests in 14.273s

FAILED (failures=6, skipped=2)
```

MCprep_addon/world_tools.py
TheDuckCow: Resolved this with https://github.com/Moo-Ack-Productions/MCprep/pull/555/commits/0d65ffd0d7778d220a2b359a96f7b94f6962397c

StandingPadAnimations:
@TheDuckCow should we go ahead and create a second RC for 3.6 containing the changes on this branch (before we merge it to `dev`)? Although we're still working on unit test issues, for the sake of time, I think an RC containing these changes (even if they're not merged) will help us catch some bugs while we work on getting unit tests working again.

From what I can tell (correct me if I'm wrong), the last major hurdle with unit tests is issues with the reported being incorrect in some areas due to some areas of MCprep not yet being updated to retrieve exporter information from the OBJ itself first (rather than exclusively checking `MCprep_exporter_type`). I think for that, we'd need to create some higher level functions that would A. retrieve the CommonMCOBJ data from an OBJ (which is stored on the parent empty) and B. return the exporter from that data (or if that data doesn't exist, fall back to older checks); perhaps a `get_cmcobj_data` and `get_exporter`. Not sure how much work this would be though

StandingPadAnimations:
(If the proposal to create a new RC is fine, make sure to create the `3.6.0-rc-2` tag)

TheDuckCow:
Hey @StandingPadAnimations yeah let's go for it, feel free to proceed to creating another release candidate while I try to think about solutions to the existing tests. Part of this stage is really testing it a lot anyways, so an RC is good for that.

MCprep_addon/world_tools.py
TheDuckCow: Turns out this seems to be the source of the problem, indeed the parse_header above if commented out makes _more_ of the tests pass. Which also means we really do need to add also tests for the common obj part too, otherwise we're not really ensuring this continues to work.

TheDuckCow: Through investigating this, I have also come to realize the existing commonmcobj parser actually always parses the _entire_ obj file, line by line, in search of the header. This could be a massive slowdown for large obj files.

FYI I'm looking into the best option to short circuit @StandingPadAnimations, most likely if we haven't seen the anchor `COMMON_MC_OBJ_START` within the first 100 lines of the header, then we're probably safe to say it's safe to exit early. Started this diff locally, will need to verify against actual Mineways and jmc2obj example commonmcOBJ exports when I get a chance:

```diff
    # Read in the header
-   for l in f:
-       tl = " ".join(l.rstrip().split())
-       if tl == "# COMMON_MC_OBJ_START":
+   lines_read = 0
+   for _l in f:
+       tl = " ".join(_l.rstrip().split())
+       lines_read += 1
+       if lines_read > 100 and tl and not tl.startswith("#"):
+           break  # no need to parse further than the true header area
        elif tl == "# COMMON_MC_OBJ_START":
            header.append(tl)
            found_header = True 
            continue
```

StandingPadAnimations: CommonMCOBJ allows for headers to be anywhere in the file technically, either in the beginning or the end. That said, we could check for both 100 lines in the beginning and 100 lines at the end, and allow it to be configurable for users

StandingPadAnimations:
How is this branch looking with regards to tests?

TheDuckCow:
> How is this branch looking with regards to tests?

I've been a bit tied up this weekend, but actively working on it tonight starting now. If all goes well, might have it out for a follow up review tonight.

MCprep_addon/world_tools.py
TheDuckCow: I [briefly looked](https://www.geeksforgeeks.org/python-reading-last-n-lines-of-a-file/) at efficiently reading the last N lines of a file without actually iterating over all of them. It's a little complicated. The other more practical consideration is that both jmc2obj and Mineways, as I've just verified, put commonobj towards the top. Currently in both exports I've done, the header starts ~line 55, and for jmc2obj it's at the very start. So, Seems like a buffer of 100 should be safe. I might time-test the difference later to see how much it would end up saving

StandingPadAnimations: If we do implement a 100 line limit, then we should also allow the user to be able to configure it somewhere. That way, if an OBJ that is a CommonMCOBJ export doesn't put it in the first 100 lines, users wouldn't have to go down into the Python code and modify the limit.

I think a good way to test would be to have an OBJ file where the header is at the end (which should be trivial since OBJ is a text format) and benchmark the time difference in MCprep. Not only would it give us insights into whether or not it makes a practical difference, it would also test to see if any unexpected issues occur

TheDuckCow: I'd be in favor of marking that a TODO if you're ok way it, to keep parsing simple and fast for now (otherwise it'll be seen as a regression).

Also spoiler I will not finish this tonight, I had to update and create a test world for v1.21 of MC and do exports for both jmc2obj and mineways. While that is working, it perhaps has revealed some issues with the mapping script which I am looking into still along with setting up tests to make sure parsing is working against real commonmcobj  world exports.

TheDuckCow:
Not that all of these correspond to the effort of this PR, but just calling out my remaining checklist of things to fix:

- [ ] Re-export my local mineways mcobj export, so it's a similar world example. once I have that, I'll do the pushes of the save and sample export files.
- [ ] Yes get tests working per above.
- [ ] Resolve the issue of the mappings not being updated.
- [ ] Mesh swao fixes (likely just mapping json updates): Per above, make sure the default jmc2obj grass (ie tall grass) is picked up in meshswap, as right now it is not. Same goes for the *top* of tall grass, the bottom is swapped but the top part is not deleted. The top of furnaces also don't get removed during meshswap.
- [ ] Could be nice to already offset the meshswap spawner snapping mode based to the selected system in preferences, though that's super minor and pre-existing that it doesn't reflect the preferences setting.
- [ ] Fix or investigate why find missing doesn't work currently; and also right now, if we have an obj with all textures missing, the prep materials operator responds with a misleading message saying that the user needs to select something. Really we should be trying to resolve paths with reconnected textures, and failing that say textures are just missing.


But outside the above, generally all other parts of MCprep that I've touched seemed to be ok. I'm going to focus on the world export+settings changes in this PR, then might create another for the other bullets.

MCprep_addon/world_tools.py
StandingPadAnimations: Sounds good to me

MCprep_addon/commonmcobj_parser.py
```diff
@@ -214,34 +214,38 @@ def clean_and_extract(line: str) -> Tuple[str, str]:
...
+
     # Read in the header
-    for l in f:
-        tl = " ".join(l.rstrip().split())
-        if tl == "# COMMON_MC_OBJ_START":
+    lines_read = 0
+    for _l in f:
+        tl = " ".join(_l.rstrip().split())
+        lines_read += 1
+        if lines_read > 100 and tl and not tl.startswith("#"):
```
StandingPadAnimations, line 238: I'll see if I can make this limit configurable in user preferences

TheDuckCow: I'd prefer we keep things simple personally, adding preferences for everything makes the preferences section long and hard to read if every minute thing becomes an option (even if it's pushed under advanced). I'm not particularly worried about running into this problem.

If anything, let's double it to 200, how does that sound? For smaller obj we're still typically 1000's of lines of the overall file so it's still a big parsing improvement over reading the while thing (twice)

TheDuckCow:
With this, the number of failing tests (even with the new ones added) is down to just 3 which matches my local run:

```
(4.1.1)   	all_tests	68	2	3	test_swap_texture_pack;test_convert_mtl_simple;test_world_import_cmcobj_jmc_full [test_mappings]
tests took 23s to run, ending with code 1
```

I think these errors are actually representing problems that need to be fixed, which is a good thing. The jcm2obj mappings appear to not be connecting anymore, which would impact texture swap and reconnecting textures.

MCprep_addon/commonmcobj_parser.py
StandingPadAnimations: 200 sounds like a good compromise. If an OBJ doesn't have the header within the first 200 lines, we can resolve it at that time (if it happens)

StandingPadAnimations:
I'll run these tests locally some time this week to confirm the results on my end (just want to check, are the tests failing in both GitHub Actions and locally on your end the same?). Now that we have GitHub Actions, we have a good reference for how tests are ran

MCprep_addon/conf.py
```diff
@@ -133,7 +133,7 @@ def __init__(self):
 		# list of material names, each is a string. None by default to indicate
 		# that no reading has occurred. If lib not found, will update to [].
 		# If ever changing the resource pack, should also reset to None.
-		self.material_sync_cache = []
+		self.material_sync_cache: Optional[List] = []
```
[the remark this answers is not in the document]
StandingPadAnimations: Resolved in f678b9f92ce73e1edb18e35233425a1509d82d58

MCprep_addon/util.py
StandingPadAnimations: Resolved in 1fec0da624d75f615590469ef1ca34b30ad432c5

TheDuckCow:
> I'll run these tests locally some time this week to confirm the results on my end (just want to check, are the tests failing in both GitHub Actions and locally on your end the same?). Now that we have GitHub Actions, we have a good reference for how tests are ran

Yes the results are the same locally and on github actions. As I mentioned, I think there's actual bugs there that need to be troubleshooted further. I need to switch to other things tonight, generally would be good to avoid dual working on the same branch in case we clash, but feel free to take over for the next say 18h if you're up for it.

TheDuckCow:
Oh and one minor thing, multiple commits for each individual action are great, but since tests run on *every* push now, that gets expensive (in terms of using the free quota) if the PR is already open. So feel free to make several commits locally and push several at once when a PR is already open (except in this case where we're collaborating on the same PR, I guess this time it matters more to keep it up to date)

StandingPadAnimations:
> > I'll run these tests locally some time this week to confirm the results on my end (just want to check, are the tests failing in both GitHub Actions and locally on your end the same?). Now that we have GitHub Actions, we have a good reference for how tests are ran
> 
> Yes the results are the same locally and on github actions. As I mentioned, I think there's actual bugs there that need to be troubleshooted further. I need to switch to other things tonight, generally would be good to avoid dual working on the same branch in case we clash, but feel free to take over for the next say 18h if you're up for it.

Sounds good. I'll be mostly working on some style things mentioned in earlier reviews (just pushed a couple actually); I'll see if I can get some of these bugs resolved

StandingPadAnimations:
> Oh and one minor thing, multiple commits for each individual action are great, but since tests run on _every_ push now, that gets expensive (in terms of using the free quota) if the PR is already open. So feel free to make several commits locally and push several at once when a PR is already open (except in this case where we're collaborating on the same PR, I guess this time it matters more to keep it up to date)

Alright, will keep that in mind

StandingPadAnimations:
Ran the tests locally on my machine (after updating this branch with `dev`, I'll push that merge after I finish some other things):, got the same result as you did (as well as GitHub Actions):
```
bversion        ran_tests       ran     skips   failed  errors
(4.1.1)         all_tests       68      2       3       test_swap_texture_pack;test_convert_mtl_simple;test_world_import_cmcobj_jmc_full [test_mappings]
tests took 41s to run, ending with code 1
```

This confirms that now, tests are reproducible (at least on macOS, GitHub Actions with the Ubuntu runner, and Arch Linux on bare metal)

StandingPadAnimations:
One of the errors I'm getting locally is the following:
```
Error: File not found, could not import obj '/home/mahid/Documents/MCprep/test_files/fake_world.obj'
```
Are we missing some files in `git-lfs`?

StandingPadAnimations:
Alright resolved `test_convert_mtl_simple`; this was due to `convert_mtl` changing the return values, where `None` now represents a success and `MCprepError` represents a failure, with booleans being removed. 

With some fixes to the test, my results look as such: 
```
bversion        ran_tests       ran     skips   failed  errors
(4.1.1)         all_tests       68      2       2       test_swap_texture_pack;test_world_import_cmcobj_jmc_full [test_mappings]
tests took 38s to run, ending with code 1
```

MCprep_addon/materials/prep.py
```diff
@@ -419,8 +423,7 @@ class MCPREP_OT_swap_texture_pack(
 
 	@classmethod
 	def poll(cls, context):
-		addon_prefs = util.get_user_preferences(context)
-		if addon_prefs.MCprep_exporter_type != "(choose)":
+		if world_tools.get_exporter(context) != world_tools.WorldExporter.Unknown:
```
TheDuckCow, line 426: Ok so this is the reason one of the tests fail actually, and points to a behavior change.

In the currently live MCprep, someone who imports a world via the built in importer, they can still do texture swap if they manually adjust the preferences importer from (Choose) to either mineways or jmc2obj.

But in this branch, it's grayed out. 

Another use case I do sometimes: Use generate materials to create a few materials, and then later I might choose to change the texture pack for just those selected ones, even though they didn't come from an actual OBJ import. 

I know we want to ideally move away from having the global `MCprep_exporter_type` but I also want to avoid the reduction of functionality for users because we'll get a lot of people complaining that they can't use texture swap any more due to being grayed out.

Since we're still going to have the mapping json (since we need to for things like reconnect textures), I'm going to rework this so it's more of a fallback system rather than strictly the new way, since we're not removing `MCprep_exporter_type` in this release. In the future, we'll think more about how the fallback works. We could probably determine it on the fly by just checking any and all mappings in our json file for instance, basically poll to see which one is most common and then use that one (and thus, we would never really gray it out as it would always fallback to best guess for texture swap).

TheDuckCow: Ok follow up question @StandingPadAnimations as I'm trying to do this fallback, is there a reason for this typing?

`def get_exporter(context: Context) -> Optional[WorldExporter]:`

We have some branches which return `None`, and some which return `WorldExporter.Unknown`, is there any reason to not convert the `None` response branches to `WorldExporter.Unknown`? So that anywhere we use this function, we don't have to do two checks for the kind of output (is not None and ≠ WorldExporter.Unknown)

TheDuckCow: My local WIP which seems to be doing ok in my mind, let me know what you think. This setup makes it so it prefers the selected world if there's any data for it, falls back to scene level, and then if still nothing then if raises a runtime error (which is the test that was failing, since it was asserting we raise an error in the right situations).

```diff
diff --git a/MCprep_addon/materials/prep.py b/MCprep_addon/materials/prep.py
index 260942d..8bfc32a 100644
--- a/MCprep_addon/materials/prep.py
+++ b/MCprep_addon/materials/prep.py
@@ -425,6 +425,9 @@ class MCPREP_OT_swap_texture_pack(
 	def poll(cls, context):
 		if world_tools.get_exporter(context) != world_tools.WorldExporter.Unknown:
 			return util.is_atlas_export(context)
+		addon_prefs = util.get_user_preferences(context)
+		if addon_prefs.MCprep_exporter_type != "(choose)":
+			return util.is_atlas_export(context)
 		return False
 
 	def draw(self, context):
diff --git a/MCprep_addon/world_tools.py b/MCprep_addon/world_tools.py
index fe458f8..7831117 100644
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -158,7 +158,7 @@ EXPORTER_MAPPING = {
 }
 
 
-def get_exporter(context: Context) -> Optional[WorldExporter]:
+def get_exporter(context: Context) -> WorldExporter:
 	"""
 	Return the exporter on the active object if it has
 	an exporter attribute.
@@ -176,7 +176,7 @@ def get_exporter(context: Context) -> Optional[WorldExporter]:
 	"""
 	obj = context.active_object
 	if not obj:
-		return None
+		return WorldExporter.Unknown
 
 	if "COMMONMCOBJ_HEADER" in obj:
 		if obj["PARENTED_EMPTY"] is not None and obj["PARENTED_EMPTY"]["exporter"] in EXPORTER_MAPPING:
@@ -195,7 +195,7 @@ def get_exporter(context: Context) -> Optional[WorldExporter]:
 		return WorldExporter.ClassicMW
 	elif prefs.MCprep_exporter_type == "jmc2obj":
 		return WorldExporter.ClassicJmc
-	return None
+	return WorldExporter.Unknown
```

StandingPadAnimations: > Ok follow up question @StandingPadAnimations as I'm trying to do this fallback, is there a reason for this typing?
> 
> `def get_exporter(context: Context) -> Optional[WorldExporter]:`
> 
> We have some branches which return `None`, and some which return `WorldExporter.Unknown`, is there any reason to not convert the `None` response branches to `WorldExporter.Unknown`? So that anywhere we use this function, we don't have to do two checks for the kind of output (is not None and ≠ WorldExporter.Unknown)

`WorldExporter.Unknown` means that there is a world exporter defined in the CommonMCOBJ header, but it's not in our existing mapping of strings to enum values. `None` means no exporter can be detected at all

StandingPadAnimations: To be specific, `WorldExporter.Unknown` is meant to represent any exporter that, although given in the CommonMCOBJ header, we have never tested against MCprep. This makes it so that features dependent on OBJs themselves, such as meshswap are locked, but other features aren't, such as textureswap. `None` on the other hand means no exporter is defined, so the feature set of what's possible is restricted (there can be some exceptions for legacy with the  `MCPREP_OBJ_HEADER` attribute but not the `MCPREP_OBJ_EXPORTER` attribute, but that's more of a backwards compatibility thing).

Thus, I'd rather we keep the `Optional[WorldExporter]` typing, since having that nuance between `WorldExporter.Unknown` and `None` is useful.

StandingPadAnimations: Now that I think about it, looking at your diff, I think I made an error with porting textureswap. I'll fix that and push my local changes

StandingPadAnimations:
Got `test_swap_texture_pack` passing on my end
```
bversion        ran_tests       ran     skips   failed  errors
(4.1.1)         all_tests       68      2       1       test_world_import_cmcobj_jmc_full [test_mappings]
tests took 37s to run, ending with code 1
```

I made an error when adapting textureswap to the new system, where it wouldn't show if `WorldExporter.Unknown` was returned (when it should have been `None`)
```diff
diff --git a/MCprep_addon/materials/prep.py b/MCprep_addon/materials/prep.py
index 260942d..f2f9e6d 100644
--- a/MCprep_addon/materials/prep.py
+++ b/MCprep_addon/materials/prep.py
@@ -423,7 +423,7 @@ class MCPREP_OT_swap_texture_pack(
 
 	@classmethod
 	def poll(cls, context):
-		if world_tools.get_exporter(context) != world_tools.WorldExporter.Unknown:
+		if world_tools.get_exporter(context) is not None:
 			return util.is_atlas_export(context)
 		return False
```

StandingPadAnimations:
Pushed my fixes for 2 of the tests, now we should have only one failing test remaining (`test_world_import_cmcobj_jmc_full [test_mappings]`)

MCprep_addon/world_tools.py
```diff
@@ -142,22 +236,17 @@ def detect_world_exporter(filepath: Path) -> None:
...
-			return
-		obj_header.set_jmc2obj()
-		# Since this is the default for Jmc2Obj,
-		# we'll assume this is what the OBJ is using
-		obj_header.set_seperated()
+		return obj_header
 
 
-def convert_mtl(filepath):
+def convert_mtl(filepath) -> Optional[MCprepError]:
```
TheDuckCow, line 249: I'm thinking about this further, I feel like the bool signature we had before made more sense. It's useful to know if it actually did something or not. The old comment was a little wrong, it should have said "Returns: True if success, None if skipped".

Since we have the error class, maybe we can just return bool and this means whether it made a change or not?

```suggestion
def convert_mtl(filepath) -> Union[bool, MCprepError]:
```

StandingPadAnimations: That sounds good to me

StandingPadAnimations: Done, I've also pushed a slight efficiency change that makes the function exit early if the current color space is in our list of "built ins" (basically what we know we works). That way, MTLs aren't unnecessarily loaded into memory at runtime.

I've also found that we didn't have AgX in our list of built ins, so I've gone ahead and added it, alongside Khronos PBR Neutral (which is coming in 4.2)

StandingPadAnimations:
By the way, how would one go about fixing the mappings? I'm assuming we have an automated script that just needs to be modified, but I'm not sure where to start with that

MCprep_addon/world_tools.py
```diff
@@ -150,7 +150,7 @@ class WorldExporter(Enum):
 }
 
 
-def get_exporter(context: Context) -> Optional[WorldExporter]:
+def get_exporter(context: Context) -> WorldExporter:
```
StandingPadAnimations, line 153: Like I mentioned earlier, the `Optional` typing allows for more nuance.

`WorldExporter.Unknown` should mean that an exporter is detected, but not one that has extensive testing with MCprep (like a new exporter). `None` should mean no exporter has been detected at all. 

Thus, I'd rather we keep the Optional typing

MCprep_addon/materials/prep.py
```diff
@@ -423,7 +423,11 @@ class MCPREP_OT_swap_texture_pack(
 
 	@classmethod
 	def poll(cls, context):
-		if world_tools.get_exporter(context) is not None:
+		if world_tools.get_exporter(context) != world_tools.WorldExporter.Unknown:
```
StandingPadAnimations, line 426: I think this was accidentally reverted. This should check for `None`, since that means no exporter has been detected

MCprep_addon/world_tools.py
TheDuckCow: Hm I can see that argument. Though it means multiple checks against None _and_ Unknown I guess we could define the "supported" enums in a global somewhere and check if the response is in that?

MCprep_addon/materials/prep.py
TheDuckCow: It's self consistent with my changes above as I intentionally had updated it to always return Unknown in the branches that were previously None, but we're discussing above now I suppose

MCprep_addon/world_tools.py
StandingPadAnimations: We could also create a "unsupported" global with the following:
```py
UNSUPPORTED_OR_NONE = (WorldExporter.Unknown, None)
```

And check against that for checks that require blocking both. The reason I'd rather keep `Optional` though is because there are certain operations that don't necessarily require an exporter to be supported, but do need it to be detected, such as textureswap. For those operations, collapsing an unknown exporter and none being detected to a single enum would make things a little more convoluted.

TheDuckCow: Ok I like that solution. I'll make a commit for this soon

StandingPadAnimations:
I'll revert the changes related to `get_exporter`

MCprep_addon/world_tools.py
TheDuckCow: Welp we both came up with the same commit 😅 using yours since it's already on remote, resolving this thread

MCprep_addon/materials/prep.py
TheDuckCow: Resolved above

TheDuckCow:
Making progress on the mapping issue - it's because, sometime since the version of jmc2obj I used to export our test world, exports started to be prefixed with "minecraft_block-" which was throwing off my test code for checking mappings.

Thing is I do check for that already in code so could be the test wasn't comprehensive enough, I'm checking further to see if there's a common prefix-stripping function we're using in code, so that this mapping check file works equally.

Also we had an arbitrary threshold of 30 keys not mapped to canonical texture names - I'm updating that threshold to be 40, since these pretty much seem reasonable to not be mapped (in that they are kind of special textures, not typical blocks). Point was never to have 100% coverage, but just a sanity check to help make sure we have _mostly_ overlapping mappings.

`['jmc2obj_banner-banner_base0', 'jmc2obj_banner-banner_base14', 'jmc2obj_banner-banner_base3', 'jmc2obj_banner-banner_base9', 'jmc2obj_block-redstone_dust_dot', 'jmc2obj_block-redstone_dust_line0', 'jmc2obj_block-redstone_dust_line0_on', 'minecraft_entity-armorstand-wood', 'minecraft_entity-bed-red', 'minecraft_entity-chest-ender', 'minecraft_entity-chest-normal', 'minecraft_entity-chest-normal_left', 'minecraft_entity-chest-normal_right', 'minecraft_entity-chest-trapped', 'minecraft_entity-chest-trapped_left', 'minecraft_entity-chest-trapped_right', 'minecraft_entity-creeper-creeper', 'minecraft_entity-end_portal', 'minecraft_entity-player-wide-steve', 'minecraft_entity-signs-acacia', 'minecraft_entity-signs-birch', 'minecraft_entity-signs-crimson', 'minecraft_entity-signs-dark_oak', 'minecraft_entity-signs-jungle', 'minecraft_entity-signs-mangrove', 'minecraft_entity-signs-oak', 'minecraft_entity-signs-spruce', 'minecraft_entity-signs-warped', 'minecraft_entity-skeleton-skeleton', 'minecraft_entity-skeleton-wither_skeleton', 'minecraft_entity-zombie-zombie', 'minecraft_painting-kebab', 'minecraft_painting-pool', 'minecraft_painting-void']
`

The separation here could actually be useful as these are mostly special entities which we might not want to accidentally overlap with block textures, something right now we don't have great control over. But, nothing to block this PR.

TheDuckCow:
Ok I think this 'otta wrap this branch up! All tests are passing 🎉  but, there are a couple outstanding issues worth addressing before the release:

- Find missing textures is not really working well, at least for jmc2obj. I will review in another branch as I've started troubleshooting
- When all textures are missing, Prep Materials also wrongly says "no materials found to modify", which kind of doesn't make sense - even if textures are missing, it should be able to prep them (and per above, if the tick box is selected, should replace the missing textures)
- Blender 4.2 has an error when trying to use the geo nodes weather effects, see:
```
mcprep_addon dev response (requests lib) {'name': '-O19oCElybIt7FYFJ2bf'} at /1/track/usage_dev.json
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.2/scripts/addons/MCprep_addon/tracking.py", line 887, in wrapper
    res = function(self, context)
          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.2/scripts/addons/MCprep_addon/spawner/effects.py", line 1057, in execute
    add_geonode_area_effect(context, effect)
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.2/scripts/addons/MCprep_addon/spawner/effects.py", line 113, in add_geonode_area_effect
    geo_update_params(context, effect, geo_mod)
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.2/scripts/addons/MCprep_addon/spawner/effects.py", line 469, in geo_update_params
    geo_mod[geo_inp_id[inp.name]] = value
    ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
TypeError: Cannot assign a 'float' value to the existing 'Input_18' Boolean IDProperty
```

I think all of these should be handled in their own, bespoke specific branches so we don't get hung up on any one individually in case we can't easily resolve.

TheDuckCow (approved):
Approving my own work from my side lol but let me know @StandingPadAnimations if you're good with the plan to merge this now, or if there's anything outstanding (related to CommonMCOBJ specifically)

StandingPadAnimations:
Merging now, I'll create a new RC release for this

Issue #440, opened 2023-07-08
Using custom properties for assets


zNightlord:
### Check against existing requests

- [X] I have checked existing feature requests, and my idea is different

### Describe the context

"mob_skip_prep" "make_real" are defined in the 
[mcprep data update json](https://github.com/TheDuckCow/MCprep/blob/b4e1f0e55d1bebdaee92af800c0145e90f85b0a6/MCprep_addon/MCprep_resources/mcprep_data_update.json#L4111)
I find this inconvenient in some way.

So I propose to move to custom properties. 
Custom Properties can be user added in and read through the API. [Custom properties](https://docs.blender.org/manual/en/latest/files/custom_properties.html#python-access) 
We can also set custom properties after doing certain operation.

Another name for this is "Extend MCprep attributes" because OBJ importer use this also

### How do you imagine your feature works?

For material we introduce something like `NO_PREP` or `SKIP_PREP` custom properties with 1 value with that we can bring that information in so it doesn't clear the existing nodes in certain rig material.
Same for collection in meshswap blend file for make real but the operator parameter can override it.

### What existing workaround (or closest thing to a workaround) do you have today (within Blender, MCprep, or any software)? If there is no workaround, explain why you feel this way.

Beside custom properties there is asset tag but it is in 3.0, we can't really use that.

StandingPadAnimations:
Marked this for `3.5.1` (albeit low priority)

zNightlord:
Here are some properties I can bring up

Material:
- MCPREP_NOPREP (bool): Bypass the prep material
- MCPREP_MOBTYPE (string): Name of vanilla mob material, useful for mob skin like villager in skinswap handle profession, level, type texture
- MCPREP_HASCOLOR (bool): Material possibly has another color variation for skin swap

Object or Collection:
- MCPREP_MOBTYPE (string): Like above, maybe can have "CUSTOM" if not a mob rig
- MCPREP_RIGVERSION (string): a string (blender) versioning of a rig, I was thinking about not import the rig through `bAppendLink()` load use case
- MCPREP_MAKEREAL (bool): For realize the rig

- MCPREP_MAKESWAP (bool): This already exist, meshswap noimport suffix in meshswap.blend
- MCPREP_NOSWAP (bool): This already exist added when spawn using meshswap spawner 

Is the naming convention good? Or something need to be added?

PR #561, opened 2024-04-15
Adding material no_prep properties

zNightlord:
This is a part of #440 splitted into many PRs. 
This first introduce `MCPREP_NO_PREP` properties under material, this is useful for material doesn't want MCprep prep material overwrites it like on some rigs have been defined in the [json](https://github.com/Moo-Ack-Productions/MCprep/blob/c7622a626e7495db5919d63149f0cfc53fe62b14/MCprep_addon/MCprep_resources/mcprep_data_update.json#L4384) file `"mob_skip_prep"`.  

Additional question:
- Should there an option on prep material to force prepping this skipped no prep?

image

### DIFF
--- a/MCprep_addon/materials/prep.py
+++ b/MCprep_addon/materials/prep.py
@@ -205,6 +205,7 @@ class MCPREP_OT_prep_materials(bpy.types.Operator, McprepMaterialProps):
 		engine = context.scene.render.engine
 		count = 0
 		count_lib_skipped = 0
+		count_no_prep = 0
 
 		for mat in mat_list:
 			if not mat:
@@ -212,6 +213,10 @@ class MCPREP_OT_prep_materials(bpy.types.Operator, McprepMaterialProps):
 					"During prep, found null material:" + str(mat), vv_only=True)
 				continue
 
+			elif mat.get("MCPREP_NO_PREP", False):
+				count_no_prep += 1
+				continue
+			
 			elif mat.library:
 				count_lib_skipped += 1
 				continue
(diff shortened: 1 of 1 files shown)


MCprep_addon/materials/prep.py
```diff
@@ -291,14 +295,21 @@ def execute(self, context):
...
+		has_no_prep = count_no_prep > 0
+		has_lib_skipped = count_lib_skipped > 0
+		has_mat = count > 0
+		
+		_info = {}
+		_info[f"modified {count}"] = has_mat
+		_info[f"skipped {count_lib_skipped} linked"] = has_lib_skipped
+		_info[f"founded {count_no_prep} no prep"] = has_no_prep
+		
+		mat_info = ", ".join(k for k, v in _info.items() if v).capitalize()
```
StandingPadAnimations, line 307: I'm not the biggest fan of this, as it's not really clear what this does. I would make this separate if statements and just push elements into some `info` list. Dictionaries don't really make sense here. In addition, `_info` doesn't make sense here, as it's not a private member of a class.

```suggestion
		has_no_prep = count_no_prep > 0
		has_lib_skipped = count_lib_skipped > 0
		has_mat = count > 0
		
		info = []
		if has_mat:
			info.append("modified {count}")
		if has_lib_skip:
			info.append("skipped {count_lib_skipped}")
		if has_no_prep:
			info.append("{count_no_prep}" not prepped")		
		mat_info = ", ".join(x for x in str).capitalize()
```

As an aside, perhaps we should lump `count_lib_skipped` and `count_no_prep` into one count, since they're both skipping materials.

StandingPadAnimations (changes requested):

StandingPadAnimations:
I don't think an option for force prep is needed. Assuming it's a user override, that should be respected always

zNightlord:
> I don't think an option for force prep is needed. Assuming it's a user override, that should be respected always

I'm just thinking about the case where you decide need prep material but have to go though each of them set it to 0 so.

StandingPadAnimations:
Overall I think we could put this under #274 (which I've been meaning to discuss more on for MCprep 4.0) and take a look at it later once we know how everything in a revamped system would fall in place

MCprep_addon/materials/prep.py
TheDuckCow: I'm fine with grouping these together. Only other consideration is to try and make info reports rather short, since they fill out the bottom bar in the UI and are fleeting and non-blocking.

I would personally prefer keeping the short message, and group them all into one category for the purpose of the info report: f"Modified {count} materials, {skip_count} skipped" and leave it at that.

I am, however, totally fine with printing out more of this to the console. I also support @StandingPadAnimations simplified string cat (just be sure to put `f` ahead of some of those strings using vars)

TheDuckCow (changes requested):
Very minor change requested, but once I see it I'd be happy to approve this as is. I don't see a reason to wait for something like this, and I actually like it better than the json approach we've been using so far (centrally defining rig names for checkups is kind of gross). I don't think this will interfere with the new material system planned, they can both coexist in my mind.

StandingPadAnimations (approved):

PR #564, opened 2024-04-28
Updated test run script and refreshed data file for MC 1.20.5

TheDuckCow:
Like the description says, a mostly automated update but also fixing the test script so it can run again with the latest blender addon builder v0.4.0 release.

Current test results (just did 1x blender version), all in the clear.

```
(3.6.2)   	all_tests	65	2	0	No errors
tests took 16s to run, ending with code 0
```


TheDuckCow:
FYI @StandingPadAnimations in case you run into the test script not working.

Would love to get to a github action auto runner at some point, even if it's just for one version of blender.

MCprep_addon/MCprep_resources/mcprep_data_update.json
```diff
@@ -869,7 +869,7 @@
 			"absorbing_hardcore_half": "gui/sprites/hud/heart/absorbing_hardcore_half",
 			"absorbing_hardcore_half_blinking": "gui/sprites/hud/heart/absorbing_hardcore_half_blinking",
 			"absorption": "mob_effect/absorption",
-			"acacia": "entity/chest_boat/acacia",
+			"acacia": "entity/signs/acacia",
```
StandingPadAnimations, line 872: Seems strange that `chest_boat` was changed to `signs`. I think we should investigate this further.

MCprep_addon/MCprep_resources/mcprep_data_update.json
```diff
@@ -1002,7 +1004,7 @@
 			"big_smoke_7": "particle/big_smoke_7",
 			"big_smoke_8": "particle/big_smoke_8",
 			"big_smoke_9": "particle/big_smoke_9",
-			"birch": "entity/chest_boat/birch",
+			"birch": "entity/signs/birch",
```
StandingPadAnimations, line 1007: Same thing here. Perhaps some paths were changed, but I'd rather we check and verify that this is valid

MCprep_addon/MCprep_resources/mcprep_data_update.json
```diff
@@ -1097,13 +1108,13 @@
 			"bubbles": "gui/sprites/container/brewing_stand/bubbles",
 			"budding_amethyst": "budding_amethyst",
 			"burn_pottery_pattern": "entity/decorated_pot/burn_pottery_pattern",
-			"burn_progress": "gui/sprites/container/furnace/burn_progress",
+			"burn_progress": "gui/sprites/container/blast_furnace/burn_progress",
```
StandingPadAnimations, line 1111: And this went from `furnace` to `blast_furnace`.

run_tests.py
```diff
@@ -41,7 +41,7 @@
 import time
 
 
-COMPILE_CMD = ["bpy-addon-build", "--during-build", "dev"]
+COMPILE_CMD = ["bab", "-b", "dev"]
```
StandingPadAnimations, line 44: Just curious, will a fast option be created now that we've upgraded to BpyBuild 0.4?

StandingPadAnimations (approved):
There's a couple of changed texture paths that I think we should check with a live test before we merge. I'm 98% sure it should be fine, but it doesn't hurt to check.

In addition, since we're now merging `milestone-3-6-0` into dev (#565), I think this would also be a good time to make an action for fast builds. Approving regardless though

MCprep_addon/MCprep_resources/mcprep_data_update.json
TheDuckCow: I also agree... but these are completely auto generated. So likely more to do with alphabetical orderings changing or folder names within the jar files of Minecraft itself. If anything, we should expand the rulesets of things to just outright ignore or more intentionally sub-category.

run_tests.py
TheDuckCow: Yeah that's my hope, I didn't get to implementing it fully though. I had it building fast by skipping the resource files, but then we still need to have at least the old copy of resource files retained in the addon folders (whereas right now, the builder always completely clears the initial install). So, some different workarounds like moving the pre-existing files to a temp place and then back into the b3d specific addons folder after is what we need to do. That will make testing so much faster, I've been missing it :)

TheDuckCow:
Going to go ahead and merge this in the meantime, the fast option will be good but can be its own task for sure.

PR #567, opened 2024-04-29
docs: Added section on signing off commits

StandingPadAnimations:
This adds a section on signing off commits (noted to be optional, but useful), which is mostly a copy-paste of [BpyBuild's section](https://github.com/Moo-Ack-Productions/bpy-build/blob/loose-ends/CONTRIBUTING.md#signing-off-commits)

To summarize: signing off commits is a way for the commit author to say, “Not only am I submitting this commit, the change made in this commit is allowed to be used”. This verifies:
- The person submitting the commit indeed made the change (what happens most of the time)
- The commit is based on existing code that is allowed to be used (ex. using code from another Blender addon, as all Blender addons are GPL licensed)
- The change in the commit comes from someone who has given permission for the change to be submitted (although they should also sign off of the commit). This includes coauthors.

As this is optional (although encouraged), this won't impact existing contributions to MCprep.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -219,11 +219,11 @@ Signing off of all commits, although not required, is good practice to certify t
 > Developer's Certificate of Origin 1.1
 > By making a contribution to this project, I certify that:
 >
-> a. The contribution was created in whole or in part by me and I have the right to submit it under the open source license indicated in the file; or
-> b. The contribution is based upon previous work that, to the best of my knowledge, is covered under an appropriate open source license and I have the right under that license to submit that work with modifications, whether created in whole or in part by me, under the same open source license (unless I am permitted to submit under a different license), as indicated in the file; or
-> c. The contribution was provided directly to me by some other person who certified (a), (b) or (c) and I have not modified it.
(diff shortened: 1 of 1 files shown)


TheDuckCow:
Getting to this now, for my understanding: is the difference here just a matter of adding additional text into one's commit? Or is this tied to some certificate setup (in which case, should we be linking to how to set that up)?

StandingPadAnimations:
> Getting to this now, for my understanding: is the difference here just a matter of adding additional text into one's commit? Or is this tied to some certificate setup (in which case, should we be linking to how to set that up)?

Yep, just adding additional text to the commit. `git-commit` also has the flag `--signoff`, which adds this automatically based on one's `.gitconfig`. Here's an example of how it would look in `git-log`
image

TheDuckCow:
Got it, I think if we could include a link to some docs showing how to use the `--signoff` flag would be quite helpful for anyone trying to quickly understand this, myself included :)

TheDuckCow (approved):
Approving, presuming we'll add the context link showing how to use the signoff git flag

StandingPadAnimations:
Done, I'll merge this in now

PR #568, opened 2024-05-01
i18n: Fix UTF-8 Encoding Issues

StandingPadAnimations:
Python really loves to make UTF-8 encoding a pain... *sigh*

This fixes the translate action to set proper encoding for POT files and `translations.py`. Will need to test on Windows since that seems to be the source of headaches


TheDuckCow (approved):
Looks good and reasonable, thanks!

PR #571, opened 2024-05-09
Milestone 3-6-0 remerge

StandingPadAnimations:
This reverts and remerges `milestone-3-6-0` into `dev` so we can keep the commit history. That way, PRs such as #555 don't look as massive


TheDuckCow (approved):
I appreciate the effort here, thanks

PR #572, opened 2024-05-10
WIP - Other rig properties, mob type, color variant

### LINKED ISSUE #440: Using custom properties for assets

zNightlord:
WIP of #440, second part of #561 
*(This is not ready for review, discuss further on how this should be implemented)*

 As in #440 this add a system for rig have `MCPREP_RIGTYPE` non "Custom" under the rig armature object to use a different skin swap tab.
 It would be an improvement [spawn_with_skin](https://github.com/Moo-Ack-Productions/MCprep/blob/c7622a626e7495db5919d63149f0cfc53fe62b14/MCprep_addon/materials/skin.py#L657C57-L658C2) since it getting directly from the resource pack path.

https://github.com/Moo-Ack-Productions/MCprep/assets/61040487/5fc96af7-5e12-4648-94da-d9de2de3611d

The UI properties is store under a class and go through store at WindowManager context.
Question:
How should it have access to the rig materials?
1) Have a  for loop iterate though the objects find matching material type property `MCPREP_MatType` (for example "Profession" "Level" "Biome" for villager)
2) Have many custom properties `MCPREP_MatVillagerProfession` `MCPREP_MatVillagerLevel` `MCPrep_MatVillagerBiome` string *name* under the armature object and `materials.get( name)` to get

and use skin swap function to set the texture inside. 

"Invert" mob type.
How or should have an "invert" or zombified type as a checkbox? 
Allay -> Vex; Piglin -> Zombified Piglin; Hoglin -> Zombified Hoglin; Zombie Villager; 

UI it would be something like Mine imator with the variant but rather upon adding the mob you do it after with an operator button to switch variant or color `MCPREP_HAS_COLOR`.

image

### DIFF
--- a/MCprep_addon/materials/skin.py
+++ b/MCprep_addon/materials/skin.py
@@ -26,14 +26,22 @@ import urllib.request
 import bpy
 from bpy_extras.io_utils import ImportHelper
 from bpy.app.handlers import persistent
-from bpy.types import Context, Image, Material
+from bpy.types import Context, Material
 
 from . import generate
 from .. import tracking
 from .. import util
-from .spawner import spawn_util
+from ..spawner import spawn_util
 
 from ..conf import env
+from .prep import McprepMaterialProps
+
+
+swap_all_imgs_desc = (
+	"Swap textures in all image nodes that exist on the selected \n"
+	"material; if off, will instead seek to only replace the images of \n"
+	"nodes (not image blocks) named MCPREP_SKIN_SWAP"
+)
 
 # -----------------------------------------------------------------------------
 # Support functions
@@ -85,7 +93,7 @@ def handler_skins_enablehack(scene):
 	"""Scene update to auto load skins on load after new file."""
 	try:
 		bpy.app.handlers.scene_update_pre.remove(handler_skins_enablehack)
-	except:
+	except Exception:
 		pass
 	env.log("Triggering Handler_skins_load from first enable", vv_only=True)
 	handler_skins_load(scene)
@@ -96,13 +104,20 @@ def handler_skins_load(scene):
 	try:
 		env.log("Reloading skins", vv_only=True)
 		reloadSkinList(bpy.context)
-	except:
+	except Exception as e:
+		print(e)
 		env.log("Didn't run skin reloading callback", vv_only=True)
 
 
-def loadSkinFile(self, context: Context, filepath: Path, new_material: bool=False):
+def loadSkinFile(
+	self,
+	context: Context,
+	filepath: Path,
+	new_material: bool = False,
+	swap_all_imgs: bool = False) -> int:
+	"""Replaces image textures with target path for use in operator."""
 	if not os.path.isfile(filepath):
-		self.report({'ERROR'}, "Image file not found")
+		self.report({'ERROR'}, f"Image file not found: {filepath}")
 		return 1
 		# special message for library linking?
 
@@ -122,21 +137,20 @@ def loadSkinFile(self, context: Context, filepath: Path, new_material: bool=Fals
 		self.report(
 			{'WARNING'}, "Skinswap skipped {} linked objects".format(skipped))
 
-	status = generate.assert_textures_on_materials(image, mats)
+	status = generate.assert_textures_on_materials(
+		image, mats, swap_all_imgs=swap_all_imgs)
 	if status is False:
 		self.report({'ERROR'}, "No image textures found to update")
 		return 1
 	else:
 		pass
 
-	if not util.bv28():
-		setUVimage(context.selected_objects, image)
-
 	# TODO: adjust the UVs if appropriate, and fix eyes
 	if image.size[0] != 0 and image.size[1] / image.size[0] != 1:
 		self.report({'INFO'}, "Skin swapper works best on 1.8 skins")
 		return 0
 	return 0
+
 
 def loadVariantFile(self, context: Context, filepath: Path, new_material: bool=False):
 	if not os.path.isfile(filepath):
@@ -317,20 +331,6 @@ def getMatsFromSelected(selected: List[bpy.types.Object], new_material: bool=Fal
 	return mat_ret, linked_objs
 
 
-def setUVimage(objs: List[bpy.types.Object], image: Image) -> None:
-	"""Set image for each face for viewport displaying (2.7 only)"""
-	for obj in objs:
-		if obj.type != "MESH":
-			continue
-		if not hasattr(obj.data, "uv_textures"):
-			env.log("Called setUVimage on object with no uv_textures, 2.8?")
-			return
-		if obj.data.uv_textures.active is None:
-			continue
-		for uv_face in obj.data.uv_textures.active.data:
-			uv_face.image = image
-
-
 def download_user(self, context: Context, username: str) -> Optional[Path]:
 	"""Download user skin from online.
 
@@ -372,7 +372,29 @@ def download_user(self, context: Context, username: str) -> Optional[Path]:
 		self.track_param = "username"
 	return saveloc
 
+def check_entity_texture(texture_path: str) -> Optional[Path]:
+	context = bpy.context
+	addon_prefs = util.get_user_preferences(context)
+	active_pack = bpy.path.abspath(context.scene.mcprep_texturepack_path)
+	active_pack = os.path.join(
+		active_pack, "assets", "minecraft", "textures", "entity")
 
+	base_pack = bpy.path.abspath(addon_prefs.custom_texturepack_path)
+	base_pack = os.path.join(
+		base_pack, "assets", "minecraft", "textures", "entity")
+
+	# if not os.path.isdir(active_pack):
+	# 	env.log(f"No models found for active path {active_pack}")
+	# 	return None
+	base_has_textures = os.path.isdir(base_pack)
+
+
+	if base_has_textures:
+		return generate.find_from_texturepack(texture_path)
+	else:
+		env.log(f"Base resource pack has no entity texture folder: {base_pack}")
+		return None
+	
 # -----------------------------------------------------------------------------
 # Operators / UI classes
 # -----------------------------------------------------------------------------
@@ -415,18 +437,22 @@ class MCPREP_OT_swap_skin_from_file(bpy.types.Operator, ImportHelper):
 		name="New Material",
 		description="Create a new material instead of overwriting existing one",
 		default=True)
+	swap_all_imgs: bpy.props.BoolProperty(
+		name="Swap All Images",
+		description=swap_all_imgs_desc,
+		default=True)
 	skipUsage: bpy.props.BoolProperty(default=False, options={'HIDDEN'})
 
 	track_function = "skin"
 	track_param = "file import"
 	@tracking.report_error
 	def execute(self, context):
-		res = loadSkinFile(self, context, self.filepath, self.new_material)
+		res = loadSkinFile(
+			self, context, self.filepath, self.new_material, self.swap_all_imgs)
 		if res != 0:
 			return {'CANCELLED'}
 
 		return {'FINISHED'}
-
 
 class MCPREP_OT_apply_skin(bpy.types.Operator):
 	"""Apply the active UIlist skin to select characters"""
@@ -443,6 +469,10 @@ class MCPREP_OT_apply_skin(bpy.types.Operator):
 		name="New Material",
 		description="Create a new material instead of overwriting existing one",
 		default=True)
+	swap_all_imgs: bpy.props.BoolProperty(
+		name="Swap All Images",
+		description=swap_all_imgs_desc,
+		default=True)
 	skipUsage: bpy.props.BoolProperty(
 		default=False,
 		options={'HIDDEN'})
@@ -451,7 +481,8 @@ class MCPREP_OT_apply_skin(bpy.types.Operator):
 	track_param = "ui list"
 	@tracking.report_error
 	def execute(self, context):
-		res = loadSkinFile(self, context, self.filepath, self.new_material)
+		res = loadSkinFile(
+			self, context, self.filepath, self.new_material, self.swap_all_imgs)
 		if res != 0:
 			return {'CANCELLED'}
 
@@ -483,6 +514,10 @@ class MCPREP_OT_apply_username_skin(bpy.types.Operator):
 			"If an older skin layout (pre Minecraft 1.8) is detected, convert "
 			"to new format (with clothing layers)"),
 		default=True)
+	swap_all_imgs: bpy.props.BoolProperty(
+		name="Use Legacy Skin Swap Behavior",
+		description=swap_all_imgs_desc,
+		default=False)
 	skipUsage: bpy.props.BoolProperty(default=False, options={'HIDDEN'})
 
 	def invoke(self, context, event):
@@ -504,24 +539,28 @@ class MCPREP_OT_apply_username_skin(bpy.types.Operator):
 			self.report({"ERROR"}, "Invalid username")
 			return {'CANCELLED'}
 
+		user_ref = self.username.lower() + ".png"
+
 		skins = [str(skin[0]).lower() for skin in env.skin_list]
 		paths = [skin[1] for skin in env.skin_list]
-		if self.username.lower() not in skins or not self.skip_redownload:
+		if user_ref not in skins or not self.skip_redownload:
 			# Do the download
 			saveloc = download_user(self, context, self.username)
 			if not saveloc:
 				return {'CANCELLED'}
 
 			# Now load the skin
-			res = loadSkinFile(self, context, saveloc, self.new_material)
+			res = loadSkinFile(
+				self, context, saveloc, self.new_material, self.swap_all_imgs)
 			if res != 0:
 				return {'CANCELLED'}
 			bpy.ops.mcprep.reload_skins()
 			return {'FINISHED'}
 		else:
 			env.log("Reusing downloaded skin")
-			ind = skins.index(self.username.lower())
-			res = loadSkinFile(self, context, paths[ind][1], self.new_material)
+			ind = skins.index(user_ref)
+			res = loadSkinFile(
+				self, context, paths[ind], self.new_material, self.swap_all_imgs)
 			if res != 0:
 				return {'CANCELLED'}
 			return {'FINISHED'}
@@ -614,7 +653,7 @@ class MCPREP_OT_remove_skin(bpy.types.Operator):
 		skin_path = env.skin_list[context.scene.mcprep_skins_list_index]
 		col = self.layout.column()
 		col.scale_y = 0.7
-		col.label(text= f"Warning, will delete file {os.path.basename(skin_path[0])} from")
+		col.label(text=f"Warning, will delete file {os.path.basename(skin_path[0])} from")
 		col.label(text=os.path.dirname(skin_path[-1]))
 
 	@tracking.report_error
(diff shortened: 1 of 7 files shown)


MCprep_addon/materials/skin.py
```diff
@@ -703,6 +733,77 @@ def execute(self, context):
...
+	def execute(self, context: Context):
+		obj = context.object
+		if obj.type != 'ARMATURE':
+			self.report({'ERROR'}, "Please select an armature")
+			return {'CANCELLED'}
+		mats, skipped = getMatsFromSelected(obj, False)
+		mob_type = obj.get("MCPREP_mob_type", "Custom")
+		
+		if mob_type == "Custom":
+			self.report({'ERROR'}, "You are not allowed to use on Custom rig")
```
StandingPadAnimations, line 761: Slight nitpick on spelling and grammar.
```suggestion
			self.report({'ERROR'}, "You are not allowed to use on a custom rig")
```

MCprep_addon/materials/skin.py
```diff
@@ -703,6 +733,77 @@ def execute(self, context):
...
+				# Assign textures materials for drown
+				pass
+		elif mob_type == "Skeleton":
+			pass
+		
+		elif mob_type in ("Allay", "Vex"):
+			if mob_type == "Allay":
+				name += "/" + name
+			else:
+				name = "illager"
```
StandingPadAnimations, line 784: I think this would be a good area to use an Enum. Perhaps something like `MobType`, where each value is rqual to a string. From there, it you could use `MobType(whatever_string)` to instantiate it.

MCprep_addon/materials/skin.py
```diff
@@ -703,6 +733,77 @@ def execute(self, context):
...
+			
+		else:
+			pass
+		return {'FINISHED'}
+
+	def doVillager(self, materials: List[Material], texture_paths: List[str]):
```
StandingPadAnimations, line 790: Nitpick about Python convention, use snake_case for functions.
```suggestion
	def do_villager(self, materials: List[Material], texture_paths: List[str]):
```

MCprep_addon/materials/skin.py
```diff
@@ -335,7 +344,29 @@ def download_user(self, context: Context, username: str) -> Optional[Path]:
 		self.track_param = "username"
 	return saveloc
 
+def check_entity_texture(texture_path: str) -> Optional[Path]:
```
StandingPadAnimations, line 347: I'd add a docstring here so we can better understand this function.

MCprep_addon/materials/skin.py
```diff
@@ -335,7 +344,29 @@ def download_user(self, context: Context, username: str) -> Optional[Path]:
...
+		active_pack, "assets", "minecraft", "textures", "entity")
+
+	base_pack = bpy.path.abspath(addon_prefs.custom_texturepack_path)
+	base_pack = os.path.join(
+		base_pack, "assets", "minecraft", "textures", "entity")
+
+	# if not os.path.isdir(active_pack):
+	# 	env.log(f"No models found for active path {active_pack}")
+	# 	return None
+	base_has_textures = os.path.isdir(base_pack)
```
StandingPadAnimations, line 361: I'm not really the biggest fan of `os.path` for filepath operations, it's all string manipulation (vs `pathlib.Path` which is much nicer in abstraction). Not a blocker, but I'd advise using `pathlib` instead of `os.path` for new code since it's much better, and works with `os.path` functions anyway if needed.

MCprep_addon/materials/skin.py
```diff
@@ -703,6 +733,77 @@ def execute(self, context):
...
+	def doVillager(self, materials: List[Material], texture_paths: List[str]):
+		""" materials texture path order follows by profession, profession level, type """
+		prof_mat, biome_mat, level_mat = None, None, None
+		if mat[0].get("MCPREP_VILLAGER_PROFESSION"):
+				image = util.loadTexture(texture_paths[0])
+				proStat = generate.assert_textures_on_materials(image, [mat])
+			if mat.get("MCPREP_VILLAGER_BIOME"):
+				image = util.loadTexture(texture_paths[1])
+				biomeStat = generate.assert_textures_on_materials(image, [mat])
+			if mat.get("MCPREP_VILLAGER_LEVEL"):
```
StandingPadAnimations, line 799: I feel like with all of these properties, we should create a new file (like `mcprep_props.py` or `variables.py`, leaning more towards the latter) of constants assigned to these variable names. That way, we can document them better and avoid typos. 

Not a blocker for this PR, but just something I wanted to say before I forget.

MCprep_addon/materials/skin.py
```diff
@@ -703,6 +733,77 @@ def execute(self, context):
...
+				image = util.loadTexture(texture_paths[2])
+				levelStat = generate.assert_textures_on_materials(image, [mat])
+			if not all([proStat, biomeStat, levelStat]):
+				self.report({'ERROR'}, "Something wrong happen during swap variant texture")
```
StandingPadAnimations, line 803: Another nitpick on grammar, although it's pretty vague anyway. 
```suggestion
				self.report({'ERROR'}, "Something wrong happen while swapping the variant texture")
```

MCprep_addon/materials/skin.py
```diff
@@ -703,6 +733,77 @@ def execute(self, context):
...
+			else:
+				pass
```
StandingPadAnimations, line 805: `else-pass` isn't needed, I'd remove this (unless `pass` is just a placeholder at the moment.

MCprep_addon/materials/skin.py
```diff
@@ -31,8 +31,17 @@
...
 
 from ..conf import env
+from .prep import McprepMaterialProps
+
+
+swap_all_imgs_desc = (
+	"Swap textures in all image nodes that exist on the selected \n"
+	"material; if off, will instead seek to only replace the images of \n"
+	"nodes (not image blocks) named MCPREP_SKIN_SWAP"
+)
```
StandingPadAnimations, line 44: Don't forget to use `env._` for i18n
```suggestion
swap_all_imgs_desc = (
	env._("Swap textures in all image nodes that exist on the selected \n")
	env._("material; if off, will instead seek to only replace the images of \n")
	env._("nodes (not image blocks) named MCPREP_SKIN_SWAP")
)
```

MCprep_addon/materials/skin.py
```diff
@@ -703,6 +733,77 @@ def execute(self, context):
 		return {'FINISHED'}
 
 
+class MCPREP_OT_swap_skin_variant(bpy.types.Operator, spawn_util.VariationProp):
+	"""Apply the active UIlist skin to select characters"""
+	bl_idname = "mcprep.swap_skin"
+	bl_label = "Swap mob skin"
+	bl_description = "Swap the mobs variant"
```
StandingPadAnimations, line 740: Same comment about i18n here too
```suggestion
	bl_label = env._("Swap mob skin")
	bl_description = env._("Swap the mobs variant")
```

MCprep_addon/mcprep_ui.py
```diff
@@ -1044,6 +1053,12 @@ def draw(self, context):
 					# datapass = scn_props.mob_list[mob_ind].mcmob_type
 					tx = f"Spawn {name} with {skinname}"
 					row.operator("mcprep.spawn_with_skin", text=tx)
+			b_row.label(text="Resource pack")
```
StandingPadAnimations, line 1056: Same story about i18n
```suggestion
			b_row.label(text=env._("Resource pack"))
```

MCprep_addon/mcprep_ui.py
```diff
@@ -1969,6 +1984,25 @@ class McprepProps(bpy.types.PropertyGroup):
 	effects_list_index: bpy.props.IntProperty(default=0)
 
 
+class MCprepWindowManager(spawn_util.VariationProp, bpy.types.PropertyGroup):
+	skin_modes : bpy.props.EnumProperty(
+		name="Modes",
+		description="Skinswap modes",
+		items=(
+			('PLAYER', "Player", ""),
+			('MOB', "Mob/Entity", "")
```
StandingPadAnimations, line 1993: These also need to use i18n functions
```suggestion
		name=env._("Modes"),
		description=env._("Skinswap modes"),
		items=(
			('PLAYER', env._("Player"), ""),
			('MOB', env._("Mob/Entity"), "")
```

MCprep_addon/spawner/spawn_util.py
```diff
@@ -41,11 +45,298 @@
 COLL_ICON = 'OUTLINER_COLLECTION' if util.bv30() else 'COLLECTION_NEW'
 
 
+class VariationProp:
```
StandingPadAnimations, line 48: I won't go through this entire class, but don't forget to use `env._` for any user facing strings.

MCprep_addon/spawner/spawn_util.py
```diff
@@ -41,11 +45,298 @@
...
+			layout.prop(self, "cat_variation")
+		elif mob_type == "Dog" or mob_type == "Wolf":
+			layout.prop(self, "dog_variation")
+			
+		counter_variant = ["Villager", "Piglin", "Hoglin", "Allay", "Vex"]
+		if mob_type in counter_variant:
+			text = "Is Vex" if mob_type == "Allay" else "Is Zombified"
+			layout.prop(self, "is_zombiefied", text=text)
+		
+		has_variant = counter_variant + ["Zombie", "Skeleton", "Llama", "Axolotl", "Rabbit", "Llama", "Parrot", "Frog", "Horse", "Cat", "Ocelot"]
```
StandingPadAnimations, line 321: Like I said in an earlier comment, a `MobType` enum would be a better option IMO for all of these if statements.

MCprep_addon/spawner/spawn_util.py
```diff
@@ -41,11 +45,298 @@
...
 # -----------------------------------------------------------------------------
 # Reusable functions for spawners
 # -----------------------------------------------------------------------------
 
 
+def getmob_type(rig: bpy.types.Object):
+	""" Get mob type from rig
+	args
+		obj: Armature Object
+	"""
```
StandingPadAnimations, line 336: I'd also add some documentation on what this function returns (don't forget the return value type as well), and what values it returns in what conditions
```suggestion
def getmob_type(rig: bpy.types.Object) -> bool:
	""" Get mob type from rig
	args
		rig: Armature Object
		
	Returns:
		True if (insert condition), False otherwise
	"""
```

StandingPadAnimations (changes requested):
I know it's a draft at the moment, but I've done a basic review; looks mostly good. There are some areas where I think things could be done differently and more safe from typos and whatnot (such as using enums over raw strings). There's also some use of `os.path`, which isn't a blocker, but I'd rather `pathlib` be used since it has better abstractions with filesystems.

There are some i18n things to remember as well, although I do get that the i18n stuff is also extremely new as well. `env._` should be used for anything user-facing, except `self.report` (since that's important for error messages). 

I've noticed a class called `VariationProp` in `spawn_util.py`. Looks fine from first glance (minus the aforementioned i18n stuff), but it's quite large? I would suggest moving the things that don't depend on the class itself (enum variants) and make them independent functions that we can reuse later. I'm still on the fence on whether we should move all the variation related stuff to a new file, but I'll let @TheDuckCow comment on that.

MCprep_addon/materials/skin.py
TheDuckCow: Though I agree with the call to use env._ here, I do think we should do it only once on the final constructed multi line string block. We have no way of knowing the number of lines or breaks for other locales.

MCprep_addon/materials/skin.py
```diff
@@ -395,7 +426,6 @@ def execute(self, context):
 
 		return {'FINISHED'}
 
-
```
TheDuckCow, line 398: Style nit: Good to keep 2 spaces between top-level entities

MCprep_addon/materials/skin.py
```diff
@@ -703,6 +733,77 @@ def execute(self, context):
...
+			self.report({'ERROR'}, "You are not allowed to use on Custom rig")
+			return {'CANCELLED'}
+		
+		texture_paths = self.get_texture_paths(mob_type)
+		name = mob_type.lower().replace(" ", "_")
+		if mob_type == "Villager":
+			self.doVillager(mats, texture_paths)
+		elif mob_type == "Zombie":
+			check_entity_texture(context)
+			if self.zombie_variation in ["ZOMBIE", "HUSK"]:
```
TheDuckCow, line 770: I think you've mentioned this before and I tend to agree: better to not hard code rig-specific data in our code files, but rather (if anywhere) into our mcprep json data file, or better yet onto properties or metadata of the rigs themselves.

Though if in this case, is it just about the player skin 1.8 format aspect ratio? If so, I actually have some code somewhere in MCprep that does this check in order to convert to 1.8 formats on the live, with the username download function.

MCprep_addon/materials/skin.py
TheDuckCow: Agreed with above comment, but going further - is there some way that such a function like this could be generalized (and not specify villager in the name)?

Imagine that we somehow overhauled all our rigs, or provided different rigs for 3.0+ (for asset browser) vs 2.8/2.9. I would rather we avoid a situation where we have to change MCprep code just because rigs changed.

So, I think these specialized functions can still exist, I would just rather we try to make them generally useable and if anything rely on property assignments we check for to make them useable (I'm happy to change the rigs to match as needed, remember we want to save them in 2.80 specifically to be forwards compatible)

MCprep_addon/mcprep_ui.py
```diff
@@ -1969,6 +1984,25 @@ class McprepProps(bpy.types.PropertyGroup):
 	effects_list_index: bpy.props.IntProperty(default=0)
 
 
+class MCprepWindowManager(spawn_util.VariationProp, bpy.types.PropertyGroup):
```
TheDuckCow, line 1987: I might also need some additional context, where and how is this manager window being used? Is this mean to be just open session property settings used to drive operator functions?

If so, might warrant some more discussion on how I'm thinking things will evolve, since we're moving towards things like resource pack layering

TheDuckCow (commented):
Just leaving a few comments. Biggest point is seeing if we can shift away from hard coding rig names/str in MCprep code, and try to do so via the json file or props on the rigs themselves. 

I'm also getting a syntax error on this as-is (at least in blender 4.1) if we could take a look at that :)

```
 File "/Users/patrickcrawford/Library/Application Support/Blender/4.1/scripts/addons/MCprep_addon/materials/skin.py", line 796
    			if mat.get("MCPREP_VILLAGER_BIOME"):
    			                                    ^
IndentationError: unindent does not match any outer indentation level
```

MCprep_addon/mcprep_ui.py
zNightlord: Yes, since this is more of an UI properties that need only one, single to operate and doesn't need to save to the context scene, blend file so WindowManager would be the best place for it.

MCprep_addon/mcprep_ui.py
zNightlord: Yeah, the code was originally written months ago. I hold this for awhile and stuck there so. So noted.

zNightlord:
> How should it have access to the rig materials?
> 1) Have a  for loop iterate though the objects find matching material type property `MCPREP_MatType` (for example "Profession" "Level" "Biome" for villager)
> 2) Have many custom properties `MCPREP_MatVillagerProfession` `MCPREP_MatVillagerLevel` `MCPrep_MatVillagerBiome` string *name* under the armature object and `materials.get( name)` to get
> and use skin swap function to set the texture inside. 

> "Invert" mob type.
> How or should have an "invert" or zombified type as a checkbox? 
> Allay -> Vex; Piglin -> Zombified Piglin; Hoglin -> Zombified Hoglin; Zombie Villager; 

There are 2 questions. 

The first one it is because there is a function in MCPrep that goes over select armature to get the materials. That is why I'm not sure of using it or not while we could add a properties under armature that have material name to get it quickly.

The second one is just to reduce the usage of only 2 variant switch type mobs like Squid only have a default and Glow variant, allay and vex, slime and magma is also in that.

MCprep_addon/spawner/spawn_util.py
zNightlord: > Like I said in an earlier comment, a `MobType` enum would be a better option IMO for all of these if statements.

I could reduce those if elif down with `{x}_variantion` self hasattr() check in the first place and leave the other complex variation there. Not sure it going confusing, readable as this.

StandingPadAnimations: I'd use `hasattr`, it's much simpler in the end, and developers can read the Python docs if needed

StandingPadAnimations:
> Some code during the merging is mess up at some spot and not tested so I want to see how it should be structured from this draft.
> 
> > How should it have access to the rig materials?
> > 
> > 1. Have a  for loop iterate though the objects find matching material type property `MCPREP_MatType` (for example "Profession" "Level" "Biome" for villager)
> > 2. Have many custom properties `MCPREP_MatVillagerProfession` `MCPREP_MatVillagerLevel` `MCPrep_MatVillagerBiome` string _name_ under the armature object and `materials.get( name)` to get
> >    and use skin swap function to set the texture inside.
> 
> > "Invert" mob type.
> > How or should have an "invert" or zombified type as a checkbox?
> > Allay -> Vex; Piglin -> Zombified Piglin; Hoglin -> Zombified Hoglin; Zombie Villager;
> 
> There are 2 questions.
> 
> The first one it is because there is a function in MCPrep that goes over select armature to get the materials. That is why I'm not sure of using it or not while we could add a properties under armature that have material name to get it quickly.
> 
> The second one is just to reduce the usage of only 2 variant switch type mobs like Squid only have a default and Glow variant, allay and vex, slime and magma is also in that.

1. I think it's better to go over the materials on the selected armature. We should try to avoid adding properties to objects unless we're constantly using them (ex. needing to access something for the UI, which updates very often)
	- Assuming we're dealing with armatures, we could simply iterate through all child objects of the armature and get their materials. For the purposes of generalization, this function could be called something like `get_child_materials`
2. For mobs with just 2 variants, a single enum to represent the 2 states might be an option, and might also make things more extensible for custom mobs. However, I'm not sure if this would make things more or less readable (I guess we'd have to try to find out)

MCprep_addon/spawner/spawn_util.py
```diff
@@ -41,11 +45,298 @@
...
+			('NORMAL', "Normal", ""),
+			('SNOW', "Snow", "")
+		]
+		return items
+	
+	def squid_items(self, context: Context) -> List[Tuple[str, str, str]]:
+		items = [
+			('NORMAL', "Squid", ""),
+			('GLOW', "Glow", "")
+		]
```
StandingPadAnimations, line 224: For example, with this, perhaps we could change `NORMAL` and `GLOW` to `BASE_MOB`/`NORMAL` and `ALTERNATE_MOB`, then use some properties like `MCPREP_MOB_BASE_MOB`/`MCPREP_MOD_NORMAL` and `MCPREP_MOB_ALTERNATE_MOB` to define what these states should be called in the UI.

StandingPadAnimations:
Since this branch hasn't been worked on in two years, I'll close this PR to keep things clean. If someone is willing to continue this work, we could possibly reopen this specific PR, or open a new PR.

PR #574, opened 2024-05-10
build(i18n): build_pot now extracts MCprep version

StandingPadAnimations:
This makes `build_pot.py` extract the version string from `MCprep_addon/__init__.py` and use that when constructing the POT file. This way, we don't have to update the action manually for every version update.

The string parsing is a bit unwieldy, but hopefully the comment I've added should clarify as to what it's doing. 

On an unrelated note, during development of this, https://github.com/Moo-Ack-Productions/bpy-build/issues/14 was discovered. I think I know why this is occurring, so I'll get a BpyBuild bug fix out soon enough.

### DIFF
--- a/action-scripts/translate_scripts/build_pot.py
+++ b/action-scripts/translate_scripts/build_pot.py
@@ -1,10 +1,12 @@
 import ast
+import re
 from datetime import datetime, tzinfo, timedelta
 import time
 from pathlib import Path
+from typing import Optional
 import polib
 
-from bpy_addon_build.api import BabContext
+from bpy_addon_build.api import BabContext, BpyError
 
 class TranslateCallVisitor(ast.NodeVisitor):
     def __init__(self):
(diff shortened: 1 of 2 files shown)


action-scripts/translate_scripts/build_pot.py
```diff
@@ -67,10 +69,40 @@ def dst(self, dt):
...
+    with open(Path(ctx.current_path, "__init__.py"), 'r') as f:
+        for line in f:
+            if re.search(VERSION_REGEX, line):
+                version_str = line
+                break
+
+    if not version_str:
+        return BpyError("Can't extract version from __init__.py!")
+    
+    try:
```
TheDuckCow, line 87: Appreciate the detailed comments. Since this one part is complicated and full try/except wrapped, maybe it could be its own vcomma2vdot function?

Another idea that is less precise in a way but could also work, something like val.repalce (",", ".").replace(" ", ""). Could split string on "version: " first too. We can be pretty confident the format there won't change. Thoughts?

TheDuckCow (approved):
Thanks for working on this! 

Just on my phone reviewing, but pretty much looks good to me. I'm ok with whatever you choose to come up with, just offered alt ideas below. I think the benefit of yours is it's more brittle so we're more likely to be alerted if it fails to parse correctly, so there's something to be said about that.

action-scripts/translate_scripts/build_pot.py
StandingPadAnimations: I think the current method is fine. The only assumption it makes is that `version` is a tuple with a comma at the end, everything else should handle differences gracefully. 

The only concern I have is the regex, but I think it should be ok.

StandingPadAnimations:
Merging now

StandingPadAnimations:
This PR aged poorly and it's barely a week old
image

PR #576, opened 2024-05-13
build: Fixed issue with BpyError return values

StandingPadAnimations:
This fixes the issue noted in https://github.com/Moo-Ack-Productions/bpy-build/issues/14, which actually isn't a BpyBuild issue, but rather an issue with our `translate` action. 

For `translate`, we have a `translate.py` script (what BpyBuild actually works with) and a bunch of subscripts in `translate_scripts`, which perform individual actions. However, the `pre_build` function in `translate.py` didn't return the results of those subscripts. So when `build_pot.py` returns an error, that error wasn't propogated, so BpyBuild wouldn't know there was an error.

Closes https://github.com/Moo-Ack-Productions/bpy-build/issues/14

### DIFF
--- a/action-scripts/translate.py
+++ b/action-scripts/translate.py
@@ -1,15 +1,24 @@
+from typing import Optional
 from translate_scripts import build_pot, build_trans_dict, compile_po_to_mo
-from bpy_addon_build.api import BabContext
+from bpy_addon_build.api import BabContext, BpyError
 import sys
 
-def pre_build(ctx: BabContext) -> None:
+def pre_build(ctx: BabContext) -> Optional[BpyError]:
     sys.stdout.reconfigure(encoding='utf-8')
-    build_pot.pre_build(ctx)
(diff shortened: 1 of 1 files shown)


action-scripts/translate.py
```diff
@@ -1,15 +1,24 @@
...
+    
+    # Go through each module, and 
+    # run its main function, while checking 
+    # for a return value
+    modules = [compile_po_to_mo, build_trans_dict]
+    for mod in modules:
+        res = mod.main()
+        if res:
+            continue 
+        return res
```
TheDuckCow, line 21: From reading the code on its own, it's not very clear to me _why_ we continue. Does a return value imply an error or not? If so, maybe this is more readable while also being shorter


```suggestion
        error = mod.main()
        if error:
            return error
```

TheDuckCow (approved):
Ran it on my side, code seems to run fine. Minor suggestion on code style, but your call

PR #578, opened 2024-05-14
Ru translation

Aspirata:
I made a translation in Ru, but since I can't compile it, the translation may have some problems due to lack of context

PR #582, opened 2024-05-23
Russian translation

StandingPadAnimations:
This is the Russian translation for MCprep, submitted by @Aspirata (new PR required since #578 was based on the `mcprep-rc` branch). Peer review was confirmed by @Spacy80893 in #579


TheDuckCow (approved):
Approving - I'd definitely only call this partial support still, as some words like Skin Swapper and the names of the Spawners I think can be translated. But still, a great step forward

Tested on my system:
Screen Shot 2024-05-23 at 10 06 01 AM

PR #586, opened 2024-06-04
feat: Enable EEVEE Next support

### LINKED ISSUE #583: Materials don't generate in EEVEE Next

- [X] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [X] I have checked known issues, and my problem is different

### What happened?

When I set the rendering engine to eevee next, the materials just don't get created properly. Fun fact: if I switch to eevee legacy or cycles > spawn a block or item, and then switch the rendering engine back to eevee next, everything works fine

image

Download Bledner 4.2 > Set render engine to EEVEE > Try to spawn any block or item

MCprep 3.5.3 and MCprep 3.6 RC-1

Blender 4.2

StandingPadAnimations:
EEVEE Next has now been enabled in Blender 4.2, [with EEVEE Legacy being removed](https://projects.blender.org/blender/blender/commit/cc0d12dd20a7ff93a962c7ebc523d85c3506b26b). In order to make sure addons don't silently error out, EEVEE Next uses a different identifier (`BLENDER_EEVEE_NEXT`), so this commit adds support for MCprep by adding the EEVEE Next identifier to the list of supported engines.

RIP EEVEE Legacy, you have served the community well

Closes #583

### DIFF
--- a/MCprep_addon/conf.py
+++ b/MCprep_addon/conf.py
@@ -44,9 +44,11 @@ class Form(enum.Enum):
 class Engine(enum.Enum):
 	"""String exact match to output from blender itself for branching."""
 	CYCLES = "CYCLES"
+
+	# Blender 2.8 to 4.1
 	BLENDER_EEVEE = "BLENDER_EEVEE"
-	# EEVEE Next is the next generation EEVEE. So in preperation for that,
-	# we've added "BLENDER_EEVEE_NEXT" as an Engine option
+
+	# Blender 4.2 and beyond
 	BLENDER_EEVEE_NEXT = "BLENDER_EEVEE_NEXT"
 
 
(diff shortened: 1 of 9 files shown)


TheDuckCow (approved):
Looks good to me. Several unit tests before this branch were failing on blender 4.2, and now they are all passing (a bit by luck, as I don't think we do any explicit render engines changes if I recall right, could be good to do some tests with cycles too I suppose one day).

...and also ignoring the `test_enable_obj_importer` one, which is failing due to `RuntimeError: Error: Add-on not loaded: "io_scene_obj", cause: No module named 'io_scene_obj'`

TheDuckCow:
The one failing test is fixed now via https://github.com/Moo-Ack-Productions/MCprep/pull/587

PR #587, opened 2024-06-09
Updated the test script for obj importing in 4.0+ and expanded gitignore

TheDuckCow:
tbh I'm a bit confused how tests were passing on github actions since the obj enabling test *should* have been failing. Let's see what happens now.


TheDuckCow:
Seeing we indeed have 65 tests run and everything passing, and that matches my 3.6 run locally, I'd say we're good now with this.

StandingPadAnimations (approved):
Looks good, merging

PR #588, opened 2024-06-17
MCprep RC Scripts to Dev

StandingPadAnimations:
This moves the RC scripts to `dev` so we can make things easier


rc-files/download_resources.sh
```diff
@@ -0,0 +1,11 @@
+# Get the latest stable release and unzip
+wget "https://github.com/Moo-Ack-Productions/MCprep/releases/download/3.5.3/MCprep_addon_3.5.3.zip"
```
TheDuckCow, line 2: Should either be a TODO or maybe even just a $1 input flag for tag version.

Could be fancier and use something like `git describe --tags --abbrev=0` to just get the latest tag and directly use that. The hard coding I just think will get overlooked is all

rc-files/HOW_TO_BUILD.md
```diff
@@ -0,0 +1,5 @@
+Step 1. Run `download_resources.sh` to add the necesary asset files
+
+Step 2. Apply all patches in the `patches` folder
```
TheDuckCow, line 3: Could be nice to include the sample command for the one patch we do have (would be this right? `git apply rc-bl_info.patch`)

TheDuckCow (approved):
Approving with some minor suggestions, could do now or in a follow up TODO

rc-files/HOW_TO_BUILD.md
StandingPadAnimations: Yeah, although we actually have 2 patches (the other changes the tracker ID so we can better track RCs in the tracker)

StandingPadAnimations:
Alright, merging now

PR #592, opened 2024-06-24
Add additional checkmark to asset submisison form

StandingPadAnimations:
We've had a few people just submit rigs and then call it a day, without commenting back at all. To counteract this, I've added a checkbox where the submitter affirms that yes, they understand that as submitter of a rig, they must be active in any discussion on said submission.

This also allows us to better justify closing submissions where the submitters haven't commented at all since submission.


TheDuckCow (approved):
No complaints from me

PR #595, opened 2024-06-30
test: Restrict test action to changes in MCprep_addon

StandingPadAnimations:
This makes it so tests only run if anything in `MCprep_addon` is changed, in order to prevent unnecessary tests on PRs that modify say documentation


TheDuckCow (approved):
Nice find! Definitely a good one

PR #603, opened 2024-07-20
Adding a missing filenames test and (commented out solution)

TheDuckCow:
Thanks to @StandingPadAnimations we have a solution to "find missing" not working. First commit here is to prove that the unit test actually fails, and then I'll push a second commit which shows the fix works.

### DIFF
--- a/MCprep_addon/materials/generate.py
+++ b/MCprep_addon/materials/generate.py
@@ -665,10 +665,10 @@ def replace_missing_texture(image: Image) -> bool:
 	# pack?
 	# Due to the issue below, must trick blender to reload the datablock
 	# https://projects.blender.org/blender/blender/issues/115984
-	#if image.source == 'FILE' and image.source != 'SEQUENCE':
-	#	old = image.source
-	#	image.source = 'SEQUENCE'
-	#	image.source = old
+	if image.source == 'FILE' and image.source != 'SEQUENCE':
+		old = image.source
+		image.source = 'SEQUENCE'
+		image.source = old
 
 	return True  # updated image block
 
--- a/MCprep_addon/materials/generate.py
+++ b/MCprep_addon/materials/generate.py
@@ -652,10 +652,7 @@ def replace_missing_texture(image: Image) -> bool:
 	env.log(f"Missing datablock detected: {image.name}")
 
 	name = image.name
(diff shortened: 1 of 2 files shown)


TheDuckCow:
In true test-driven development fashion: As intended, the firstly added test is [failing here](https://github.com/Moo-Ack-Productions/MCprep/actions/runs/10023009393/job/27703427165?pr=603#step:9:27682)

But then after pushing the fix, demonstrates the [test is now passing](https://github.com/Moo-Ack-Productions/MCprep/actions/runs/10023046582/job/27703513138?pr=603#step:10:13).

test_files/materials_test.py
```diff
@@ -98,6 +98,15 @@ def _set_test_mcprep_texturepack_path(self, reset: bool = False):
             raise Exception("Failed to set test texturepack path")
         bpy.context.scene.mcprep_texturepack_path = path
 
+    def _count_missing_files(self) -> int:
+        missing = 0
+        for img in bpy.data.images:
+            if not os.path.isfile(img.filepath):
+                missing += 1
+            elif img.channels == 0:
```
TheDuckCow, line 106: The check for channels was a good proxy to see if data is loaded or not, found anecdotally that it's set to zero if the image failed to actually load (which could be true even if the above isfile passes, due to the blender native issue of failing to load GPU textures)

MCprep_addon/materials/generate.py
```diff
@@ -666,6 +663,12 @@ def replace_missing_texture(image: Image) -> bool:
 	image.filepath = str(image_path)
 	# image.reload() # not needed?
 	# pack?
+	# Due to the issue below, must trick blender to reload the datablock
+	# https://projects.blender.org/blender/blender/issues/115984
+	if image.source == 'FILE' and image.source != 'SEQUENCE':
```
TheDuckCow, line 668: The check for SEQUENCE was necessary as it was otherwise causing a ripple effect downstream for some reason.

test_files/materials_test.py
StandingPadAnimations: I think we should add a small comment as well explaining why checking channels is needed, in case future developers come across this

StandingPadAnimations (approved):
Looks good to me

PR #611, opened 2024-08-05
Improved error message when prepping only mats with missing image paths.

TheDuckCow:
Also resolves some pep8 warnings.

Local tests pass, but let's see about the github action runner. Tested off of a version of dev initially where the find missing function wasn't yet working, to prove the new message displays then. Also cleans up error code responses (though hm maybe we should just use the MCprep error class? though it's helpful to have the specific errno codes to check against right after).

### DIFF
--- a/MCprep_addon/materials/generate.py
+++ b/MCprep_addon/materials/generate.py
@@ -30,6 +30,11 @@ from ..conf import MCprepError, env, Form
 
 AnimatedTex = Dict[str, int]
 
+# Error codes used during mat prep
+NO_DIFFUSE_NODE = 1
+IMG_MISSING = 2
+
+
 class PackFormat(Enum):
 	SIMPLE = 0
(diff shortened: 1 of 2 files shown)


StandingPadAnimations (approved):
Looks good. My only issue with it is the use of integer error codes instead of the deticated error object, but that's something we can tackle at a later date

TheDuckCow:
Yup that's fair, I can take a quick stab at it as I'm wrapping up. This is the last change before go-live anyhow.

TheDuckCow:
It was not, in fact, as simple as I had hoped - in that it really needed updating and trickle through handling of several functions, and felt weird to just do it halfway. So I created https://github.com/Moo-Ack-Productions/MCprep/issues/612 to capture what I did so far and as a place for (anyone) completing the work. Merging this PR in the meantime.

PR #613, opened 2024-08-05
Dev to master for v3.6.0

TheDuckCow:
This PR represents the merge into master for the 3.6.0 release. Automated tests are looking good, though I need to cross check whether the 2.80 and 2.90 failures are actual concerns or not. I also want to maintain the philosophy of always doing a full UAT before a release if there have been any code changes since the last UAT (which there were). So, I'll aim to do that tomorrow.

```
(4.2.0)   	all_tests	69	2	0	No errors
(4.0.2)   	all_tests	69	2	0	No errors
(3.6.2)   	all_tests	69	2	0	No errors
(3.5.1)   	all_tests	69	2	0	No errors
(3.4.0)   	all_tests	69	2	0	No errors
(3.3.1)   	all_tests	69	2	0	No errors
(3.2.1)   	all_tests	69	2	0	No errors
(3.1.0)   	all_tests	69	2	0	No errors
(3.0.0)   	all_tests	69	2	0	No errors
(2.93.0)   	all_tests	68	3	0	No errors
(2.90.1)   	all_tests	8	0	8	test_disable_enable;test_enable;asset_qa_test;materials_test;spawner_test;tracker_test;util_test;world_tools_test
(2.80.75)   	all_tests	8	0	8	test_disable_enable;test_enable;asset_qa_test;materials_test;spawner_test;tracker_test;util_test;world_tools_test
tests took 226s to run, ending with code 1
```

A side goal of this would be to create a UAT template issue which could be used and attached to releases in the future. For now, will just do so manually.


TheDuckCow:
Alright I did another round of UAT and found just a couple things - some are quick asset fixes I'll attempt, others can be pursued in follow up updates. I've also used this as a chance to create a draft issue template for UAT testing, which I'll share later as another PR:

- Default imported fire renders black in eevee, should select alpha clip for both materials. Same for campfire materials (meshswap).
- The "furnace on" meshswap has a visible white box (probably for particle spawning) which should be changed to not be visible in render (in 4.2 both eevee and cycles)
- Add MC sky: Could not use the "add sun/moon mesh" option in 2.80 nor in 4.2, this seems pretty important to fix as the option for mesh is basically broken right now (not sure why yet)
- In blender 2.80: Render pano started with 1/6 done, should have started saying 0/6 done; also is rendering panos just broken entirely? Doesn't seem to go past the first one.
  - Works fine in 4.x though, so really just need to see where no longer working and just disable this niche feature in that version and lower I'd say, rather than spend time fixing
- Crashing when adding particle effects in 2.80
  - Working fine in 4.2 though, so will consider deprecating the feature in this version and see which b version doesn't crash.

Will see if I can address these things in the morning,

TheDuckCow:
I've updated the couple assets locally. I also did a check against the current released version of MCprep, and indeed we also can't add the mesh sun/moon there either (in either Cycles or Eevee; also it's much more broken in general in blender 4.2).

So, it's definitely in the best interest to release this ASAP and just treat the mesh sky as an improvement.

PR #619, opened 2024-08-12
fix: Remove references to MCprep optimizer

### LINKED ISSUE #618: Remove leftover optimizer references in prep mats
So far the most common bug report we have with the new version is due to some left over code of the optimizer. I guess we didn't test every permutation of settings 🤷 

At this point, we have 35 users who have encountered this error:

```
^^^^^^^^^^^^^^^^^^^^^^^
 File "<addon_path>/prep.py", line 302, in execute
 bpy.ops.mcprep.optimize_scene()
 File "<addon_path>/ops.py", line 109, in __call__
 ret = _op_call(self.idname_py(), kw)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 AttributeError: Calling operator "bpy.ops.mcprep.optimize_scene" error, could not be found

```

Due to the following lines (though we should also just remove the optimize checkmark everywhere too); doing a search, I don't see any other places where `optimize_scene` is used

```python
		if self.optimizeScene and engine == 'CYCLES':
			bpy.ops.mcprep.optimize_scene()
```

StandingPadAnimations:
Resolves #618

### DIFF
--- a/MCprep_addon/materials/prep.py
+++ b/MCprep_addon/materials/prep.py
@@ -74,10 +74,6 @@ class McprepMaterialProps():
 		name="Improve UI",
 		description="Automatically improve relevant UI settings",
 		default=True)
-	optimizeScene: bpy.props.BoolProperty(
-		name="Optimize scene (cycles)",
-		description="Optimize the scene for faster cycles rendering",
-		default=False)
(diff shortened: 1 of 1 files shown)


TheDuckCow (approved):
Looks good to me, thanks for the quick fix

PR #620, opened 2024-08-12
fix: Allow an exporter to be opened without an object selected

StandingPadAnimations:
This fixes an edge case where a user may want to open an exporter from MCprep without an object selected, such as in a startup file without the default cube.

### DIFF
--- a/MCprep_addon/mcprep_ui.py
+++ b/MCprep_addon/mcprep_ui.py
@@ -753,7 +753,11 @@ class MCPREP_PT_world_imports(bpy.types.Panel):
 		row = col.row(align=True)
 		row.prop(addon_prefs, "MCprep_exporter_type", expand=True)
 		row = col.row(align=True)
-		exporter = world_tools.get_exporter(context)
+
+		compliant_exporter = world_tools.get_exporter(context)
+		exporter = compliant_exporter if compliant_exporter is not None else world_tools.get_explicit_exporter_legacy(context)
+		del compliant_exporter
+
 		if exporter is None:
 			row.operator(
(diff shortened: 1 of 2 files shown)


TheDuckCow (approved):
Base change, looks good. I realize we can still end up with mismatches since we're specifying a fallback, but we know those fallbacks aren't likely for valid MC common obj exports anyways, so seems safe to me.

I'm suggesting we go slightly further though to ensure the UI remains consistent, as indeed [this UI](https://github.com/Moo-Ack-Productions/MCprep/issues/621) is weird to see.

See my suggested patch, not committing to avoid messing up your git branch in case you have anything locally:

```
diff --git a/MCprep_addon/mcprep_ui.py b/MCprep_addon/mcprep_ui.py
index 418fe4a..651f705 100644
--- a/MCprep_addon/mcprep_ui.py
+++ b/MCprep_addon/mcprep_ui.py
@@ -788,7 +788,9 @@ class MCPREP_PT_world_imports(bpy.types.Panel):
                if context.mode == "OBJECT":
                        col.operator("mcprep.meshswap", text=env._("Mesh Swap"))
                        exporter = world_tools.get_exporter(context)
-                       if exporter is None or exporter is world_tools.WorldExporter.Unknown:
+                       if not context.object and not context.selected_objects:
+                               # Ops will return nothing selected anyways, don't show warning
+                               pass
+                       elif exporter is None or exporter is world_tools.WorldExporter.Unknown:
                                col.label(text=env._("Select exporter!"), icon='ERROR')
                if context.mode == 'EDIT_MESH':
                        col.operator("mcprep.scale_uv")
```

If you can add that change in, I'd be good to merge this one too.

PR #624, opened 2024-08-20
build: Rename blender_manifest.toml to avoid bugs with 4.2 and bl_info + fix: Add max bound for BpyBuild

StandingPadAnimations:
This renames `blender_manifest.toml` to `_blender_manifest.toml` to avoid issues in 4.2 when building legacy addons. When an extension build is created, this is changed to the correct filename.

In addition, since BpyBuild 0.5 will have breaking changes, a max bound is added to avoid developers accidentally installing 0.5 before we migrate to it.


TheDuckCow (approved):
Seems to install and work fine, no complaints from my side - thanks for thinking ahead for other developers

PR #629, opened 2024-08-24
Using updated production bab build command

TheDuckCow:
Appears this change fixes the building problem. Confirmed that this results in both no dev txt file and the mo files are indeed now present, where they weren't in the prior build.


TheDuckCow:
Going to proceed with the merge, but let me know @StandingPadAnimations if you have any post comments

Issue #627, opened 2024-08-23
MCprep 3.6.1 Update is Missing Compiled Translations


StandingPadAnimations:
### Did the issue happen after restarting Blender?

- [X] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [X] I have checked known issues, and my problem is different

### What happened?

MCprep 3.6.1 is missing compiled translation files, which causes the following erorr for end users:
```
Traceback (most recent call last):
  File "/usr/share/blender/4.2/scripts/modules/addon_utils.py", line 407, in enable
    mod = importlib.import_module(module_name)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/importlib/__init__.py", line 90, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 995, in exec_module
  File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
  File "/home/mahid/.config/blender/4.2/scripts/addons/MCprep_addon/__init__.py", line 59, in <module>
    from . import load_modules
  File "/home/mahid/.config/blender/4.2/scripts/addons/MCprep_addon/load_modules.py", line 28, in <module>
    from . import conf
  File "/home/mahid/.config/blender/4.2/scripts/addons/MCprep_addon/conf.py", line 325, in <module>
    env = MCprepEnv()
          ^^^^^^^^^^^
  File "/home/mahid/.config/blender/4.2/scripts/addons/MCprep_addon/conf.py", line 149, in __init__
    self.languages[language.name] = gettext.translation("mcprep",
                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/gettext.py", line 538, in translation
    t = _translations.setdefault(key, class_(fp))
                                      ^^^^^^^^^^
  File "/usr/lib/python3.12/gettext.py", line 272, in __init__
[...]

PR #628, opened 2024-08-23
fix: Gracefully handle exceptions during direct i18n loading

StandingPadAnimations:
If translations.py doesn't exist, MCprep moves to direct i18n, which directly uses MO files for translations (at the cost of some UI updating issues). However, if these MO files do not exist, then MCprep will throw an exception. Since this exception occurs in the creation of `MCprepEnv`, MCprep is rendered unusable.

This patch adds a basic try-except statement to allow `MCprepEnv` to handle these exceptions gracefully.

PR #630, opened 2024-08-24
More graceful translations startup handling

TheDuckCow:
Closes https://github.com/Moo-Ack-Productions/MCprep/issues/627 hopefully for real.

Was to fix this error that was seen on install/startup with the initial v3.6.1.1 release which has now been pulled:

```
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/scripts/modules/addon_utils.py", line 407, in enable
    mod = importlib.import_module(module_name)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/__init__.py", line 59, in <module>
    from . import load_modules
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/load_modules.py", line 118, in <module>
    from . import mcprep_ui
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/mcprep_ui.py", line 106, in <module>
    class MCPREP_MT_mob_spawner(bpy.types.Menu):
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/mcprep_ui.py", line 108, in MCPREP_MT_mob_spawner
    bl_label = env._("Mob Spawner")
               ^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/conf.py", line 166, in _
    if bpy.context.preferences.view.language in self.languages:
                                                ^^^^^^^^^^^^^^
AttributeError: 'MCprepEnv' object has no attribute 'languages'

```

Working on another patch for you to review.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,12 @@ venv/
 *.sublime-*
 MCprep_addon/import_bridge/conf
 MCprep_addon/import_bridge/nbt
-MCprep_addon/MCprep_resources/
+MCprep_addon/MCprep_resources/*.blend
+MCprep_addon/MCprep_resources/mcprep_data.json
+MCprep_addon/MCprep_resources/effects
+MCprep_addon/MCprep_resources/resourcepacks
+MCprep_addon/MCprep_resources/rigs
+MCprep_addon/MCprep_resources/skins
+MCprep_addon/MCprep_resources/textures
+!MCprep_addon/MCprep_resources/Languages/*
 MCprep_addon/.vscode
--- a/MCprep_addon/MCprep_resources/Languages/zh_CN/LC_MESSAGES/mcprep.po
+++ b/MCprep_addon/MCprep_resources/Languages/zh_CN/LC_MESSAGES/mcprep.po
@@ -1,0 +1,596 @@
+# SOME DESCRIPTIVE TITLE.
+# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
+# This file is distributed under the same license as the PACKAGE package.
+# FIRST AUTHOR <EMAIL@ADDRESS>, YEAR.
+#
+msgid ""
+msgstr ""
+"Project-Id-Version: \n"
+"Report-Msgid-Bugs-To: \n"
+"POT-Creation-Date: 2024-02-03 22:03-0600\n"
+"PO-Revision-Date: 2024-02-03 22:16-0600\n"
+"Last-Translator: \n"
+"Language-Team: \n"
+"Language: zh_CN\n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"X-Generator: Poedit 3.4.1\n"
+
+#: MCprep_addon/mcprep_ui.py:98
+msgid "Restart blender"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:100
+msgid "to complete update"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:109
+msgid "Mob Spawner"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:111
+msgid "Menu for placing in the shift-A add object menu"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:121
+msgid "Load mobs"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:150
+msgid "Meshswap Objects"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:157
+msgid "No meshswap blocks found!"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:179
+msgid "Item Spawner"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:185
+msgid "No items found!"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:202
+msgid "Effects Spawner"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:245
+msgid "Entity Spawner"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:252
+msgid "No entities found!"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:268
+msgid "Model Spawner"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:274
+msgid "No models found!"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:314
+msgid "Load spawners"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:516
+msgid "World Importing & Meshswapping"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:520
+msgid "Default Exporter:"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:525
+msgid "jmc2obj executable"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:530
+msgid "Mineways executable"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:535
+msgid "World OBJ Exports Folder"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:540
+msgid "Meshwap assets"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:546 MCprep_addon/mcprep_ui.py:871
+#: MCprep_addon/mcprep_ui.py:1337 MCprep_addon/mcprep_ui.py:1485
+msgid "MeshSwap file not found"
+msgstr "未找到网格变形文件"
+
+#: MCprep_addon/mcprep_ui.py:550
+msgid "Entity assets"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:556 MCprep_addon/mcprep_ui.py:1435
+msgid "Entity file not found"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:562 MCprep_addon/mcprep_ui.py:1658
+msgid "Effects folder not found"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:566
+msgid "Texture / Resource packs"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:570 MCprep_addon/mcprep_ui.py:843
+#: MCprep_addon/mcprep_ui.py:1651
+msgid "Texture pack folder"
+msgstr "材质包文件夹"
+
+#: MCprep_addon/mcprep_ui.py:575
+msgid "Install to folder"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:579
+msgid "Open texture pack folder"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:584
+msgid "Mob spawning"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:588
+msgid "Rig Folder"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:593
+msgid "Select/install mobs"
+msgstr ""
+
+#: MCprep_addon/mcprep_ui.py:596
+msgid "Install file for mob spawning"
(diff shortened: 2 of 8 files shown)


TheDuckCow:
It's late, and this is still causing startup errors now. Now I'm getting errors like:

```
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/conf.py", line 166, in _
    return self.languages["en_US"].gettext(msg)
           ~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'en_US'
Error: 'en_US'
```

@StandingPadAnimations need your input here, since you're more familiar with the translation startup sequence. Whatever fix we try, we should please try to do an install over a previously prod v3.6.0 release to sanity check it's working, that's where I keep hitting a wall. You have my permission to take over this branch if you have ideas of a more suitable fix.

In general, I do think it's not ideal to only conditionally define a property to exist, which was the reason for the original bug above. I'm also not sure why it's in the branch where content isn't loaded, since I've confirmed my local build's zip does indeed contain the .mo and .po files for each of the languages above.

MCprep_addon/conf.py
```diff
@@ -137,28 +134,28 @@ def __init__(self):
...
 		# If ever changing the resource pack, should also reset to None.
 		self.material_sync_cache: List = []
-	
+
 		# Whether we use PO files directly or use the converted form
 		self.use_direct_i18n = False
 		# i18n using Python's gettext module
 		#
 		# This only runs if translations.py does not exist
+		self.languages: dict[str, gettext.NullTranslations] = {}
```
TheDuckCow, line 143: This is the only main change introduced, ensuring the property is not defined only within an if statement. But it causes problems later I guess

MCprep_addon/conf.py
```diff
@@ -137,28 +134,28 @@ def __init__(self):
...
-					self.languages[language.name] = gettext.translation("mcprep", 
-											 self.languages_folder, 
-											 fallback=True,
-											 languages=[language.name])
+					self.languages[language.name] = gettext.translation(
+						"mcprep",
+						self.languages_folder,
+						fallback=True,
+						languages=[language.name])
 			self.use_direct_i18n = True
```
StandingPadAnimations, line 152: This and the line underneath need to be indented by one more tab. I think #628 didn't properly indent these

TheDuckCow: Do you mean from a wrong behavior or codestyle perspective? Hard to see on github, but it looks good to my eyes and aligns with pep8 (unintended 2x for clarify)

```python
if not self.translations.exists():
	for language in self.languages_folder.iterdir():
		self.languages[language.name] = gettext.translation(
			"mcprep",
			self.languages_folder,
			fallback=True,
			languages=[language.name])
self.use_direct_i18n = True
self.log("Loaded direct i18n!")
```

StandingPadAnimations: It's wrong behavior. `self.use_direct_i18n` is what tells `MCprepEnv._` to use the MO files directly, and thus should be in the if condition.

(Last-night me forgot to move that over when adding the try-except block)

TheDuckCow:
FYI I think I have a decent fix locally, will push in a bit. Doing some testing

MCprep_addon/conf.py
```diff
@@ -137,30 +134,38 @@ def __init__(self):
...
-			self.use_direct_i18n = True
-			self.log("Loaded direct i18n!")
-
-		except Exception:
+					print(prefix, "In langdir:", language.name)
+					self.languages[language.name] = gettext.translation(
+						"mcprep",
+						localedir=self.languages_folder,
+						fallback=True,
+						languages=[language.name])
```
TheDuckCow, line 157: Hey @StandingPadAnimations are we sure this function is working well in Blender? I'm stuck here, can't get it to work properly in the test suite, despite confirming I'm pointing it to the right folder location and everything. Translations work in my blender instance, but that's because the translations.py file shows up (but if we use that, then we can't do direct tests for translations working below, which is the point).

This is the internal error that it runs into when calling the `self.languages[language.name] = gettext.translation(` line:

```
ERROR: test_translations (addon_test.AddonTest.test_translations)
Safely ensures translations are working fine still.
  File "/Users/patrickcrawford/Documents/gits/mcprep/test_files/addon_test.py", line 53, in test_translations
    self._test_translation()
  File "/Users/patrickcrawford/Documents/gits/mcprep/test_files/addon_test.py", line 75, in _test_translation
    test_env._load_translations(prefix=">>>\t")
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/conf.py", line 165, in _load_translations
    raise e
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/conf.py", line 153, in _load_translations
    self.languages[language.name] = gettext.translation(
                                    ^^^^^^^^^^^^^^^^^^^^
  File "/Applications/blender 4.2/Blender.app/Contents/Resources/4.2/python/lib/python3.11/gettext.py", line 538, in translation
    t = _translations.setdefault(key, class_(fp))
                                      ^^^^^^^^^^
  File "/Applications/blender 4.2/Blender.app/Contents/Resources/4.2/python/lib/python3.11/gettext.py", line 272, in __init__
    self._parse(fp)
  File "/Applications/blender 4.2/Blender.app/Contents/Resources/4.2/python/lib/python3.11/gettext.py", line 424, in _parse
    catalog[str(msg, charset)] = str(tmsg, charset)
                                 ^^^^^^^^^^^^^^^^^^
LookupError: unknown encoding: CHARSET
```

test_files/addon_test.py
```diff
@@ -33,6 +42,63 @@ def test_disable_enable(self):
...
+            ("zh_HANS", "Texture pack folder", "材质包文件夹"),
+            ("en_US", "Mob Spawner", "Mob Spawner"),
+        ]
+        for lang, src, dst in test_translations:
+            with self.subTest(lang):
+                # First ensure the mo files exist
+                self.assertTrue(
+                    os.path.isfile(
+                        lang_folder / lang / "LC_MESSAGES" / "mcprep.mo"),
+                    f"Missing {lang}'s mo file")
```
TheDuckCow, line 95: If nothing else, we could reduce the test down to just this, which already goes a long way to prove that translations are present when they should be.

MCprep_addon/conf.py
StandingPadAnimations: I think I've traced the source of the problem
https://github.com/Moo-Ack-Productions/MCprep/blob/b72a7a38bb79ae70ebffb464751344182ac4cfe6/MCprep_addon/MCprep_resources/Languages/en_US/LC_MESSAGES/mcprep.po#L17

Try replacing `CHARSET` with `UTF-8` (to match the Chinese and Russian translations)

TheDuckCow: Legendary! Thanks a mill, I'll clean this branch up and formally set to review in a bit - need to step away

```
(4.2.0)   	test_translations	1	0	0	No errors

```

test_files/addon_test.py
```diff
@@ -33,6 +42,58 @@ def test_disable_enable(self):
...
+        self.assertIn(
+            "en_US", test_env.languages, "Missing default translation key")
+        self.assertTrue(
+            test_env.use_direct_i18n, "use_direct_i18n should be True")
+
+        # Magic string evaluations, will break if source po's change
+        test_translations = [
+            ("ru_RU", "Restart blender", "Перезапустите блендер"),
+            # Blender 4.0+ only has 'zh_HANS', 'zh_HANT'
+            ("zh_HANS" if bpy.app.version > (4, 0) else "zh_CN", "Texture pack folder", "材质包文件夹"),
```
TheDuckCow, line 82: Explanation here, and also why I've added a new PO file (simplest option I could think of without further consequences/code changes)

Thanks to this test, I discovered we weren't supporting Chinese pre 4.0 due to Blender splitting into more sub regional names. The enum key zh_HANT only exists as os 4.0 (thus this test was failing pre blender 4.0), and meanwhile zh_CN only exists before. I understand from some reading that zh_HANT is more regional version of simplified Chinese, so I feel it's appropriate as a stopgap for now.

TheDuckCow (commented):
That was a doozy, one thing after another - but, we're there. All local tests pass, as do the ones on github actions above thankfully

```
(4.2.0)   	all_tests	70	2	0	No errors
(3.4.0)   	all_tests	70	2	0	No errors
(3.3.1)   	all_tests	70	2	0	No errors
(3.2.1)   	all_tests	70	2	0	No errors
(2.90.1)   	all_tests	69	3	0	No errors
Compiled in 30.6s + tests ran in 269.9s
Total of 300.5s with exit code 0
```

.gitignore
```diff
@@ -19,5 +19,12 @@ venv/
 *.sublime-*
 MCprep_addon/import_bridge/conf
 MCprep_addon/import_bridge/nbt
-MCprep_addon/MCprep_resources/
+MCprep_addon/MCprep_resources/*.blend
```
TheDuckCow, line 22: Before this change, we were masking .po files, felt it was worth making sure we had it recognized explicitly.

MCprep_addon/MCprep_resources/Languages/zh_CN/LC_MESSAGES/mcprep.po
TheDuckCow, line 1: An as-is copy of the zh_HANS.

I live-tested in both blender 3.x and 4.x to make sure that translations work fine in both cases (I'd say, it feels odd that the translations actually were working in blender 4.0 before since the contents of the po reference zh_CN and not zh_HANS, but I'm not going to dive into understanding that right now)

StandingPadAnimations: If you're curious, `zn_CN` is the old code for Simplified Chinese. It was changed to `zn_HANS` since many countries outside of China also use Simplified Chinese (at least that's what I was able to gather when I looked into it a while back)

MCprep_addon/conf.py
```diff
@@ -81,11 +75,14 @@ class MCprepEnv:
...
-		self.json_path_update: Path = Path(os.path.dirname(__file__), "MCprep_resources", "mcprep_data_update.json")
+		self.json_path: Path = Path(
+			os.path.dirname(__file__), "MCprep_resources", "mcprep_data.json")
+		self.json_path_update: Path = Path(
+			os.path.dirname(__file__), "MCprep_resources", "mcprep_data_update.json")
 
 		self.dev_file: Path = Path(os.path.dirname(__file__), "mcprep_dev.txt")
-		self.languages_folder: Path = Path(os.path.dirname(__file__), "MCprep_resources", "Languages")
+		self.languages_folder: Path = Path(
+			os.path.dirname(__file__), "MCprep_resources", "Languages")
```
StandingPadAnimations, line 85: I think we should have a `MCPREP_RESOURCES` constant just to reduce repetition. Then, we can just make a `Path` object for each of these like this:
```py
MCPREP_RESOURCES: Path =  Path(os.path.dirname(__file__), "MCprep_resources")
#...
self.languages_folder: Path = Path(MCPREP_RESOURCES, "Languages")
```

Not a blocker though

StandingPadAnimations (approved):
Looks good to me. Just a minor stylistic nitpick

MCprep_addon/conf.py
TheDuckCow: Done

PR #639, opened 2024-10-11
build: Update BpyBuild to 0.5

StandingPadAnimations:
This updates the toolchain to use BpyBuild 0.5, and with it some improvements have been made to our config and actions:
- BpyBuild config now uses the new version shorthands introduced in 0.4.X (which we didn't use to avoid breakage)
- Our override `extension` action now takes advantage of `lib_bpybuild_ext` to perform manifest verification and compatibility checks in the MCprep code. In addition, the action now does a search and replace in `__init__.py` for `bl_info` when building as an extension, which A. removes the need for a patch that constantly breaks and B. means the extension build will work out of the box
- MCprep now builds both an extension and legacy build. Although we'll likely only continue to offer the legacy build for the foreseeable future, it's nice to build both just for wider coverage.

This being a major release, there are definitely some bugs I've caught while working on this update (which will be fixed in the coming weeks), but nothing that can't be worked around (they're all "out of the box" bugs so to speak, we can easily work around them).

Edit October 11th, 2024: Pushed some bug fixes on https://github.com/Moo-Ack-Productions/bpy-build/pull/21


bpy-build.yaml
```diff
@@ -1,22 +1,13 @@
...
-  - 3.3
-  - 3.2
-  - 3.1
-  - 3.0
-  - 2.93
-  - 2.9
-  - 2.83
-  - 2.8
+  - 4.3+
+  - 2.8..4.3
```
TheDuckCow: Is it possible to do this or no?

```suggestion
  - 2.83..4.3
```

TheDuckCow (commented):
Leaving as a comment initially, will test this again once we have the final bpy-build update ready (presuming the other recent PRs there should included right?)

bpy-build.yaml
StandingPadAnimations: Yep, that's possible as well, though are we dropping 2.8 to 2.82 support?

StandingPadAnimations:
@TheDuckCow tagging you to let you know that this is ready for further review. BpyBuild 0.5.1 is also out, which fixes the main issues I caught while working on this (though BpyBuild 0.5 also works fine with this config, so I don't think it's necessary to raise the minimum)

StandingPadAnimations:
I do have a few questions though:
- Assuming we're only providing legacy builds for the foreseeable future, is the updated config (where we build both an extension and legacy addon) fine as is, or would it be preferable to only build legacy builds?
- If we build both an extension and legacy addon, would it be preferable to have some slight name change for the extension build (again assuming we're only providing legacy builds in the long run), maybe something like "MCprep Extension Build"? 
- Slightly off-topic, but I noticed it was suggested that the minimum be changed to 2.83, are we planning on dropping 2.8 to 2.82 with the update?

StandingPadAnimations:
@TheDuckCow just tagging as a reminder since it's been a while

TheDuckCow:
Sorry for such a long delay, I'll try to prioritize this by the end of the week and ask any follow up questions.

> Slightly off-topic, but I noticed it was suggested that the minimum be changed to 2.83, are we planning on dropping 2.8 to 2.82 with the update?

I recall an issue with test runners or something in 2.80 but was working in blender 2.83, I think it's more than fair to set that as a the minimum at this stage, being the LTS version.

bpy-build.yaml
```diff
@@ -6,8 +6,8 @@ extension_settings:
   build_legacy: true
 
 install_versions:
-  - 4.3+
-  - 2.83..4.3
+  - 4.4+
+  - 2.83..4.4
```
TheDuckCow, line 10: 👍  was going to suggest this change

TheDuckCow (approved):
Approving, but two change requests - really just expecting that the first one is done. If more changes than that are made, I can be re-requested

1. The addon now seems to install do a different path; on my system, that results in having two different MCprep addons installed at once (see screenshot below). I don't want any chance or accident of us distributing such a version, so I'd suggest we disable this - at *least* until we provide extension builds officially, which I'm still in favor of pushing out since we don't benefit really from it yet and will only result in users installing the wrong version, since extensions for some reason were chosen to still be .zip files (sigh, we could have had .bzp's or something else like 3ds Max's mzp setup which works nicely). See this screenshot, where I had an older version of MCprep installed (hence the addon one without legacy in the name): 
  - One option in the future would be to just rename the zip file to have "legacy" in it, but then the actual folder contained inside is still MCprep_addon. 
3. Looks like the build script installs both the extension and addon together, so for 4.2+ we are getting both installed at once. I'd argue for disabling extension install by default, but I guess it's up to the developer to choose which one they test so I'm a bit torn. Your call I'd say. I'll mostly be testing the addon form though, since that's what we release currently.

Console output below:

```
(venv) mcprep $ ./compile-run.sh 
Building POT...
Building MO files...
Building Translations...
Patching bl_info
Setting up the Blender Manifest
Performing validation of extension
Installed to /Users/patrickcrawford/Library/Application Support/Blender/4.4/extensions/user_default
Installed to /Users/patrickcrawford/Library/Application Support/Blender/4.4/extensions/user_default
Installed to /Users/patrickcrawford/Library/Application Support/Blender/4.3/extensions/user_default
Installed to /Users/patrickcrawford/Library/Application Support/Blender/4.2/extensions/user_default
Building POT...
Building MO files...
Building Translations...
Installed to /Users/patrickcrawford/Library/Application Support/Blender/4.4/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/4.4/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/4.2/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/4.1/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/3.6/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/3.5/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/3.4/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/3.3/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/3.2/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/3.1/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/2.93/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/2.92/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/2.91/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/2.90/scripts/addons
Installed to /Users/patrickcrawford/Library/Application Support/Blender/2.83/scripts/addons
(venv) mcprep $ 
```

StandingPadAnimations:
Disabled extension builds by default, so both issues noted should be resolved. I'll merge this in

TheDuckCow:
Thanks!

TheDuckCow:
Thanks!

Edit, i realize I forgot to include the screenshot before so anyhow here it is as I am cleaning up my desktop
Screen Shot 2025-02-15 at 9 02 17 AM

StandingPadAnimations:
It should automatically be reverted with `build_extension` disabled

PR #643, opened 2024-12-17
Fix resourcepack format version 22 `short_grass` and filtering out block models

zNightlord:
Small fix on the grass format issue, haven't tested on JMC, Mineways so far MCprep default and Barebone 1.21 pack work. The "colormap/grass" isn't supposed to be there anyway, I think.

And filter out most of the no geometry json file, hope I didn't miss anything.


zNightlord:
Just noticed the test with the colormap/grass issue.
```diff
diff --git a/MCprep_addon/materials/generate.py b/MCprep_addon/materials/generate.py
index 04a5e69..d10060d 100644
--- a/MCprep_addon/materials/generate.py
+++ b/MCprep_addon/materials/generate.py
@@ -16,6 +16,7 @@
 #
 # ##### END GPL LICENSE BLOCK #####
 
+import json
 import os
 from typing import Dict, Optional, List, Any, Tuple, Union, cast
 from pathlib import Path
@@ -220,6 +221,31 @@ def find_from_texturepack(blockname: str, resource_folder: Optional[Path]=None) 
        return res
 
 
+def get_format_version_texturepack(resource_folder: Optional[Path]=None) -> Union[int, MCprepError]:
+       """ Get texturepack format version"""
+       if resource_folder is None:
+               # default to internal pack
+               resource_folder = Path(cast(
+                       str,
+                       bpy.path.abspath(bpy.context.scene.mcprep_texturepack_path)
+               ))
+
+       if not resource_folder.exists() or not resource_folder.is_dir():
+               env.log("Error, resource folder does not exist")
+               line, file = env.current_line_and_file()
+               return MCprepError(FileNotFoundError(), line, file, f"Resource pack folder at {resource_folder} does not exist!")
+
+       # Resource folder is ame level as assets folder
+       file = Path(resource_folder, "pack.mcmeta")
+       if (file.is_file()):
+               with open(file, 'r') as f:
+                       data = json.load(f)
+                       print(data)
+                       return data["pack"]["pack_format"]
+       # return the unaffected change version
+       return 21
+
+
 def detect_form(materials: List[Material]) -> Optional[Form]:
        """Function which, given the input materials, guesses the exporter form.

@@ -371,6 +397,10 @@ def set_texture_pack(
        run the swap (and auto load e.g. normals and specs if avail.)
        """
        mc_name, _ = get_mc_canonical_name(material.name)
+       texture_packformat = get_format_version_texturepack(folder)
+       if (material.name == "grass" and texture_packformat > 21):
+               mc_name = "short_grass"
        image = find_from_texturepack(mc_name, folder)
        if isinstance(image, MCprepError):
                if image.msg:
                 ```
 Maybe a different approach rather changing the json file, checking the pack format version and condition from there.

File: MCprep_addon/materials/generate.py [hunk 1]
```diff
@@ -216,6 +217,32 @@ def find_from_texturepack(blockname: str, resource_folder: Optional[Path]=None)
...
+		return MCprepError(FileNotFoundError(), line, file, f"Resource pack folder at {resource_folder} does not exist!")
+
+	# Resource folder is same level as assets folder
+	file = Path(resource_folder, "pack.mcmeta")
+	if (file.is_file()):
+		with open(file, 'r') as f:
+			data = json.load(f)
+			return data["pack"]["pack_format"]
+	# return the unaffected change version, 22
+	return 21
```
StandingPadAnimations, line 250: All `pack.mcmeta` files should define the pack format, so I wouldn't include this catch-all `return 21`. It could result in some silent errors.

I'd prefer an `MCprepError` object be returned if `pack.mcmeta` doesn't contain the pack format or if the file doesn't exist. To me at least, that's a bug with the resource pack.

A check for the `pack` and `pack_format` keys should also be added, just because there's no guarantee that either key will exist. I would have also suggested a `TypedDict` since `pack.mcmeta` has a well-defined structure, but I don't think Blender 2.8X has a new enough Python version for that.

MCprep_addon/materials/generate.py
```diff
@@ -367,6 +394,10 @@ def set_texture_pack(
 	run the swap (and auto load e.g. normals and specs if avail.)
 	"""
 	mc_name, _ = get_mc_canonical_name(material.name)
+	texture_pack_format = get_format_version_texturepack(folder)
+	# grass rename to short_grass in MC 1.20.3 with pack format version 22
+	if (material.name == "grass" and texture_pack_format > 21):
+		mc_name = "short_grass"
```
StandingPadAnimations, line 411: Maybe we should extend this further to all changes, just to properly handle older resource packs, though I'll defer that decision to @TheDuckCow

Also, I think we should make sure that the OBJ actually needs this particular change. OBJs from worlds predating this change won't need us to handle pack format 22 compatibility, though checking might be a bit of a pain (this is making me wish CommonMCOBJ V1 included the world's version in the header, probably will open a proposal for that for V2 in the next couple of days)

MCprep_addon/materials/generate.py
```diff
@@ -216,6 +217,32 @@ def find_from_texturepack(blockname: str, resource_folder: Optional[Path]=None)
...
+		))
+
+	if not resource_folder.exists() or not resource_folder.is_dir():
+		env.log("Error, resource folder does not exist")
+		line, file = env.current_line_and_file()
+		return MCprepError(FileNotFoundError(), line, file, f"Resource pack folder at {resource_folder} does not exist!")
+
+	# Resource folder is same level as assets folder
+	file = Path(resource_folder, "pack.mcmeta")
+	if (file.is_file()):
```
StandingPadAnimations: Extra parenthesis unnecessary, `if file.is_file()` should be sufficient. 

Also suggest extending this to `if file.exists() and file.is_file()` for consistency reasons

StandingPadAnimations (changes requested):
Looks good for the most part, though a couple of changes are needed.

One thing I think we may need more discussion on is how we should handle older pack formats. Generally, we include the newest resource pack at the time with the assets, which does open a can of worms with compatibility.

MCprep_addon/materials/generate.py
zNightlord: I was thinking of the current default MCprep resourcepack doesn't have the meta file and if you make it return like that this whole [check_dirs](https://github.com/Moo-Ack-Productions/MCprep/blob/d1eb1465f81a214ef9acb809f16ee7f79258db92/MCprep_addon/materials/generate.py#L162) part is considered as unnecessary, throw away the legacy behavior, you can swap any sub folder like from the "textures" folder can also work.

zNightlord:
Think more on it, we can also introduce `short_grass` as it own block rather map it back to grass.
Since the problem comes from using pre 1.20.3 exporters which is "grass" and 1.20.3+ exporters would also have the "short_grass" to use.

EDIT: Additional gives warn based on versions detected "short_grass" or "grass" exists in the resourcepack but the materials has the opposite of "short_grass" or "grass".

TheDuckCow:
> Think more on it, we can also introduce short_grass as it own block rather map it back to grass.

I'm ok with splitting things out more widely. The main consideration from my side is trying to keep parity as much as possible with the generated outputs of the mapping scripts below. Not sure if that influences anything here, I only lightly brushed up on the issue

https://github.com/Moo-Ack-Productions/MCprep/blob/master/mcprep_data_refresh.py

Which is used to generate the updates for:
https://github.com/Moo-Ack-Productions/MCprep/blob/master/MCprep_addon/MCprep_resources/mcprep_data_update.json

To avoid ever making _manual_ changes to the above file, I do also have this base dictionary template below which is hand authored for special cases - so maybe that's an option here too, if we need to override some behavior and split something out:
https://github.com/Moo-Ack-Productions/MCprep/blob/master/mcprep_data_base.json

MCprep_addon/MCprep_resources/mcprep_data_update.json
TheDuckCow, line 1: Per callout above, we want to modify the source mappings, not this file itself - otherwise it'll get overwritten directly.

zNightlord:
Close in favor of #655

PR #644, opened 2024-12-18
Fix block model leaves vegetation not doing saturated color

### LINKED ISSUE #642: Spawned vegetation can't be preped

- [X] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [X] I have checked known issues, and my problem is different

### What happened?

When I click “prep materials” on a spawned leaves block, vines or any other “gray stuff”, nothing happens.

1) Spawn Acacia Leaves 
2) Press "Prep Materials"

Mcprep 3.6.1.2

Blender 4.3.1

zNightlord:
Fix #642

Change `add_material()` to `add_get_material()` to get reuse material, avoid creating a .001 .002 materials mess.

### DIFF
--- a/MCprep_addon/spawner/mcmodel.py
+++ b/MCprep_addon/spawner/mcmodel.py
@@ -116,11 +116,14 @@ def add_element(
 	return verts, edges, faces
 
 
-def add_material(
+def add_get_material(
 	name: str = "material", path: str = "", use_name: bool = False
 ) -> Optional[Material]:
-	"""Creates a simple material with an image texture from path."""
+	"""Creates or get an existing created simple material with an image texture from path."""
 	engine = bpy.context.scene.render.engine
+	mat = bpy.data.materials.get(name)
+	if mat:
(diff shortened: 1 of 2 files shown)


MCprep_addon/spawner/mcmodel.py
```diff
@@ -116,11 +116,14 @@ def add_element(
 	return verts, edges, faces
 
 
-def add_material(
+def add_get_material(
```
StandingPadAnimations, line 119: Can you double-check that all instances of `add_material` have been changed to `add_get_material`?

zNightlord: The only place it is used is model spawn add_model() so.

StandingPadAnimations (approved):

TheDuckCow:
Thanks both @StandingPadAnimations and @zNightlord for staying on top here and getting this hot fixes, such is the life of the manual string mapping world we live in.

Issue #653, opened 2025-04-03
Grass overlay is gray and is floating and making holes in mesh


FireDasher:
### Did the issue happen after restarting Blender?

- [x] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [x] I have checked known issues, and my problem is different

### What happened?

When I import the grass block, the side grassy part is gray and I have to fix it manually with the add color node, and the overlay is offset.
And if I put multiple grass blocks next to eachother, the overlay's overlap and z-fight

<img width="670" alt="Image" src="https://github.com/user-attachments/assets/34c1d560-b53e-482d-84cb-12a744b8dcd7" />
<img width="670" alt="Image" src="https://github.com/user-attachments/assets/9ca9ca3f-2a60-4d4c-b546-c5ab84538ba5" />

### What steps are needed to produce the bug?

1) Have MCPrep installed
2) Load spawners
3) Spawn grass block model

### What version of MCprep are you using?

MCPrep 3.6.1.2

### What version of Blender are you using?

Blender 4.4.0

FireDasher:
changing the vec=0.02 to vec=0.001 on line 374 in spawners/mcmodel.py fixes the offset

FireDasher:
unmuting the add color and setting B to #83CE54FF in the grass block overlay material fixes the color

FireDasher:
these need to be fixed by default though

zNightlord:
It is either zfight with itself or zfight with other overlay (grass) block around it since Minecraft has that `cullface` the closest to it is merge 2 texture together.
Inflate offset is the easiest option to deal with that but you supposed not to place it and not going into edit mode.

Putting this here as later reminder
[Desaturate](https://github.com/Moo-Ack-Productions/MCprep/blob/6514aa5b9bf39c1241865c9fbefe8c0d150305c6/MCprep_addon/MCprep_resources/mcprep_data_update.json#L4297)
[grass_block.json](https://github.com/InventivetalentDev/minecraft-assets/blob/1.20.1/assets/minecraft/models/block/grass_block.json)

TheDuckCow:
Marking this as closed for tracking purposes, will be in the next release

Issue #510, opened 2023-12-02
Grass Block texture not working!


LOLOLOL08:
### Did the issue happen after restarting Blender?

- [X] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [X] I have checked known issues, and my problem is different

### What happened?

When Importing a grass block with MCPrep into Blender the top and bottom texture aren´t working as they should.
![Bild_2023-12-02_115833085](https://github.com/Moo-Ack-Productions/MCprep/assets/152694412/5d8a5455-64f4-4bbc-8eaf-b68912e72370)

### What steps are needed to produce the bug?

Import a grass block trough the mcprep block model spawner

### What version of MCprep are you using?

I dont know :(

### What version of Blender are you using?

3.2.2

zNightlord:
Please update your MCprep version to the latest as it was fixed in 3.5.0 #437

LOLOLOL08:
Thanks! It fixed the problem but now it renders the block with a second layer. How can I dissable it so the 2nd layer doesnt apear  or is there a way to delete the 2nd [layer?[](url)](![Bild_2023-12-02_183104755](https://github.com/Moo-Ack-Productions/MCprep/assets/152694412/33f0720b-2793-47b9-a4b9-edfcfb950a3f))

zNightlord:
[...]

Issue #649, opened 2025-02-07
block (model) spawner missing blocks


CrazyMonkeyYT:
### Did the issue happen after restarting Blender?

- [x] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [x] I have checked known issues, and my problem is different

### What happened?

not all default blocks are present in the "block (model) spawner" list 
specifically "hopper" is not there but "hopper_side" is both .json models are present in "MCprep_addon\MCprep_resources\resourcepacks\mcprep_default\assets\minecraft\models\block" and there is no problem importing "hopper" from there 

![Image](https://github.com/user-attachments/assets/239b2671-7620-4627-8003-450df8e11723)

![Image](https://github.com/user-attachments/assets/f442becb-3d07-4a0f-9e46-acdfac958dde)

### What steps are needed to produce the bug?

1) select block (model) spawner
2) scroll to "hopper"
3) not find it...

### What version of MCprep are you using?

MCprep 3.6.1.2

### What version of Blender are you using?

Blender 4.3.2

Aspirata:
Btw hopper has wrong UV

![Image](https://github.com/user-attachments/assets/32926174-b692-4838-acf5-5ae984bbc9cc)

StandingPadAnimations:
Closing this as fixed in #655. Hopefully we can get a new release by the end of the year 

Edit: I also have some builds on both the MCprep server and [my website](https://www.standingpad.org/mcprep-mb-builds/) which includes some recent bug fixes, though note they are unofficial

PR #655, opened 2025-04-07
General model spawner issues, with short grass and grass block

zNightlord:
Previously #643 
~~This should deal with the `short_grass` problem by remap it back to its general `grass` name.~~
The grass block problem for now lowering the offset (it might be start to z fight when zoom out), guess temporally til there is a better proper solution **Maybe with numpy*

Filtering out the blocks supposed not to be in the spawner

Change the material index mapping a little to handle things better no duplicates, on block with remaping `#side` for example logs, pillars
image

Fix for cauldron, hopper, scraffolding. Adjust block uv that are animated have weirder texture ratios (lava, sculk, camp fire, sea lantern,...)

I am quite unsure about the `preprocess_material(mat)` changing the name which cause the problem so changed the name in the base json file.

Fix for #510 #653

Related #267


MCprep_addon/materials/generate.py
```diff
@@ -104,6 +104,8 @@ def get_mc_canonical_name(name: str) -> Tuple[str, Optional[Form]]:
 	if "_stem" in general_name:
 		# For melon/pumpkin stem stage map to only melon_stem/pumpkin_stem
 		general_name = general_name.rsplit("_", 1)[0]
+	if "short_grass" in general_name:
+		general_name = "grass"
```
StandingPadAnimations: I think we actually don't need to do this, at least once we update the `mcprep_data_update.json`. Maybe we'd need to do the opposite.
image

MCprep_addon/spawner/mcmodel.py
```diff
@@ -376,7 +387,7 @@ def add_model(
 			# Give slight offset by normal for overlay geometry
 			if face_mat == "#overlay":
 				bmesh.ops.translate(bm, verts=face.verts,
-						    vec=0.02 * face.normal)
+						    vec=0.005 * face.normal)
```
StandingPadAnimations: I think we should test and see if this causes z-fighting issues

MCprep_addon/spawner/mcmodel.py
```diff
@@ -461,15 +476,35 @@ def update_model_list(context: Context):
 
 		# Filter out models that can't spawn. Typically those that reference
 		# #fire or the likes in the file.
-		if "template" in name:
+		# These blocks just don't make sense to put in the for "unspawnable_for_now"
+		# Template base of that block for example candle, cake with candles
+		# Orient blocks base, cube same as orientable (no texture)
+		# Light blocks are just special no geometry block with 15 states of light levels
+		# Shulkers, Hanging Signs, Signs are entities, put it here for now since they have a lot of variants
+		# pitcher_crop_top_stage_1 is a top part of a double plant. I have no idea why it have no geometry.
```
StandingPadAnimations: Formatting comment: I'd add a `-` in front of each item to make a bulleted list, just so it's easier to read

MCprep_addon/spawner/mcmodel.py
```diff
@@ -461,15 +476,35 @@ def update_model_list(context: Context):
...
+			light_0| light_1| \
+			pitcher_crop_top| custom_fence_",
+			name
+		)
+		if is_contains:
+			continue
+		# Single word condition filter
+		# block single block parent, base parent of most MC contain gui displays
+		# Air, Barrier, Structure void has no geometry
+		if name in ["block", "air", "barrier", "structure_void", "thin_block"]:
```
StandingPadAnimations: Could we perhaps combine this into a single regex statement? Maybe have some string construction that takes these names and adds them to the regex string? That way, we'd only need one conditional statement

StandingPadAnimations (changes requested):
Just some small changes needed

zNightlord:
Hopper and Cauldron and its variation blocks now have their dedicated uv data stored in resources json
Fixing the stretched UV of spawned blocks related to #646 

https://github.com/user-attachments/assets/dff67a1a-7ea4-4256-8e4c-71e3771c79e9

MCprep_addon/materials/generate.py
StandingPadAnimations: @zNightlord can you test with this `mcprep_data.json` file and see if this section would still work with OBJs from older worlds? Just make sure you don't commit the updated JSON to the branch.

[mcprep_data.json](https://github.com/user-attachments/files/19697457/mcprep_data.json)

zNightlord: added the desaturated for "short_grass" in the base json and removed the above if.
Yeah it doesn't work since old worlds use "grass" and that caused by "grass": "colormap/grass" in the json. I have to manually restore deleted json and reload change it to work, I guess have to fix in [mcprep_data_refresh.py](https://github.com/Moo-Ack-Productions/MCprep/blob/6514aa5b9bf39c1241865c9fbefe8c0d150305c6/mcprep_data_refresh.py#L458) add it in the overrides.

MCprep_addon/spawner/mcmodel.py
zNightlord: It's already quite small gap.
image

MCprep_addon/materials/generate.py
StandingPadAnimations: I think a better route then would just be to ignore the change, or check if `short_grass` is already in the JSON. IMO we should avoid removing texture names from the JSON file as much as possible

EDIT: I see you've already removed the conditional, ignore this comment

MCprep_addon/spawner/mcmodel.py
StandingPadAnimations: Yeah as you noted in Discord it also causes z-fighting at far distances. Combining textures with Numpy might be the best option then

zNightlord:
Added override for "grass" since the get_vanilla_list map it to colormap/grass texture which causing issue
This is the [colormap/grass texture]
(https://github.com/InventivetalentDev/minecraft-assets/blob/1.21.5/assets/minecraft/textures/colormap/grass.png)

Edit: Numpy grass texture maybe it's better to save it for later.

MCprep_addon/materials/generate.py
```diff
@@ -101,8 +101,9 @@ def get_mc_canonical_name(name: str) -> Tuple[str, Optional[Form]]:
 		# mixed up with the new "water": "painting/water" texture.
 		general_name = "water_still"
 
-	if "_stem" in general_name:
+	if "_stem" in general_name and general_name not in ["pumpkin_stem", "melon_stem", "attached_pumpkin_stem", "attached_melon_stem"]:
```
StandingPadAnimations: I'd make this list a tuple

MCprep_addon/spawner/mcmodel.py
```diff
@@ -350,11 +360,24 @@ def add_model(
...
+					_uv_from = _uv_from.get(f"from,{e['from']}", {"all": [0, 0, 16, 16]})
+					uv_coords = _uv_from.get(face_dir[i], [0, 0, 16, 16])
+				elif "cauldron" in obj_name:
+					_uv_from = uv_from.get("cauldron")
+					_uv_from = _uv_from.get(f"from,{e['from']}", {"all": [0, 0, 16, 16]})
+					uv_coords = _uv_from.get(face_dir[i], [0, 0, 16, 16])
+				elif "scaffolding" in obj_name:
+					_uv_from = uv_from.get("scaffolding")
+					_uv_from = _uv_from.get(f"from,{e['from']}", {"all": [0, 0, 16, 16]})
+					uv_coords = _uv_from.get(face_dir[i], [0, 0, 16, 16])
```
StandingPadAnimations: This could probably be consolidated into something like this:
```py
				elif any(x in obj_name for x in ("hopper", "cauldron", "scaffolding")):
					name = next((x for x in ("hopper", "cauldron", "scaffolding") if x in obj_name), False)
					_uv_from = uv_from.get(name)
					_uv_from = _uv_from.get(f"from,{e['from']}", {"all": [0, 0, 16, 16]})
					uv_coords = _uv_from.get(face_dir[i], [0, 0, 16, 16])
```

though a bit nightmarish, and I would prefer this syntax if we were using Python 3.8 (which unfortunately isn't available in Blender 2.83)
```py
				elif (name := next((x for x in ("hopper", "cauldron", "scaffolding") if x in obj_name), False)) is not False:
					_uv_from = uv_from.get(name)
					_uv_from = _uv_from.get(f"from,{e['from']}", {"all": [0, 0, 16, 16]})
					uv_coords = _uv_from.get(face_dir[i], [0, 0, 16, 16])
```

MCprep_addon/spawner/mcmodel.py
```diff
@@ -372,22 +395,57 @@ def add_model(
...
 				# will be 2 then 3, 0, 1.
 				face.loops[j][uv_layer].uv = uvs[(j + uv_idx) % len(uvs)]
 
-			# Assign the material on face
-			if face_mat is not None and face_mat in materials:
-				face.material_index = materials.index(face_mat)
+			# Using materials_remap to remap the index, used for the block with remapping "#side"
+			# Stored material index for getting the texture
+			material_index = 0
+			if face_mat is not None and (face_mat in materials or face_mat in materials_remap):
```
StandingPadAnimations: This could be reduced to the following
```suggestion
			if face_mat and (face_mat in materials or face_mat in materials_remap):
```

MCprep_addon/spawner/mcmodel.py
```diff
@@ -372,22 +395,57 @@ def add_model(
...
+				face_mat_ref = materials_remap.get(face_mat)
+				if face_mat_ref is not None and "#" in face_mat_ref:
```
StandingPadAnimations: Same here
```suggestion
				if face_mat_ref and "#" in face_mat_ref:
```

StandingPadAnimations (changes requested):
Looks mostly fine from a quick skim, some stylistic changes are needed though

zNightlord:
Add composter UV mentioned in #658 


https://github.com/user-attachments/assets/50cfc9f8-3832-46fb-9663-ad77e1db5d08

MCprep_addon/spawner/mcmodel.py
```diff
@@ -350,11 +360,22 @@ def add_model(
 				# Cake and cake slices don't store the UV keys
 				# in the JSON model, which causes issues. This
 				# workaround this fixes those texture issues
+				uv_from = env.json_data.get("block_model_uv")
+				stored_uv_blocks = ("hopper", "cauldron", "scaffolding", "composter")
```
StandingPadAnimations, line 364: I assume this is to fix issues such as those mentioned in https://github.com/Moo-Ack-Productions/MCprep/issues/649#issuecomment-2642253946?

StandingPadAnimations (commented):
Looks good, and things work on my end pretty nicely. I just need to verify that this will work with an updated resource pack like 1.21.5 before I merge this in

StandingPadAnimations:
I'll also merge `dev` into this branch to handle the merge conflicts

StandingPadAnimations (approved):

StandingPadAnimations:
Merging this PR into dev

TheDuckCow:
Thanks for reviewing @StandingPadAnimations before I could get to it, was also going to grant my approval. Thanks @zNightlord for the continued contributions!

PR #656, opened 2025-04-07
feat: Add ability for users to locally refresh MCprep Resources

StandingPadAnimations:
Sometimes, there are circumstances where a user may want to update the MCprep resources folder locally without having to wait for an MCprep update. Although we plan to eventually implement separate updates for assets, this adds the ability to locally refresh textures based on the user's locally installed version of Minecraft.

In theory this can also be used as part of the move towards not bundling Minecraft textures, which A. are a legal gray area and B. could be a GPL violation (at least in discussions I had with the FSF, their logic being that Minecraft is a proprietary game, though I'd argue this is more of a gray area since we're dealing with non-code assets).

This also updates push_latest.sh to use the now moved script


MCprep_addon/mcprep_data_refresh.py
```diff
@@ -470,14 +470,14 @@ def get_current_json(backup=False):
...
-		"MCprep_addon", "MCprep_resources",
+		PARENT_PATH, "MCprep_resources",
 		"mcprep_data_update"+suffix+".json")
 	return filepath
 
 
 def read_base_mapping():
 	"""Read in the existing mcprep_data_update.json file shipped with MCprep"""
-	filepath = "mcprep_data_base.json"
+	filepath = PARENT_PATH + "/MCprep_resources/" + "UpdateJson/" + "mcprep_data_base.json"
```
TheDuckCow: Probably want to ensure this remains cross compatible using os.join or the path lib

MCprep_addon/mcprep_data_refresh.py
```diff
@@ -575,7 +575,7 @@ def run_all(auto=False):
 	# Goal: generate, best we can, the actual mapping file to use for materials
 
 	# save the output
-	fileout = "mcprep_data_update_staging.json"
+	fileout = PARENT_PATH + "/MCprep_resources/" + "UpdateJson/" + "mcprep_data_update_staging.json"
```
TheDuckCow: Same here

MCprep_addon/mcprep_ui.py
TheDuckCow, line 1: UI comment: the label gets truncated

Screen Shot 2025-06-07 at 11 00 32 AM

Maybe we can do something like this: "Minecraft path" for the first label instead of "Path to..." part.
Then, perhaps a label above, some further instruction on how to properly pick the right path.

And then more than anything, I'd love for us to try and provide the default location and populate that as a visible value (if it exists). Seems like it should be in a consistent place on your windows or mac machine right? I can cross check on a windows machine if you don't have access to one.

We could put a warning icon below if the default path isn't found, so they can still reference the original folder structure to know what to look for if they have a custom install locaiton.

If we can do that, it will make this substantially easier to use, and we could even potentially auto detect _when_ to perform the action if we can see more recently modified images than the last time locally refreshed for instance (maybe asking the user if they want to refresh installed asset data).

MCprep_addon/mcprep_data_refresh.py
TheDuckCow, line 1: I got a weird error when trying the operation while having an empty path specified (which as I read it *should* just use the default path I was using before). Didn't troubleshoot further yet, but curious if you have an idea.

```
('23w12a', '23w14a', '23w51b', 'fabric-loader-0.14.19-1.19.3', 'fabric-loader-0.14.19-1.19.4', '1.18-rc3', '1.19-forge-41.1.0', '1.21-pre2', '1.16.5-forge-36.1.0', '1.16.5', '1.17', '1.17.1-pre1', '1.17.1', '1.18.1', '1.18.2', '1.19', '1.19.1-rc2', '1.19.2-rc1', '1.19.4-rc2', '1.19.1', '1.19.2', '1.19.3', '1.19.4', '1.19.5', '1.20.6-rc1', '1.20.2', '1.20.4', '1.20.5', '1.20.6', '1.21', '1.21.6-pre3', '1.21.5')
Extracting from jar /Users/patrickcrawford/Library/Application Support/minecraft/versions/1.21.5/1.21.5.jar
Got jar version: 1.21.5.jar
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/util_operators.py", line 235, in execute
    mcprep_data_refresh.run_all(auto=True, versions_path=bpy.path.abspath(addon_prefs.minecraft_versions_path))
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/mcprep_data_refresh.py", line 520, in run_all
    jmc = get_jmc2obj_mapping()  # Legacy, most jmc2obj mats are now live-parsed.
          ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/mcprep_data_refresh.py", line 50, in get_jmc2obj_mapping
    raw_str = save_file_str(JMC_1_13)
              ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/mcprep_data_refresh.py", line 27, in save_file_str
    result = urllib.request.urlopen(request, timeout=10)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 216, in urlopen
    return opener.open(url, data, timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 525, in open
    response = meth(req, response)
               ^^^^^^^^^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 634, in http_response
    response = self.parent.error(
               ^^^^^^^^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 563, in error
    return self._call_chain(*args)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 496, in _call_chain
    result = func(*args)
             ^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 643, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 429: Too Many Requests
Error: Python: Traceback (most recent call last):
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/util_operators.py", line 235, in execute
    mcprep_data_refresh.run_all(auto=True, versions_path=bpy.path.abspath(addon_prefs.minecraft_versions_path))
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/mcprep_data_refresh.py", line 520, in run_all
    jmc = get_jmc2obj_mapping()  # Legacy, most jmc2obj mats are now live-parsed.
          ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/mcprep_data_refresh.py", line 50, in get_jmc2obj_mapping
    raw_str = save_file_str(JMC_1_13)
              ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/mcprep_data_refresh.py", line 27, in save_file_str
    result = urllib.request.urlopen(request, timeout=10)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 216, in urlopen
    return opener.open(url, data, timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 525, in open
    response = meth(req, response)
               ^^^^^^^^^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 634, in http_response
    response = self.parent.error(
               ^^^^^^^^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 563, in error
    return self._call_chain(*args)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 496, in _call_chain
    result = func(*args)
             ^^^^^^^^^^^
  File "/Applications/Blender 4.3/blender.app/Contents/Resources/4.3/python/lib/python3.11/urllib/request.py", line 643, in http_error_default
    raise HTTPError(req.full_url, code, msg, hdrs, fp)
urllib.error.HTTPError: HTTP Error 429: Too Many Requests
```

TheDuckCow (commented):
Sorry for the crazy long delay here but I do really like the idea of this, especially given our release cadence. A few initial thoughts, but this would be well worth cutting a release for as soon as it's ready.

MCprep_addon/mcprep_ui.py
StandingPadAnimations: ~~I've attempted to work on default folders and... let's just say Blender has some weird issues (tried functions, f-strings, getters and setters, it's really weird why none of those work properly), specifically if the user then changes the path (internally, the property still uses the "default"). I'll keep the function just in case anyone wants to give a crack at it, but for now, the default will remain `''`.~~

EDIT: Ignore the above, I had the wrong property name for the UI...

I will however implement the warning icon, as well as gray out the button if the path is invalid.

MCprep_addon/mcprep_data_refresh.py
StandingPadAnimations: Weird, though now inputting empty paths will just gray out the operator, so I'm not too concerned about this

StandingPadAnimations:
Turns out this didn't actually copy textures over, so I made a slight change to actually make MCprep perform that. Testing on my end with a texture-less build, it seems to work (though weirdly when I set a path in preferences, a `versions` subdir is added automatically, can you check on your end and see if that happens?).

@TheDuckCow For the next release, should we remove textures from the MCprep assets, as now users can just provide a copy of the Minecraft JAR?

StandingPadAnimations:
Maybe we could extend this to textures for rigs as well, but those are a bit more complicated to work with for now (although I think modern versions of Blender have a way to handle bulk file changes through scripting, but I haven't looked into that).

Edit: it's a separate open source utility by Blender Studio, [`bbatch`](https://projects.blender.org/studio/blender-studio-tools/src/branch/main/scripts/bbatch). I'll see if I can implement the rig part as well (though that's not a blocker and can be made as a separate PR if needed)

StandingPadAnimations:
Alternatively (and this might be simpler), what we could do is use unpacked textures for the rigs to the paths that would exist if the user had the resource pack downloaded (as opposed to packing them in rig files), and then have MCprep relink the texture when appending

TheDuckCow:
@StandingPadAnimations  back from a driving trip (and a busy work week)

> Edit: it's a separate open source utility by Blender Studio, [bbatch](https://projects.blender.org/studio/blender-studio-tools/src/branch/main/scripts/bbatch). I'll see if I can implement the rig part as well (though that's not a blocker and can be made as a separate PR if needed)

Sounds a little complex for right now is my immediate thought, at the very least would be worth adding as a secondary PR if we think it's worth it. 

> @TheDuckCow For the next release, should we remove textures from the MCprep assets, as now users can just provide a copy of the Minecraft JAR?

I think we should at least have one release out to ensure things are stable, and then see how it goes then.

MCprep_addon/util_operators.py
```diff
@@ -225,6 +225,15 @@ def execute(self, context):
 			return {'CANCELLED'}
 		return {'FINISHED'}
 
+class MCPREP_OT_local_refresh(bpy.types.Operator):
+	bl_idname = "mcprep.refresh_data"
+	bl_label = "Locally refresh MCprep data"
```
TheDuckCow: Thinking about it further, maybe this would be more intuitive of a name? Mostly because it clearly articulates we're really pulling data from somewhere and not just some in-memory isolated reloading, and that we are actually pulling resources not just data (resources will also resonate more with users, being used to resource packs, thus self-describing a bit better). You could make it a bit more concise by calling it `mcprep.extract_mc_resources`.

```suggestion
class MCPREP_OT_extract_minecraft_resources(bpy.types.Operator):
	bl_idname = "mcprep.extract_minecraft_resources"
	bl_label = "Extract resources from Minecraft"
```

MCprep_addon/mcprep_ui.py
```diff
@@ -609,6 +627,28 @@ def draw(self, context):
...
+			split = util.layout_split(box, factor=factor_width)
+			col = split.column()
+			col.label(text=env._("Path to Minecraft versions folder"))
+			col = split.column()
+			col.prop(self, "save_folder", text="")
+			split = util.layout_split(box, factor=factor_width)
+			col = split.column()
+			col.label(text=env._("Refresh"))
+			col = split.column()
+			col.operator("mcprep.refresh_data", text=env._("Locally Refresh MCprep JSON data"))
```
TheDuckCow: We can simplify this section by just using the operator name itself I'd say

```suggestion
			col.operator("mcprep.refresh_data")
```

TheDuckCow: Also since we have the functionality to define a default path, why not add another button operator that resets the path? similar to here:

Screen Shot 2025-06-21 at 10 50 07 AM

MCprep_addon/mcprep_ui.py
```diff
@@ -609,6 +627,28 @@ def draw(self, context):
 			col = split.column()
 			p = col.operator("mcprep.openfolder", text=env._("Open texture pack folder"))
 			p.folder = self.custom_texturepack_path
+			
+			row = layout.row()
+			row.scale_y = 0.7
+			row.label(text=env._("Refresh MCprep data with local Minecraft install"))
```
TheDuckCow:
```suggestion
			row.label(text=env._("Extract data/resources from local Minecraft install"))
```

MCprep_addon/util_operators.py
```diff
@@ -225,6 +225,15 @@ def execute(self, context):
...
+	
+	def execute(self, context):
+		from . import mcprep_data_refresh
+		addon_prefs = util.get_user_preferences(context)
+		mcprep_data_refresh.run_all(auto=True, versions_path=bpy.path.abspath(addon_prefs.save_folder), copy_file=True)
```
TheDuckCow: I noticed it forces users to restart after. It's a rare thing to do so I'm not too concerned, but is there a reason it's strictly necessary?

Screen Shot 2025-06-21 at 10 53 24 AM

MCprep_addon/mcprep_ui.py
```diff
@@ -609,6 +627,28 @@ def draw(self, context):
...
+			split = util.layout_split(box, factor=factor_width)
+			col = split.column()
+			col.label(text=env._("Refresh"))
+			col = split.column()
+			col.operator("mcprep.refresh_data", text=env._("Locally Refresh MCprep JSON data"))
+			col.enabled = False
+			
+			if not os.path.isdir(bpy.path.abspath(self.save_folder)):
+				row = box.row()
+				row.label(text=env._("Saves folder not found"), icon="ERROR")
```
TheDuckCow: Reads a bit more consistently with above changes.
```suggestion
				row.label(text=env._("Versions folder not found"), icon="ERROR")
```

MCprep_addon/mcprep_ui.py
```diff
@@ -609,6 +627,28 @@ def draw(self, context):
...
 			p = col.operator("mcprep.openfolder", text=env._("Open texture pack folder"))
 			p.folder = self.custom_texturepack_path
+			
+			row = layout.row()
+			row.scale_y = 0.7
+			row.label(text=env._("Refresh MCprep data with local Minecraft install"))
+			box = layout.box()
+			split = util.layout_split(box, factor=factor_width)
+			col = split.column()
+			col.label(text=env._("Path to Minecraft versions folder"))
```
TheDuckCow: Screen Shot 2025-06-21 at 10 55 27 AM

Gets truncated anyways, I think this will fit:

```suggestion
			col.label(text=env._("Minecraft versions"))
```

MCprep_addon/mcprep_ui.py
```diff
@@ -461,13 +478,14 @@ def update_exp_vivy_file_path(self, context):
...
 		default="Mineways")
+
 	save_folder: bpy.props.StringProperty(
 		name="MC saves folder",
 		description=(
 			"Folder containing Minecraft world saves directories, "
 			"for the direct import bridge"),
 		subtype='FILE_PATH',
-		default='')
+		default=save_path_default())
```
TheDuckCow: Per below, would be great to connect the save path default to its own operator too for quick resetting.

MCprep_addon/util_operators.py
```diff
@@ -225,6 +225,15 @@ def execute(self, context):
 			return {'CANCELLED'}
 		return {'FINISHED'}
 
+class MCPREP_OT_local_refresh(bpy.types.Operator):
+	bl_idname = "mcprep.refresh_data"
+	bl_label = "Locally refresh MCprep data"
+	
+	def execute(self, context):
```
TheDuckCow, line 233: Idea: maybe better for a follow up PR, since I'd really like to merge this. But if we want to get to a point where users have to load their minecraft resources, then we may as go the rest of the way of doing it basically automatically or at least prompting it, both the first time when there are no resources, but also when there are _new_ resources to update.

So my thought is that we could have this operator save or update some metadata as to which version of mc versions it pulled the data from. Then in theory, we could have another startup function which checks the json file vs the versions folder (if path is valid) to just detect if the most recent version to extract from is more recent than the one last extracted. And if so, show some button in the main MCprep n panel UI to generate it.

Let me know your thoughts, again suggesting either way it should be a follow up PR (but a good one, otherwise 95% of users won't even know about this new functionality)

MCprep_addon/mcprep_ui.py
```diff
@@ -406,6 +419,10 @@ def update_exp_vivy_file_path(self, context):
 			"with material prepping"),
 		subtype='DIR_PATH',
 		default=f"{scriptdir}/MCprep_resources/resourcepacks/mcprep_default/")
+	minecraft_versions_path: bpy.props.StringProperty(
+		name="Minecraft Versions Path",
+		description="Path containing the installed version of Minecraft",
```
TheDuckCow:
```suggestion
		description="Path to the installed Minecraft versions folder",
```

TheDuckCow (approved):
Last round of a few comments, but I trust you can make the remaining changes and then merge once ready. My nitpicks here are mostly around making sure the feature reads clearly in the UI and avoiding some truncation. Happy to chat if you disagree with any of the points

MCprep_addon/mcprep_ui.py
StandingPadAnimations: Just realized I somehow mixed up the old `save_folder` property for the import bridge with this

MCprep_addon/util_operators.py
StandingPadAnimations: This automatically occurs, since I believe the code that does this for updates just checks if `MCprep_resources` directory has changed at all

MCprep_addon/util_operators.py
StandingPadAnimations: I think that would be pretty useful to implement. I'll save the bulk of it for a follow-up PR, but while I'll add a basic version output file to `mcprep_data_refresh.py` that we can build on.

StandingPadAnimations:
I'll go ahead and merge this now

PR #662, opened 2025-06-12
fix: Only check for scaled UVs if the OBJ does not have CommonMCOBJ data

### LINKED ISSUE #661: Cannot swap texture pack using jmc2obj

- [x] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [x] I have checked known issues, and my problem is different

### What happened?

When using jmc2obj, trying to change a texture pack with normals/specular errors out with "Detected scaled UV's (all in one texture), be sure to use Mineway's 'Export Individual Textures To..'' feature".
This happens when exporting all textures + specular/normal, all textures, and no textures (and also + alpha/textures in separate files). Well, every possible combination.

When using one object only the texture pack does not apply. When using objects per material, the **(default)** texture pack, without normals/specular textures applying either.

1. Generate one of the few combinations tested in jmc2obj
2. Open new blender file
3. Import obj in the jmc2obj tab.
4. Immediately press swap texture pack, selecting either Specular or SEUS.
5. Plain (default) textures are applied, but not specular/normal data.

3.6.1.2

4.2.3 LTS

StandingPadAnimations:
(Doesn't resolve #661 completely, but it does reduce confusion for this case)

This PR (in addition to adding a helper function to check for the existance of CommonMCOBJ data) changes the scaled UVs check so it only runs if the OBJ does not have CommonMCOBJ data, since `texture_type` is more reliable than the scaled UVs check anyway.

Maybe in the future we could compare the resource pack to the `texture_type` value and make it smarter, but for now, this simple fix will do.

### DIFF
--- a/MCprep_addon/materials/prep.py
+++ b/MCprep_addon/materials/prep.py
@@ -489,7 +489,12 @@ class MCPREP_OT_swap_texture_pack(
 			self.report({'ERROR'}, "No materials found on selected objects")
 			return {'CANCELLED'}
 		_ = generate.detect_form(mat_list)
-		invalid_uv, affected_objs = uv_tools.detect_invalid_uvs_from_objs(obj_list)
+		
+		# If we have a commonmcobj header, we don't need 
+		# to check for scaled UVs
+		invalid_uv, affected_objs = (False, [])
+		if not world_tools.is_commonmc_obj(context):
+			invalid_uv, affected_objs = uv_tools.detect_invalid_uvs_from_objs(obj_list)
 
 		# NOTE: This is temporary
 		addon_prefs = util.get_user_preferences(context)
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -151,6 +151,12 @@ EXPORTER_MAPPING = {
 
 UNSUPPORTED_OR_NONE = (WorldExporter.Unknown, None)
 
+def is_commonmc_obj(context: Context) -> bool:
+	obj = context.object
+	if obj and "COMMONMCOBJ_HEADER" in obj:
+		return True
+	return False
+
 def get_exporter(context: Context) -> Optional[WorldExporter]:
 	"""
 	Return the exporter on the active object if it has


TheDuckCow (approved):
Change seems fine to me as well, thought it is also a slight head scratcher that it's not being picked up originally the correct way. But the fact that it's just some specific objects like stems that throw it off make me think we just have brittle logic in place for the uv checks, so this solution is better all around anyways to rely on the metadata. Kudos once again for getting that into mineways and jmc2obj

MCprep_addon/materials/prep.py
```diff
@@ -489,7 +489,12 @@ def execute(self, context):
 			self.report({'ERROR'}, "No materials found on selected objects")
 			return {'CANCELLED'}
 		_ = generate.detect_form(mat_list)
-		invalid_uv, affected_objs = uv_tools.detect_invalid_uvs_from_objs(obj_list)
+		
+		# If we have a commonmcobj header, we don't need 
+		# to check for scaled UVs
```
TheDuckCow, line 494: Actually, is that right? shouldn't it be more that we know the answer, not that it's automatically false? (I didn't live test the before / after)

StandingPadAnimations: Assuming the scaled UVs check is just for determining the use of an atlas vs individual textures, the CommonMCOBJ header would let us know if the OBJ uses an atlas or individual textures

StandingPadAnimations:
> Change seems fine to me as well, thought it is also a slight head scratcher that it's not being picked up originally the correct way. But the fact that it's just some specific objects like stems that throw it off make me think we just have brittle logic in place for the uv checks, so this solution is better all around anyways to rely on the metadata.

To be honest, I'm partially on the fence that it's something with the reporter's specific OBJ's geometry, since I wasn't able to reproduce it with any new exports I made with the same settings, and was only able to reproduce with the same OBJ. I would need to do further testing to determine if indeed that is the case though.

MCprep_addon/materials/prep.py
TheDuckCow: Ah perfect, it happens downstream - then allllll goood 👍

TheDuckCow:
At this stage if this resolves their issue, I personally won't invest that much extra time testing. If it comes up again, then we could consider.

PR #663, opened 2025-06-12
build(rc): Make download_resources.sh restore Languages and JSON

StandingPadAnimations:
To avoid issues when building MCprep (and for convinience reasons), the
download_resources.sh script has been updated to restore the `Languages`
and `mcprep_data_update.json` files after introducing the other assets.

### DIFF
--- a/rc-files/download_resources.sh
+++ b/rc-files/download_resources.sh
@@ -2,7 +2,7 @@
 # 
 # TODO: Automatically get the latest tag and download link. Perhaps 
 # through the GitHub API?
-wget "https://github.com/Moo-Ack-Productions/MCprep/releases/download/3.5.3/MCprep_addon_3.5.3.zip"
+wget "https://github.com/Moo-Ack-Productions/MCprep/releases/download/3.6.1.2/MCprep_addon_3.6.1.2.zip"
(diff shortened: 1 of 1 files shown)


rc-files/download_resources.sh
```diff
@@ -2,7 +2,7 @@
 # 
 # TODO: Automatically get the latest tag and download link. Perhaps 
 # through the GitHub API?
-wget "https://github.com/Moo-Ack-Productions/MCprep/releases/download/3.5.3/MCprep_addon_3.5.3.zip"
+wget "https://github.com/Moo-Ack-Productions/MCprep/releases/download/3.6.1.2/MCprep_addon_3.6.1.2.zip"
```
TheDuckCow: Small suggestion to avoid future mistakes, maybe create a bash var for the version?

TheDuckCow (approved):
Presuming it works, didn't live test - but I guess we'd quickly realize if there's any issue

rc-files/download_resources.sh
StandingPadAnimations: Yeah I would agree the version should be split out into its own variable. Maybe in the future we could use GitHub's API to get the latest release.

PR #669, opened 2025-06-28
refactor: Move custom node properties to PropertyGroup

### LINKED ISSUE #668: Blender 5.0 breaks the `generate` function in MCprep

- [x] Yes, the issue indeed continued after restarting Blender

### Check known issues

- [x] I have checked known issues, and my problem is different

### What happened?

In Blender 5.0, when spawning rigs or blocks, or prepping materials, the following errors are thrown

- Block spawning:
```
File "/home/mahid/.config/blender/5.0/scripts/addons/MCprep_addon/materials/generate.py", line 1221, in generate_base_material
    node_diff["MCPREP_diffuse"] = True
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^
TypeError: bpy_struct[key] = val: id properties not supported for this type
```
- Rig spawning (though the spawner seems to be able to gracefully handle this):
```
File "/home/mahid/.config/blender/5.0/scripts/addons/MCprep_addon/materials/generate.py", line 564, in get_textures
    elif "MCPREP_diffuse" in node:
         ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: bpy_struct: this type doesn't support IDProperties
```
- Prep materials:
```
File "/home/mahid/.config/blender/5.0/scripts/addons/MCprep_addon/materials/generate.py", line 564, in get_textures
    elif "MCPREP_diffuse" in node:
         ^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: bpy_struct: this type doesn't support IDProperties
```

This is due to [a recent change](https://projects.blender.org/blender/blender/commit/7276b2009a8819619263a1b897389662307b0ee0) in Blender in how properties can be accessed. To quote the [release notes](https://developer.blender.org/docs/release_notes/5.0/python_api/#removal-of-unsuported-access-to-runtime-defined-properties-storage-data)
> Removal of (Unsuported) Access to Runtime-Defined Properties Storage Data[¶](https://developer.blender.org/docs/release_notes/5.0/python_api/#removal-of-unsuported-access-to-runtime-defined-properties-storage-data)
> 
> Properties defined by the [`bpy.props`](https://docs.blender.org/api/5.0/bpy.props.html) API are no longer stored in the same container as user-defined [Custom Properties](https://docs.blender.org/manual/en/5.0/files/custom_properties.html). As a consequence, it is no more possible to access them directly through the Python 'dict-like' syntax.
>
> E.g. `bpy.context.scene['cycles']` will not give access to Cycles' scene settings.

Based on what I can tell in the [accompanying issue thread](https://projects.blender.org/blender/blender/issues/123232), the new changes in Blender 5.0 massively restrict what addons can modify at runtime. I've yet to do a deep dive on how this impacts us specifically, but we should be expecting some massive refactoring at the very least.

1. Run Blender 5.0
2. Use either the block spawner, rig spawner, or prep materials functions

`3.7-MB-3`, which is everything currently on the `dev` branch (though I would assume this impacts the latest stable release)

Blender 5.0

StandingPadAnimations:
Closes #668

Previously, for the last 7 years, MCprep added custom properties to nodes directly. However, in Blender 5.0, this was changed (see https://projects.blender.org/blender/blender/commit/7276b2009a8819619263a1b897389662307b0ee0), causing MCprep to break when using the `generate` function.

To fix this, this patch moves those custom properties to a `PropertyGroup` called `MCprepNodeProps`, which is then registered and added as part of `bpy.types.Node`. Apparently this is the intended way for custom properties to be added to nodes, so it should work in previous versions of Blender.

As part of backwards compatibility with old files, the `np_is_mcprep_node_prop` function was added to `util`, which performs 3 steps:

- Check if the old properties exist on the node
- If old property exists, set the new equivilent
- Return new equivilent of property

When ran, this function will act as both a compatibility layer and migration function.

Credit goes to [snzhongcheng](https://github.com/snzhongcheng) on the MCprep Discord server for suggesting this solution, and [BradyAJonston](https://github.com/BradyAJohnston) from the Erindale server for confirming this is the intended way of adding custom properties to nodes (and thus shouldn't break in future releases of Blender)

### DIFF
--- a/MCprep_addon/materials/generate.py
+++ b/MCprep_addon/materials/generate.py
@@ -34,6 +34,12 @@ AnimatedTex = Dict[str, int]
 NO_DIFFUSE_NODE = 1
 IMG_MISSING = 2
 
+MCPREP_DIFFUSE = "MCPREP_diffuse"
+MCPREP_SPECULAR = "MCPREP_specular"
+MCPREP_NORMAL = "MCPREP_normal"
+MCPREP_DISPLACE = "MCPREP_displace"
+SATURATE = "SATURATE"
+
 
 class PackFormat(Enum):
 	SIMPLE = 0
@@ -433,7 +439,7 @@ def set_cycles_texture(
 		is_grayscale = is_image_grayscale(image)
 
 	for node in material.node_tree.nodes:
-		if node.type == "MIX_RGB" and util.np_is_mcprep_node_prop(node, "SATURATE"):
+		if node.type == "MIX_RGB" and util.np_is_mcprep_node_prop(node, SATURATE):
 			node.mute = not is_grayscale
 			node.hide = not is_grayscale
 			env.log(" mix_rgb to saturate texture")
@@ -442,11 +448,11 @@ def set_cycles_texture(
 		# saved as an attribute on the node
 		if node.type != 'TEX_IMAGE':
 			continue
-		elif util.np_is_mcprep_node_prop(node, "MCPREP_diffuse"):
+		elif util.np_is_mcprep_node_prop(node, MCPREP_DIFFUSE):
 			node.image = image
 			node.mute = False
 			node.hide = False
-		elif util.np_is_mcprep_node_prop(node, "MCPREP_normal"):
+		elif util.np_is_mcprep_node_prop(node, MCPREP_NORMAL):
 			if "normal" in img_sets:
 				new_img = util.loadTexture(img_sets["normal"])
 				node.image = new_img
@@ -464,7 +470,7 @@ def set_cycles_texture(
 				# normal_map = node.outputs[0].links[0].to_node
 				# principled = ...
 
-		elif util.np_is_mcprep_node_prop(node, "MCPREP_specular"):
+		elif util.np_is_mcprep_node_prop(node, MCPREP_SPECULAR):
 			if "specular" in img_sets:
 				new_img = util.loadTexture(img_sets["specular"])
 				node.image = new_img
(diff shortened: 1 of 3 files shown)


MCprep_addon/materials/generate.py
```diff
@@ -442,11 +442,11 @@ def set_cycles_texture(
 		# saved as an attribute on the node
 		if node.type != 'TEX_IMAGE':
 			continue
-		elif "MCPREP_diffuse" in node:
+		elif util.np_is_mcprep_node_prop(node, "MCPREP_diffuse"):
```
TheDuckCow: Minor thought - as long as we are touching each of these lines anyways, maybe worth defining these labels as constants at the top of the file, since we re-use them a few times? Could just be `MCPREP_diffuse = "MCPREP_diffuse"`, or `SATURATE = "SATURATE"`, saving a small chance of inconsistent typos in the future.

MCprep_addon/util.py
```diff
@@ -84,6 +84,38 @@ def is_vivy_enabled(context) -> bool:
 	addon_prefs = get_user_preferences(context)
 	return is_experimental(context) and addon_prefs.exp_vivy_material_system
 
+# TODO: Use Literal["MCPREP_diffuse", "MCPREP_specular",
```
TheDuckCow, line 87: Could we also use Enum.value instead? I think that's already available, and would better work with autocompletion in more editors if I understand it right.

TheDuckCow (approved):
Only minor suggestions for this one, tests passing and things working fine on my side in Blender 4.5 in a live test, plus automated ones below:

```
(4.5.3)   	all_tests	70	2	0	No errors
(4.3.1)   	all_tests	70	2	0	No errors
Compiled in 37.8s + tests ran in 283.4s
Total of 321.2s with exit code 0
```

StandingPadAnimations:
Merging this now

PR #670, opened 2025-06-28
fix: Check if Vivy is enabled before creating JSON file

StandingPadAnimations:
Small fix to make sure the Vivy JSON isn't created if the user didn't enable it.

### DIFF
--- a/MCprep_addon/mcprep_ui.py
+++ b/MCprep_addon/mcprep_ui.py
@@ -380,6 +380,9 @@ class McprepPreference(bpy.types.AddonPreferences):
 		env.reload_vivy_json(path)
 		env.vivy_name_changes = {}
 		env.vivy_cache = None
+	
+	def update_enable_vivy(self, context):
+		env.vivy_enabled = self.exp_vivy_material_system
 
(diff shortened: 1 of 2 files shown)


MCprep_addon/mcprep_ui.py
```diff
@@ -380,6 +380,9 @@ def update_exp_vivy_file_path(self, context):
 		env.reload_vivy_json(path)
 		env.vivy_name_changes = {}
 		env.vivy_cache = None
+	
+	def update_enable_vivy(self, context):
+		env.vivy_enabled = self.exp_vivy_material_system
```
TheDuckCow, line 385: Just checking, this doesn't need to save any values out to work properly on next restart of Blender does it?

TheDuckCow (approved):
Changes themselves look fine, I trust the change to work effectively

MCprep_addon/mcprep_ui.py
StandingPadAnimations: Not sure, added an additional setting of the variable in `register` to make sure it's set on startup

PR #671, opened 2025-07-07
readme: Remove Cycles Optimizer referencs

StandingPadAnimations:
We removed the Cycles Optimizer a few versions back, but this seems to have slipped under our radar.


TheDuckCow (approved):
Thanks for the cleanup

PR #672, opened 2025-07-15
license: Relicense CommonMCOBJ parser under 0BSD

StandingPadAnimations:
To make the CommonMCOBJ parser more easily useable for projects, I've decided to relicense it under the 0BSD license (which is effectively public domain, though without some of the legal headaches with actually releasing the code into the public domain).


TheDuckCow (approved):
No complaints

PR #676, opened 2025-08-09
refactor: Create Vivy V2 JSON types

StandingPadAnimations:
This creates the basic types for Vivy V2, which cleans up the overall
JSON structure. This does not yet implement support for V2 in MCprep,
but it sets the foundation for later.

Also included in this change is the addition of typing_extensions 4.7.1,
which backports some new typing features to older versions of Python
(this comes directly from the Python Foundation). We're using an old
version here due to Blender 2.83 using Python 3.7.

Also includes in this are some refactors that take advantage of `TypedDict`,
as well as a readability improvement from the use of the `get` method for
certain areas, as opposed to conditionals. The `TypedDict` refactor shouldn't
break anything in Vivy currently, but I'll test locally to confirm that.

### DIFF
--- a/MCprep_addon/conf.py
+++ b/MCprep_addon/conf.py
@@ -30,6 +30,7 @@ import bpy
 from bpy.utils.previews import ImagePreviewCollection
 import bpy.utils.previews
 
+from .materials.vivy import vivy_v1_json_struct as vjs1
 from .materials.vivy import vivy_v2_json_struct as vjs2
 
 
@@ -150,7 +151,7 @@ class MCprepEnv:
 		self.vivy_cache = None
 		
 		# The JSON file for Vivy's materials
-		self.vivy_material_json: Optional[vjs2.VivyJSON] = None
+		self.vivy_material_json: Union[vjs1.VivyJSON, vjs2.VivyJSON, None] = None
 		self.vivy_enabled = False
 
 		# State for name changes in the Vivy config
--- a/MCprep_addon/conf.py
+++ b/MCprep_addon/conf.py
@@ -30,6 +30,8 @@ import bpy
 from bpy.utils.previews import ImagePreviewCollection
 import bpy.utils.previews
 
+from .materials.vivy import vivy_v2_json_struct as vjs2
+
 
 # -----------------------------------------------------------------------------
 # TYPING UTILITIES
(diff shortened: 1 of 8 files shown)


StandingPadAnimations:
Looks like we got a bit of a bug with accessing the Vivy blend file, though JSON parsing seems to be fine based on the prep materials dialogue
image

StandingPadAnimations:
Alright got that fixed, just was missing a `bpy.path.abspath` call in `get_vivy_blend`

TheDuckCow:
I'll admit @StandingPadAnimations that I got somewhat intimidated by the diff count and have been procrastinating hard on this. Maybe we can just do another followup review when we have a common time? We can chat over discord to find a good slot sometime soon, I think we are well due for an MCprep release with many goodies that have been added

MCprep_addon/conf.py
```diff
@@ -30,6 +30,9 @@
 from bpy.utils.previews import ImagePreviewCollection
 import bpy.utils.previews
 
+from .materials.vivy import vivy_v1_json_struct as vjs1
+from .materials.vivy import vivy_v2_json_struct as vjs2
```
TheDuckCow, line 34: As discussed, goal is to phase out v1 and go only with v2.

MCprep_addon/lib/typing_extensions.py
TheDuckCow, line 1: Maybe a docstring at the top to share where this is derived from, ie the python git repo, and at what point we would no longer need this in python 3.8 (aside from one type: NotRequired which is used in typed dictionaries) and in 3.11 wouldn't need it at all.

MCprep_addon/materials/vivy/vivy_v1_json_struct.py
```diff
@@ -0,0 +1,67 @@
...
+# but for support in older Blender versions, we bundle
+# an older version of typing_extensions, which backports
+# older Python modules
+if sys.version_info >= (3, 11):
+	from typing import TypedDict, NotRequired
+elif sys.version_info >= (3, 8) and sys.version_info < (3, 11):
+	from typing import TypedDict
+	from ...lib.typing_extensions import NotRequired
+else:
+	from ...lib.typing_extensions import TypedDict, NotRequired
```
TheDuckCow, line 38: As is evident here, once we upgrade to minimum Blender 3.0, this gets at least a little simpler.

TheDuckCow (approved):
As discussed, will be merging this into the feature branch. Did only light review, can do more exhaustive user review in the future. Maybe at that point, we won't need the typing copied file.

StandingPadAnimations:
One thing I've just realized is this does include a fix for Vivy V1, so I'll create a new branch and cherrypick that commit into it

StandingPadAnimations:
Since we've now taken care of documenting the source of the local copy of `typing_extensions`, I'll merge this in the feature branch.

PR #677, opened 2025-08-10
feature: Add Import Multiple OBJs as Chunks

StandingPadAnimations:
This adds the "Import Multiple OBJs	as Chunks" feature, which allows users to import multiple OBJs as part of a larger chunk, and have the offsets automatically handled by MCprep.

This is done by first loading the CommonMCOBJ headers of all the selected OBJs, making sure the block scales and world are the same. Then, we go through each header and store both the min and max export bounds for X, Y, and Z in 3 arrays. Then, we sort those 3 arrays, and create a new min bound using the first values of all 3 arrays, and a new max value using the last values in all 3 arrays. Then, we calculate a center point and negate it. Finally, we import all of the OBJs, undo the export offset (and block offset), and apply the new offset to them based on the center point.

The nice thing about this method is that it doesn't matter if there's holes in the final assembled world, nor does it matter if it's a perfect cube or if there's protruding bits.

EDIT: here's a demonstration of the feature:

https://github.com/user-attachments/assets/71c17600-ab20-49fd-8d14-110bed5f47dc

### DIFF
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -22,7 +22,7 @@ from enum import Enum, auto
 import os
 import math
 from pathlib import Path
-from typing import List, Optional, Union
+from typing import List, Optional, Tuple, Union
 import shutil
 from .commonmcobj_parser import CommonMCOBJ, CommonMCOBJTextureType, parse_header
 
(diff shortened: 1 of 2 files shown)


TheDuckCow:
Very nice! Though I never really got back to it, I still think one day it'd be awesome to make a direct importer with a chunk based system, calling jmc2obj or mineways from the command line.

Main initial comment is that I'm thinking we should try to do something to make the collections more linked. Right now, they each show up as their own separate collections - and maybe for right now, that's just as well. But when we get to an auto importer, imagine we'd create one top-level collection, and then either subcollections for each chunk, or even just created an instanced empty representing the chunk in question (and thus if you want to "mess" with the chunk by making it real, your work isn't lost if you do a re-export).

So - what are your thoughts in this context, keep them separate top-level collections for now? I could see it either way - as technically, nothing forces each of these to actually be in chunk format, you could be just importing different individual structures for instance.

StandingPadAnimations:
I think for now, we could work on at least putting the imports in a single subcollection. Blender 4.5 also introduced a new `Import OBJ` node (https://developer.blender.org/docs/release_notes/4.5/geometry_nodes/#import-nodes), which we could utilize as part of some sort of "adaptive" setup (ex. the user exports a larger area for one of the chunks, or adds another chunk)

TheDuckCow:
If we put it into one parent collection, I guess part of the question then becomes what to name that collection. We could either attempt to find a base common name between the files, and failing that, maybe use the parent directory name. In an actual automated context though, I think this kind of thing would matter less and you're more likely to just have one world name and we name the chunks after, well, chunks (or some variation thereof). So, probably doesn't matter too much at this stage since I'm guessing it won't be used manually that much.

I'll have more time to test this out and see the edge cases after I'm back from SIGGRAPH to see if I come up with any other thoughts

MCprep_addon/world_tools.py
```diff
@@ -846,6 +846,299 @@ def split_world_by_material(self, context: Context) -> bpy.types.Collection:
...
+		# There are a number of bug reports that come from the generic call
+		# of obj importing. If this fails, should notify the user to try again
+		# or try exporting again.
+		#
+		# In order to not overly supress error messages, below we only capture
+		# very tight specific errors possible, so that other errors still get
+		# reported so it may be passed off to blender devs. It is understood the
+		# below errors are not internationalized, so someone with another lang
+		# set will get the raw bubbled up traceback.
+		obj_import_err_msg = (
```
TheDuckCow, line 960: Really my only true comment for this PR is that I presume a lot of this is copied over from the main obj importer code. Could we break this down into its own utility function and reference in both places? that way, we don't end up with quirks in different edge cases. If it were just 10-30 lines of code I'd be more neutral about it, but this seems likely a decent sized chunk of code.

Basically, we could treat a normal OBJ import as a singular "chunk".

It would mean also splitting out the warning so that we still do that as just one check at the end.

TheDuckCow (changes requested):
So, unlike I expected, I do have a comment heh 😅 

I'd vote for one of these options:
- Either merge the chunk importer with the primary importer obj, thus letting the main operator import multiple OBJs (but still allowing single selection of non header obj files)
- Or, split the core functionality into a utility used in both places (and then, I'd probably put the chunk importer under the advanced section too, as it'll be less common until we create some more automated tooling around it).

Open to your thoughts.

MCprep_addon/world_tools.py
StandingPadAnimations: Personally, I'm of the opinion that, at least for now, we should keep these 2 separate operators (but share the core functionality). The main reason is because the chunk import has a hard requirement for the CommonMCOBJ header, while the normal OBJ importer currently doesn't have a CommonMCOBJ requirement. If we do decide to eventually require CommonMCOBJ for all OBJ imports (IMO we should wait till MCprep 4.0 for that), then we could merge the 2 operators then.

TheDuckCow: I was thinking, without reading it _too_ deeply, that we could try and split out the requirement or have different behavior to fallback. If your stance is that it would be too challenging to do, then sure, let's keep it the same. Extra code is preferable over forcing this requirement, likely even for MCprep 4.0 as there will always be cases of old obj imports laying around.

TheDuckCow (approved):
Per last comment and your counter arg, approving

PR #678, opened 2025-09-26
fix: Add backface culling to redstone torches

### LINKED ISSUE #647: Backface Culling

### Describe the context

In Minecraft some blocks (glass, doors and the nether portal) have backface culling, while Mcprep does not, so I asked why and got [this answer](https://discord.com/channels/737871405349339232/737872746700079235/1315674714613743666). Indeed, it wasn't necessary until Minecraft 1.21.2, in which Mojang changed the redstone torch model, so without backface culling it looks weird.

image

Add a "Use Backface Culling" bool property to the Prep Materials panel (It should be enabled by default) . So if it's enabled Mcprep will add this node setup to the materials that should have backface culling.

image

That setup works fine on eevee, eevee next and cycles. It should also work fine on different versions of blender 

Edit:
I made a better node setup
image

I'm not sure what you mean by "What existing workaround do you have today ?". Is it like "Is there any existing implementation of that node setup ?"

StandingPadAnimations:
It seems that redstone torches repeaters, and comparators need special handling in newer versions of Minecraft when it comes to rendering. Specifically, a backface culling setup needs to be added to the material that subtracts the backface from the alpha  channel of the texture, making backfaces transparent (and thus properly rendering redstone torches).

This could in theory also be needed for `Non-solid` blocks, but based on the list provided by Minecraft.Wiki (https://minecraft.wiki/w/Opacity), it seems redstone torches, repeaters, and comparators are the only thing that require this special handling (perhaps due to that "second layer" that they have as of 1.21.5?). On the top of my head, everything else in that list should render fine without backface culling.

Unfortunately this does impact the old redstone torch texture, so extra handling will need to be done for that. Now I wish CommonMCOBJ V1 had MC version in the header, that would make things so much simpler -_-

Fixes #647

### DIFF
--- a/MCprep_addon/MCprep_resources/mcprep_data_update.json
+++ b/MCprep_addon/MCprep_resources/mcprep_data_update.json
@@ -4595,6 +4595,12 @@
 			"water_flow",
 			"water_overlay",
 			"water_still"
+		],
+		"backface_culling": [
+			"redstone_torch",
+			"redstone_torch_on",
+			"repeater_*tick_on*",
+			"comparator_on*"
 		]
 	},
 	"make_real": [
@@ -4632,5 +4638,5 @@
 		"sunflower",
 		"water_still",
 		"water"
-	]
+	]	
 }
--- a/MCprep_addon/MCprep_resources/mcprep_data_update.json
+++ b/MCprep_addon/MCprep_resources/mcprep_data_update.json
@@ -4595,13 +4595,7 @@
 			"water_flow",
 			"water_overlay",
 			"water_still"
-		],
-		"backface_culling": [
-			"redstone_torch",
-			"redstone_torch_on",
-			"repeater_*tick_on*",
-			"comparator_on*"
-		]
+		]	
 	},
 	"make_real": [
 		"chest",
--- a/MCprep_addon/MCprep_resources/mcprep_data_update.json
+++ b/MCprep_addon/MCprep_resources/mcprep_data_update.json
@@ -1,4 +1,898 @@
 {
+	"block_model_uv": {
+		"cauldron": {
+			"from,[0, 0, 0]": {
+				"down": [
+					0,
+					14,
+					4,
+					16
+				],
+				"east": [
+					14,
+					13,
+					16,
+					16
+				],
+				"north": [
+					12,
+					13,
+					16,
+					16
+				],
+				"south": [
+					0,
+					13,
+					4,
+					16
+				],
+				"west": [
+					0,
+					13,
+					2,
+					16
+				]
+			},
+			"from,[0, 0, 12]": {
+				"down": [
+					0,
+					2,
+					2,
+					4
+				],
+				"east": [
+					2,
+					13,
+					4,
+					16
+				],
+				"north": [
+					14,
+					13,
+					16,
+					16
+				],
+				"west": [
+					12,
+					13,
+					14,
+					16
+				]
+			},
+			"from,[0, 0, 14]": {
+				"down": [
+					0,
+					0,
+					4,
+					2
+				],
+				"east": [
+					0,
+					13,
+					2,
+					16
+				],
+				"north": [
+					12,
+					13,
+					16,
+					16
+				],
+				"south": [
+					0,
+					13,
+					4,
+					16
+				],
+				"west": [
+					14,
+					13,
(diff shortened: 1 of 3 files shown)


MCprep_addon/MCprep_resources/mcprep_data_update.json
TheDuckCow, line 1: For the changes in this file, it shoudl be possible to run the update script without losing changes. So, I think we need to define these backface culling properties into the base file right? `UpdateJson/mcprep_data_base.json`. Worth testing it works this way when running the update script.

StandingPadAnimations: Ah, I forgot about the base file. `mcprep_data_base.json` is the source used for generating `mcprep_data_update.json` if I understand correctly?

EDIT: yeah that seems to be the case, will move the list over

TheDuckCow: Looks like this PR doesn't have the latest version of the "compiled" json update, can you run `python MCprep_addon/mcprep_data_refresh.py -auto` on your machine to see what you get? I'm getting a rather large diff, with some manual-looking things like `colormap/grass` removed and a whole new section showing up for `"block_model_uv":` implying maybe it just wasn't compiled after other recent changes to the base json file?

StandingPadAnimations: On my end, the large changes I get in the diff are (and this is with Minecraft 1.21.8):
- `block_model_uv` (I think from a recent PR from zNight where UVs for some JSON models were fixed)
- `backface_culling` (this PR)
- New MC textures (creaking, pale garden, temperature variations for some mobs, happy ghast, etc)
- `colormap/grass` was moved to `short_grass` (this I believe was from a PR zNight made which fixed a texture issue in exported 1.21.5 worlds)

I'll commit the new JSON update

TheDuckCow: Meant to upload this before - my resultant json for context, if you want to compare

[mcprep_data_update.json](https://github.com/user-attachments/files/22703538/mcprep_data_update.json)

TheDuckCow: Ah and I didn't refresh the page to see you responded, but yes - that all makes sense. Thanks for the update!

TheDuckCow: Ok one more check: what version of MC do you have on your system? I just updated mine and see that the extract `Got jar version: 1.21.9.jar` results in this diff

[resource_diff.txt](https://github.com/user-attachments/files/22703597/resource_diff.txt)

StandingPadAnimations: > Ok one more check: what version of MC do you have on your system? I just updated mine and see that the extract `Got jar version: 1.21.9.jar` results in this diff

I have 1.21.8 on my system currently. Haven't yet updated to 1.21.9

TheDuckCow (approved):
Approved. As discussed, there's only the one slight downside for older OBJ models where backface culling as a slight unintended effect, but it's honestly pretty minor and users can override with their own custom material setups as they please.


Notice how the back side of the lamp is invisible, becomes visible as it rotates around:

Screen Shot 2025-10-04 at 1 36 09 PM

TheDuckCow:
Something to look at:

```
(4.5.3)   	all_tests	70	2	1	test_canonical_test_mappings
```

Also seems to be failing in github actions, we _should_ look at that.

Looking more at the details, turns out that indeed we probably just need to update the tests:

```
FAIL: test_canonical_test_mappings (materials_test.MaterialsTest.test_canonical_test_mappings)
Test some specific mappings to ensure they return correctly.
  File "/home/runner/work/MCprep/MCprep/test_files/materials_test.py", line 664, in test_canonical_test_mappings
    self.assertEqual(
AssertionError: 'grass' != 'colormap/grass'
- grass
+ colormap/grass
 : grass should map to grass (mc, not colormap/grass
```

StandingPadAnimations:
While tests don't work with the changes in the PR, I have opened #683 to track it, and I'll probably open a follow up PR to fix that issue. Since however the main feature works as is for the most part, I'll merge this in

PR #679, opened 2025-09-27
feat: Add drag and drop handler for JSON models

StandingPadAnimations:
This is a small thing, but Blender 4.1 introduced the new FileHandler API, which allows defining custom drag and drop behavior. Since we already have an operator for importing JSON models, we can extend it relatively easily by creating a subclass of `FileHandler`.

Since it's exclusive to 4.1 and above, we wrap the `FileHandler` type around another type called `FileHandlerIfSupported`, which otherwise will simply be a dummy type at runtime if the Blender version is 4.0 or below. In addition, we also only register the `MCPREP_FH_import_minecraft_model_file` class if Blender 4.1 and above is being used.

### DIFF
--- a/MCprep_addon/spawner/mcmodel.py
+++ b/MCprep_addon/spawner/mcmodel.py
@@ -731,12 +731,12 @@ class MCPREP_OT_place_json_model_with_gizmo(bpy.types.Operator):
 	filename_ext = ".json"
 	filepath: bpy.props.StringProperty(subtype='FILE_PATH', options={'SKIP_SAVE'})
 	def invoke(self, context, event):
-		from mathutils import Vector
+		from mathutils import Vector, Quaternion
 
 		self.has_hit = False
-		self.hit_location = Vector((0.0, 0.0, 0.0))
-		self.hit_normal = Vector((0.0, 0.0, 1.0))
-		self.update_raycast(context, event)
+		self.hit_location = Vector((0, 0, 0))
+		self.hit_normal = Vector((0, 0, 1))
+		self.rotation_quat = Quaternion()
 		self.draw_handle = bpy.types.SpaceView3D.draw_handler_add(
 			self.draw_callback, (context,), 'WINDOW', 'POST_VIEW'
 		)
@@ -746,15 +746,13 @@ class MCPREP_OT_place_json_model_with_gizmo(bpy.types.Operator):
 	def modal(self, context, event):
 		context.area.tag_redraw()
 		if event.type == 'MOUSEMOVE':
-			self.update_raycast(context, event)
+			from .spawner_gizmo import update_raycast
+			self.has_hit, self.hit_location, self.hit_normal, self.rotation_quat = update_raycast(context, event)
 		elif event.type == 'LEFTMOUSE' and event.value == 'PRESS':
 			if self.has_hit:
-				# Get the difference between Z-up and the normal of the raycast
-				up_axis = Vector((0.0, 0.0, 1.0))
-				rotation_quat = up_axis.rotation_difference(self.hit_normal)
 				bpy.ops.mcprep.import_model_file(filepath=self.filepath,
 												location=self.hit_location,
-												rotation=rotation_quat.to_euler())
+												rotation=self.rotation_quat.to_euler())
 				self.finish(context)
 				return {'FINISHED'}
 			else:
@@ -769,57 +767,13 @@ class MCPREP_OT_place_json_model_with_gizmo(bpy.types.Operator):
 		bpy.types.SpaceView3D.draw_handler_remove(self.draw_handle, 'WINDOW')
 		context.area.tag_redraw()
 
-	def update_raycast(self, context, event):
-		from bpy_extras import view3d_utils
-		from mathutils import Vector
-		from mathutils.geometry import intersect_line_plane
-
-		mouse_pos = (event.mouse_region_x, event.mouse_region_y)
-		region, region_3d = context.region, context.space_data.region_3d
-		ray_origin = view3d_utils.region_2d_to_origin_3d(region, region_3d, mouse_pos)
-		ray_direction = view3d_utils.region_2d_to_vector_3d(region, region_3d, mouse_pos)
-		depsgraph = context.evaluated_depsgraph_get()
-		result, location, normal, _, _, _ = context.scene.ray_cast(depsgraph, ray_origin, ray_direction)
-
-		if result:
-			self.has_hit = True
-			self.hit_location, self.hit_normal = location, normal
-		else:
-			intersection = intersect_line_plane(ray_origin, ray_origin + ray_direction, Vector(), Vector((0,0,1)))
-			if intersection:
-				self.has_hit = True
-				self.hit_location, self.hit_normal = intersection, Vector((0,0,1))
-			else:
-				self.has_hit = False
-
-	def draw_callback(self, context):
-		import gpu
-		from mathutils import Vector, Matrix
-
-		from .spawner_gizmo import draw_fading_grid
+	def draw_callback(self, context) -> None:
+		from .spawner_gizmo import draw_callback
 
 		if not self.has_hit:
 			return
-
-		shader_info = gpu.shader.from_builtin('UNIFORM_COLOR')
-
-		original_blend = gpu.state.blend_get()
-		original_depth_test = gpu.state.depth_test_get()
-		gpu.state.blend_set('ALPHA')
-		gpu.state.depth_test_set('NONE')
-
-		up_axis = Vector((0.0, 0.0, 1.0))
-		rotation_quat = up_axis.rotation_difference(self.hit_normal)
-		transform_matrix = Matrix.Translation(self.hit_location) @ rotation_quat.to_matrix().to_4x4()
-
-		gpu.matrix.push()
-		gpu.matrix.multiply_matrix(transform_matrix)
-
-		draw_fading_grid(shader_info, size=2.0, subdivisions=20, rings=10, base_color=(0.7, 0.7, 0.7))
-
-		gpu.matrix.pop()
-		gpu.state.blend_set(original_blend)
-		gpu.state.depth_test_set(original_depth_test)
+		
+		draw_callback(self.hit_location, self.rotation_quat)
 
 classes = (
 	MCPREP_OT_spawn_minecraft_model,
--- a/MCprep_addon/spawner/mcmodel.py
+++ b/MCprep_addon/spawner/mcmodel.py
@@ -731,12 +731,10 @@ class MCPREP_OT_place_json_model_with_gizmo(bpy.types.Operator):
 	filename_ext = ".json"
 	filepath: bpy.props.StringProperty(subtype='FILE_PATH', options={'SKIP_SAVE'})
 	def invoke(self, context, event):
-		from mathutils import Vector, Quaternion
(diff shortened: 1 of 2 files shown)


StandingPadAnimations:
Here's a video demonstrating this in action. I've also tested in Blender 3.6 to make sure that the wrapper workaround works.

https://github.com/user-attachments/assets/1f7d6e1a-15c9-440a-a2a1-7bb673a3a3c5

MCprep_addon/spawner/mcmodel.py
```diff
@@ -39,6 +39,11 @@
 Element = Sequence[Union[Dict[str, VectorType], TexFace]]
 Texture = Dict[str, str]
 
+# This is wrapper type that we use for FileHandler, since it's
+# only availible in Blender 4.1 and above. In older versions of
+# Blender, we just set it to the generic object type
+FileHandlerIfSupported = bpy.types.FileHandler if util.min_bv((4, 1)) else object
```
StandingPadAnimations: Just want to ask, should we create a new file that contains these wrapped classes? We could also then create a single `register` wrapper that would handle the version checks. 

Alternatively we could also move the entire file handler to some new file, maybe something like `ops_not_fully_compatible.py` with the role of containing any operators not compatible with the entire range of versions MCprep supports.

TheDuckCow:
Wow I did not know this was possible, very cool. Going to live try this in a bit, thanks for pulling this together

MCprep_addon/spawner/mcmodel.py
```diff
@@ -683,6 +689,21 @@ def execute(self, context):
...
+
+class MCPREP_FH_import_minecraft_model_file(FileHandlerIfSupported):
+	bl_idname = "MCPREP_FH_import_minecraft_model_file"
+	bl_label = "File handler for JSON import"
+	bl_import_operator = "mcprep.import_model_file"
+	bl_file_extensions = ".json"
+
+	@classmethod
+	def poll_drop(cls, context) -> bool:
+		return (context.area and context.area.type == 'VIEW_3D')
```
TheDuckCow, line 700: This only polls on the drop release right, or is it running often as you are dragging?

One thought for me is whether we could return false if we can somehow detect this is not a json in the right format, to avoid conflicts. Obviously the downside would mean we are parsing the json multiple times, but ideally we could just do our own lightweight file io stream to check the first few lines (rather than load the whole json).

Main reason for asking this: in case other addons also define a handler for json files. Obviously, if they don't also do the same narrower check things would still be broken, but maybe there's a way for us to be good citizens and be as narrow as possible in where it applies.

MCprep_addon/spawner/mcmodel.py
TheDuckCow: I don't have a strong conviction, other than towards simplicity where possible. If we create different files, it opens up the complexity of what should be in these files if we have multiple different classes that are introduced in different versions, do we need to have files per each specific class?

In the updater code, not that I'd say it's an excellent reference of code structure, but it does work: we just define an on the fly [mocked class](https://github.com/Moo-Ack-Productions/MCprep/blob/6514aa5b9bf39c1241865c9fbefe8c0d150305c6/MCprep_addon/addon_updater_ops.py#L42) if it fails due to any exceptions.

TheDuckCow (approved):
I'm going to proactively approve as this alone is already pretty great. Ideas for potential improvements, which could be done in this PR or a follow up one:

- Automatically initiate the object placement tool, similar to adding a cube or others from the add menu. Having it always be placed right into the center of the scene I think is not that convenient.
- Per prior comment, being able to detect and skip the poll for the file handler if not a valid file (presuming that means it doesn't "consume" the drop, letting other handlers attempt to consume it). To test if there's a way to actually accomplish this, could do a little experiment where we just check for something like a keyword in the filename to enable/disable the poll, and flip it with a mocked version of MCprep installed at the same time, to see if you can get one of them to specifically handle in one case or the other.
- Responded to the other comment below on my 2c

Happy to re-review if there are updates (edit: I did try on my machine, it's working well)

MCprep_addon/spawner/mcmodel.py
StandingPadAnimations: I was thinking we could store all of those into a single file that simply represents all operators/classes that can't be registered in the entire supported version range. That way, to other developers, there's at least a clear separation of "operators/classes that are bound to exist and be registered" and "operators/classes that may or may not exist and be registered". That said, I think for now we could leave it in this file, since it's the only case currently of an operator/class that has those constraints. If we do add another thing like it, we can revisit the idea.

MCprep_addon/spawner/mcmodel.py
StandingPadAnimations: > This only polls on the drop release right, or is it running often as you are dragging?

If I'm understanding [the docs](https://docs.blender.org/api/master/bpy.types.FileHandler.html#bpy.types.FileHandler.poll_drop) right, it's only on the actual drop itself.

> One thought for me is whether we could return false if we can somehow detect this is not a json in the right format, to avoid conflicts. Obviously the downside would mean we are parsing the json multiple times, but ideally we could just do our own lightweight file io stream to check the first few lines (rather than load the whole json).

I think that would be a nice idea, though I'm not sure if it's possible. From what I understand, the `FileHandler` class simply calls the operator defined in `bl_import_operator` by passing the file path, but there don't appear to be any method of getting the dropped file path in the actual handler. It seems like that's left up to the operator to handle. In general, `poll_drop` seems to mostly restrict which parts of the Blender UI the handler is called for.

> Main reason for asking this: in case other addons also define a handler for json files.

This made me curious as to how Blender handles two addons with a handler for the same file extension. Testing it locally with a simple script that prints out the contents of a JSON file, it appears that Blender uses whatever handler was last registered for a particular file extension. 

Oddly though, [the pull request for file handlers](https://projects.blender.org/blender/blender/pulls/116047) says in the case of conflicting file handlers, a menu should appear with all possible handlers (and the user just selects the intended one), and it does appear that if I swap `FileHandlerIfSupported` for `bpy.types.FileHandler`, it works, so I'm not sure what Python does with the alias to cause that bug.

StandingPadAnimations:
Alright pushed a quick bug fix related to conflicting file handlers. It turns out the workaround from earlier somehow didn't cause Blender to create the menu of handlers defined for a particular file extension (in the case of 2 or more file handlers defined for a specific file extension), so instead we're just importing the `FileHandler` type directly and creating a dummy alias for it if the import fails.

MCprep_addon/spawner/mcmodel.py
TheDuckCow: > last registered for a particular file extension

Yeah that pretty much makes sense. At this stage though, I'm guessing it'd be quite rare so I'm not too concerned. let's leave the polling as it is then, since we can't check further. Thanks for the explanations

StandingPadAnimations:
Since it was requested, I've extended the feature to allow for custom placement.

https://github.com/user-attachments/assets/a0ee0711-a716-4a11-ba7a-1f3797881212

@TheDuckCow this is going to need some extra review (possibly in Saturday's meeting), since this uses the `gpu` module to implement the gizmo, alongside some ray-casting-logic for placement. A good amount of the code was done with Gemini, since I don't have much experience with the `gpu` module.

StandingPadAnimations:
Did some refactoring to move much of the gizmo logic into `spawner_gizmo.py`, to make it more easily reusable (as well as keeping the main spawner code clean)

TheDuckCow:
Wow really nice job, it looks great - but yes, agreed will need to make sure it's safe and working well (and that we have sufficient error logging too of course). I could actually imagine this operation being useful for other spawners as well, since I get the feeling fewer and fewer people are really spending time using the 3D cursor these days. But, we can take it one step at a time.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -694,6 +723,57 @@ def execute(self, context):
...
+	def modal(self, context, event):
+		context.area.tag_redraw()
+		if event.type == 'MOUSEMOVE':
+			from .spawner_gizmo import update_raycast
+			self.has_hit, self.hit_location, self.hit_normal, self.rotation_quat = update_raycast(context, event)
+		elif event.type == 'LEFTMOUSE' and event.value == 'PRESS':
+			if self.has_hit:
+				bpy.ops.mcprep.import_model_file(filepath=self.filepath,
+												location=self.hit_location,
+												rotation=self.rotation_quat.to_euler())
```
StandingPadAnimations: Just so I don't forget, we do call an operator in an operator with the gizmo. I'm aware it generally isn't advised to call operators in operators, so would it be best to do something similar to what `MCPREP_OT_spawn_minecraft_model` does and essentially clone the operator?

Although that does make me wonder though why we have a separate `MCPREP_OT_spawn_minecraft_model` from `MCPREP_OT_import_minecraft_model_file` that's mostly identical (minus some additional base classes that `MCPREP_OT_import_minecraft_model_file` has).

MCprep_addon/spawner/mcmodel.py
```diff
@@ -694,6 +723,57 @@ def execute(self, context):
...
+	
+	filename_ext = ".json"
+	filepath: bpy.props.StringProperty(subtype='FILE_PATH', options={'SKIP_SAVE'})
+	def invoke(self, context, event):
+		from mathutils import Vector, Quaternion
+
+		self.has_hit = False
+		self.hit_location = Vector((0, 0, 0))
+		self.hit_normal = Vector((0, 0, 1))
+		self.rotation_quat = Quaternion()
```
StandingPadAnimations: Also thinking it might be a good idea to have a basic dataclass to encapsulate these variables so we can simplify the raycast function arguments, and avoid juggling around 4 variables. 

While I'm at it, I'll look into eliminating these instance variables completely, and just have them as local variables to a function (honestly not a fan of having mutable instance variables in an operator, or really classes in general, it just feels weird), though that might be a bit difficult given we have a callback to deal with).

StandingPadAnimations: Looking into it, it doesn't appear that there's a clean way to remove these instance variables since we have a draw handler for the 3D view (and it uses variables that need to be updated in `modal`), so I'll just condense this down to a dataclass and call it a day

MCprep_addon/spawner/mcmodel.py
StandingPadAnimations: Alright I've moved the main spawning code that's identical to all three operators to a new function that's in the `ModelSpawnBase` class, so all three operators can also call it. This also avoids the use of a nested operator, which should hopefully reduce issues with undoing operations.

TheDuckCow (approved):
As discussed:

- Would be nice if we're able to activate our own redo last operators at the end, mostly to support the snapping placements (offset)
  - if we do that, would be downstream from the actual raycast placement system
  - Then if we want to also correct for orientation (since there's normal output), woudl need to do a simple voxel-check and clearing of orientation as a part of it
  - Also: we don't have to include that in this PR, could be a followup. Feature is already good as-is
- Minor flake8 things, double spacing between toplevel function and removing trailing whitespace.

StandingPadAnimations:
> Would be nice if we're able to activate our own redo last operators at the end, mostly to support the snapping placements (offset) 

Was able to get this done, and snapping works, though adjusting location and rotation after the fact seems broken. Maybe we'll fix that in a later PR, but I'll merge this in since snapping was the primary reason for getting the redo last operators working

PR #681, opened 2025-10-01
fix: Don't try to read Vivy JSON file if Vivy isn't enabled

StandingPadAnimations:
Previously, we only checked if both the JSON existed and if Vivy was enabled. If one of those conditions was `False`, then MCprep would attempt to read the Vivy JSON, which won't exist if Vivy was disabled.

This adds a simple check for whether Vivy is enabled or not, and if it isn't, to just return early from the `reload_vivy_json` function.


TheDuckCow (approved):
No complaints

PR #685, opened 2025-10-07
refactor: Deduplicate shared OBJ importer functionality

StandingPadAnimations:
To avoid large chunks of duplicated code between 2 OBJ import operators, this PR extracts the common parts into 2 functions.

One function, `validate_and_return_header` performs file validation (making sure the file exists, is a file, is an OBJ and/or corrects MTL selection, etc) and extracts the header (if it exists). Another, `import_obj_file`, performs the import steps and object property setups, as well as perform offsets if provided.

A downside of the deduplication is that warnings are a bit harder to propogate. For now, we've simply done a sort of "hack" where we have a dummy exception type for warnings, and use that for anything intended to be shown in an `INFO` report (as opposed to an `ERROR` report). Maybe it would be better to create a deticated object (perhaps `MCprepInfo`), but that's beyond the scope of this commit.

### DIFF
--- a/MCprep_addon/conf.py
+++ b/MCprep_addon/conf.py
@@ -351,6 +351,9 @@ class MCprepError(object):
 	file: str
 	msg: Optional[str] = None
 
+class MCprepWarning(BaseException):
+	"""Dummy exception class that denotes a warning"""
+	pass
 
 def updater_select_link_function(self, tag):
 	"""Indicates what zip file to use for updating from a tag structure.
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -30,7 +30,7 @@ import bpy
 from bpy.types import Context, Camera
 from bpy_extras.io_utils import ExportHelper, ImportHelper
 
-from .conf import MCprepError, env, VectorType
+from .conf import MCprepError, MCprepWarning, env, VectorType
 from . import util
 from . import tracking
 from .materials import generate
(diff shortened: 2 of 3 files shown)


StandingPadAnimations:
Was originally going to make this part of #677, but that was merged before I completed this commit, hence the separate PR

TheDuckCow:
Ahh sorry, thought you were saying it'd be too much trouble to split out. Ok, will take a look at this tongiht/tomorrow to just validate it works on my side. Appreciate the effort here 🙏

StandingPadAnimations:
> Ahh sorry, thought you were saying it'd be too much trouble to split out.

Yeah I should have probably clarified, splitting out common functionality isn't really too difficult (minus propagation of warnings). My main issue was with merging the two operators to one, since in my view, it would be weird if a singular OBJ importer operator accepted an OBJ file on its own but not with other OBJ files being imported at the same time.

MCprep_addon/world_tools.py
TheDuckCow, line 1: UI-level comment - how likely / many people do you think will use this feature? If it's less than 5 or 10%, I'd say we should move it, to avoid confusion from users of which to use:

Screen Shot 2025-10-11 at 1 44 15 PM

MCprep_addon/conf.py
TheDuckCow, line 1: Unrelated to this PR, but I do notice in my console this excpetion whenever opening Blender - would be good to silence this error

```
...
Failed to register scene update handler
Error in bpy.app.handlers.load_post[5]:
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/mcprep_ui.py", line 2104, in defer_vivy_init
    env.reload_vivy_json(path)
  File "/Users/patrickcrawford/Library/Application Support/Blender/4.3/scripts/addons/MCprep_addon/conf.py", line 170, in reload_vivy_json
    with open(json_path, 'r') as f:
         ^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'vivy_materials.json'
mcprep updater: Running background check for update
...
```

TheDuckCow (approved):
Works well, thanks for the refactor. Recommend making the one UI change and then here or in another PR would be good to address the mentioned exception. Good to merge otherwise! Nice cleanup


My multi blender run test:

```
(4.5.3)   	all_tests	70	2	0	No errors
(4.3.1)   	all_tests	70	2	0	No errors
Compiled in 35.9s + tests ran in 263.9s
Total of 299.8s with exit code 0
```

MCprep_addon/conf.py
StandingPadAnimations: This should be fixed with #681, which is merged into `dev`. I just haven't updated this branch to match `dev` yet

StandingPadAnimations: Updated this branch against `dev`, can confirm the error no longer occurs

MCprep_addon/world_tools.py
StandingPadAnimations: Done

StandingPadAnimations:
Merging this into `dev`

PR #694, opened 2025-11-22
Major overhaul of JSON model importer for stability, texture resolution, and tweaks to UV

BolWaffy:
Significantly overhauled how `mcmodel.py` handles textures, materials, and UVs when importing Minecraft JSON models. It fixes several issues with texture lookup, `#` reference resolution, different resource packs, individual JSON model imports, incorrect material indices, etc.

Improves importer reliability across various resource packs.

### **Key Improvements**

* **Reliable texture reference resolution**
  Introduces `get_final_texture_key` to correctly follow chained `#` texture references and avoid circular recursion.

* **Improved image lookup**
  `locate_image` now normalizes paths and searches in this order:

  1. The model’s own resource pack
  2. The active resource pack
  3. The addon’s default fallback

  This fixes many “missing textures” produced by the old version.

* **Correct material creation & indexing**
  Materials are only created for valid, resolved textures.
  Faces now receive consistent `material_index` values, fixing "index out of bounds" errors and other material assignment bugs.

* **Improved UV scaling for non-square textures**
  UV adjustments/scaling only occur when the texture’s **height < width**, preventing distortions on models using 64×32, 128×64, or other non-1:1 textures. UV scaling is not needed when **height > width**.

* **Better error handling & logging**
    * Added better error handling using `try/except` inside the element and face creation loops of `add_model`. This allows the script to **skip problematic model faces or elements** and continue importing the rest of the geometry, preventing the object from completely not importing.

---

I tested on 12 different resource packs and custom mod assets.

If you want to quickly test the new importer. Here is a script that automatically imports and validates each JSON object in a directory.
[Auto Test Minecraft Model Importer.py](https://github.com/user-attachments/files/23691454/Better.Auto.Test.Minecraft.Model.Importer.py)

### DIFF
--- a/MCprep_addon/spawner/mcmodel.py
+++ b/MCprep_addon/spawner/mcmodel.py
@@ -316,18 +316,14 @@ def locate_image(
 
 	model_path = Path(model_filepath).resolve()
 
-	# ---------------------------------------------------------------
-	#  RELATIVE PATH HANDLING ("./texture")
-	# ---------------------------------------------------------------
+	# RELATIVE PATH HANDLING ("./texture")
 	if local_path.startswith("."):
 		relative_candidate = (model_path.parent / cleaned_png).resolve()
 		if relative_candidate.is_file():
 			return str(relative_candidate)
 		# Continue to search order A -> B -> C if not found
 
-	# ---------------------------------------------------------------
-	#  NAMESPACE SUPPORT
-	# ---------------------------------------------------------------
+	# NAMESPACE SUPPORT
 	if ":" in cleaned:
 		namespace, texpath = cleaned.split(":", 1)
 	else:
@@ -338,10 +334,7 @@ def locate_image(
 	tex_fragment = texpath + ".png"
 
 	try:
-	# ---------------------------------------------------------------
-	#  SEARCH ORDER A: ALL detected pack roots (multi-pack support)
-	# ---------------------------------------------------------------
-
+		# SEARCH ORDER A: ALL detected pack roots (multi-pack support)
 		pack_roots = find_all_pack_roots(model_filepath)
 		for root in pack_roots:
 			root = Path(root).resolve()
@@ -363,17 +356,14 @@ def locate_image(
 				env.log(f"Texture found in model origin: {flat2}", vv_only=True)
 				return str(flat2)
 
-	# ---------------------------------------------------------------
-	#  SEARCH ORDER B: Active Resource Pack
-	# ---------------------------------------------------------------
+		# SEARCH ORDER B: Active Resource Pack
 		root = Path(bpy.path.abspath(context.scene.mcprep_texturepack_path)).resolve()
 		candidate = root / "assets" / namespace / "textures" / tex_fragment
 		if candidate.is_file():
 			env.log(f"Texture found in Active Resource Pack: {candidate}", vv_only=True)
 			return str(candidate)
-	# ---------------------------------------------------------------
-	#  SEARCH ORDER C: MCprep default texture pack
-	# ---------------------------------------------------------------
+
+		# SEARCH ORDER C: MCprep default texture pack
 		prefs = util.get_user_preferences(context)
 		root = Path(bpy.path.abspath(prefs.custom_texturepack_path)).resolve()
 		candidate = root / "assets" / namespace / "textures" / tex_fragment
@@ -387,10 +377,7 @@ def locate_image(
 
 		return MCprepError(e, line, file, f"Error searching texture packs for '{img}': {e}")
 
-	# ---------------------------------------------------------------
-	#  FALLBACK
-	#  If nothing is found, return the last candidate path.
-	# ---------------------------------------------------------------
+	# If nothing is found, return the last candidate path.
 	candidate = str(candidate if 'candidate' in locals() else model_path.parent / cleaned_png)
 	env.log(f"All Failed trying to find pack path, fallbacking to {candidate}", vv_only=True)
 	return candidate
--- a/MCprep_addon/spawner/mcmodel.py
+++ b/MCprep_addon/spawner/mcmodel.py
@@ -514,7 +514,6 @@ def add_model(
 				if isinstance(tex_pth, MCprepError):
 					return tex_pth
 
-
 				if len(textures) > 1 and img != "particle":
 					# Use unique name for non-particle textures if multiple textures exist
 					name = f"{obj_name}_{img}"
@@ -522,9 +521,8 @@ def add_model(
 					# Use generic name if only one main texture, or for the 'particle' texture
 					name = f"{obj_name}"
 
-				mat = None
-
 				# 3. Create or retrieve the material only if its value is NOT a reference ('#')
+				mat = None
 				if not textures[img].startswith("#"):
 					mat = add_get_material(name, tex_pth, use_name=False)
 
@@ -535,7 +533,6 @@ def add_model(
 
 				# 5. If have a valid material object, link it up.
 				if mat is not None:
-
 					# Material is only linked to the object's mesh once
 					if mat.name not in obj_mats:
 						obj_mats.append(mat)
--- a/MCprep_addon/spawner/mcmodel.py
+++ b/MCprep_addon/spawner/mcmodel.py
@@ -304,107 +304,115 @@ def get_final_texture_key(texture_ref, textures, visited=None):
 	return key
 
 
-def locate_image(
-	context: Context, textures: Dict[str, str], img: str, model_filepath: Path) -> str:
+def locate_image(context, textures, img, model_filepath):
 	"""
 	Finds and returns the final texture path from a texture key/reference in the model JSON.
-	
-	The search hierarchy is:
-	1. Recursively resolving texture references (e.g., '#side' -> '#all').
-	2. Cleaning the resulting path fragment (e.g., namespace, extension handling).
-	3. Searching the file in this order: Model's Origin Pack (A), Active Scene Resource Pack (B), and Addon Default Resource Pack (C).
 	"""
 
-	# 1. Resolve to final texture key. (e.g. #texture -> #all -> block/texture)
-	# Uses the `get_final_texture_key` function to get the final key
-	final_key = get_final_texture_key(f"#{img}", textures)
-	if final_key is None or final_key not in textures:
-		# Fallback if key is not found or is part of a circular reference
-		local_path = textures.get(img, img)
-	else:
-		local_path = textures[final_key]
-		
-	# Path Cleaning/Normalization
-	# Remove leading path separators AND the .png extension if present. For cleaner & safer searching
-	local_path_clean = local_path.replace("/", os.sep).replace("\\", os.sep)
-	local_path_clean = local_path_clean.lstrip(os.sep) # Prevent leading slash vulnerability
-	if local_path_clean.lower().endswith(".png"): # Prevent '.png.png' double extension
-		local_path_clean = local_path_clean[:-4]
+	# ---- 1. Resolve texture key (#foo -> foo -> actual path) ----
+	try:
+		final_key = get_final_texture_key(f"#{img}", textures)
+	except Exception as e:
+		line, file = env.current_line_and_file()
+		return MCprepError(
+			ValueError(f"Failed to resolve texture key '{img}'"),
+			line,
+			file,
+			f"Texture key resolution failed: {e}")
 	
-	# Append the extension once for a complete/clean path fragment
-	texture_path_fragment = local_path_clean + ".png"
+	if isinstance(final_key, MCprepError):
+		# Pass error upward instead of continuing
+		return final_key
 
-	# 2. Local relative paths (starting with `.`)
-	if local_path.startswith('.'):
-		# Path is relative to the model JSON file's directory.
-		directory = os.path.dirname(model_filepath)
-		return os.path.normpath(os.path.join(directory, texture_path_fragment))
+	# Get raw texture value or fallback
+	local_path = textures.get(final_key, textures.get(img, img))
 
-	# 3. Parse Namespace (e.g., 'minecraft:block/foo' -> 'block/foo')
-	if ":" in local_path:
-		namespace, local_path_remainder = local_path.split(":", 1)
+	# ---- 2. Normalize MC-style path ----
+	cleaned = normalize_texture_path(local_path)
+	cleaned_png = cleaned + ".png"
+
+	model_path = Path(model_filepath).resolve()
+
+	# ---------------------------------------------------------------
+	#  RELATIVE PATH HANDLING ("./texture")
+	# ---------------------------------------------------------------
+	if local_path.startswith("."):
+		relative_candidate = (model_path.parent / cleaned_png).resolve()
+		if relative_candidate.is_file():
+			return str(relative_candidate)
+		# Continue to search order A → B → C if not found
+
+	# ---------------------------------------------------------------
+	#  NAMESPACE SUPPORT
+	# ---------------------------------------------------------------
+	if ":" in cleaned:
+		namespace, texpath = cleaned.split(":", 1)
 	else:
 		namespace = "minecraft"
-		local_path_remainder = local_path
-	
-	# Re-apply path cleaning after namespace removal (safe, catches stray separators)
-	local_path_clean = local_path_remainder.replace("/", os.sep).replace("\\", os.sep)
-	local_path_clean = local_path_clean.lstrip(os.sep) # No leading separator
-	if local_path_clean.lower().endswith(".png"):
-		local_path_clean = local_path_clean[:-4]
-	texture_path_fragment = local_path_clean + ".png"
+		texpath = cleaned
 
-	# Retrieval of image file:
-	# --- A: Check Models Origin Resource Pack ---
-	# Search in the resource pack folder where the model JSON file is
+	texpath = texpath.lstrip("/")
+	tex_fragment = texpath + ".png"
+
 	try:
-		model_path_obj = Path(model_filepath).resolve()
-		pack_root = None
+	# ---------------------------------------------------------------
+	#  SEARCH ORDER A: ALL detected pack roots (multi-pack support)
+	# ---------------------------------------------------------------
+		pack_roots = find_all_pack_roots(model_filepath)
+		for root in pack_roots:
+			root = Path(root).resolve()
 
-		# Walk up the directory tree to find the 'assets' folder
-		for parent in [model_path_obj] + list(model_path_obj.parents):
-			if (parent / "assets").is_dir():
-				pack_root = parent
-				break
+			# NORMAL PACK (with assets/)
+			candidate = root / "assets" / namespace / "textures" / tex_fragment
+			if candidate.is_file():
+				env.log(f"Texture found in model origin: {candidate}", vv_only=True)
+				return str(candidate)
+				
+			# FLAT-PACK (not a normal structured pack)
+			flat1 = root / namespace / "textures" / tex_fragment
+			if flat1.is_file():
+				env.log(f"Texture found in model origin: {flat1}", vv_only=True)
+				return str(flat1)
 
-		if pack_root:
-			path_model = pack_root / "assets" / namespace / "textures" / texture_path_fragment
-			path_model_str = str(path_model)
+			flat2 = root / "textures" / tex_fragment
+			if flat2.is_file():
+				env.log(f"Texture found in model origin: {flat2}", vv_only=True)
+				return str(flat2)
 
-			if os.path.isfile(path_model_str):
-				env.log(f"Texture found in model origin: {path_model_str}", vv_only=True)
-				return path_model_str
+	# ---------------------------------------------------------------
+	#  SEARCH ORDER B: Active Resource Pack
+	# ---------------------------------------------------------------
+		root = Path(bpy.path.abspath(context.scene.mcprep_texturepack_path)).resolve()
+		candidate = root / "assets" / namespace / "textures" / tex_fragment
+		if candidate.is_file():
+			env.log(f"Texture found in Active Resource Pack: {candidate}", vv_only=True)
+			return str(candidate)
+	# ---------------------------------------------------------------
+	#  SEARCH ORDER C: MCprep default texture pack
+	# ---------------------------------------------------------------
+		prefs = util.get_user_preferences(context)
+		root = Path(bpy.path.abspath(prefs.custom_texturepack_path)).resolve()
+		candidate = root / "assets" / namespace / "textures" / tex_fragment
+		if candidate.is_file():
+			env.log(f"Texture found in MCprep default texture pack: {candidate}", vv_only=True)
+			return str(candidate)
+			
+
 	except Exception as e:
-		env.log(f"Error checking model origin path: {e}", vv_only=True)
+		line, file = env.current_line_and_file()
+		return MCprepError(
+			FileNotFoundError(f"Failed to resolve texture '{img}'"),
+			line,
+			file,
+			f"Error locating texture '{img}': {e}")
 
-	# --- B: Check Active Resource Pack ---
-	# If `A` failed, then try the active resource pack set in the scene properties
-	try:
-		resource_folder = bpy.path.abspath(context.scene.mcprep_texturepack_path)
-		path_active = os.path.join(
-			resource_folder, "assets", namespace, "textures", texture_path_fragment)
-		path_active = os.path.normpath(path_active)
-
-		if os.path.isfile(path_active):
-			return path_active
-	except Exception as e:
-		env.log(f"Error checking assigned Resource Pack: {e}", vv_only=True)
-	
-	# --- C: Check mcprep_default Resource Pack ---
-	# Fallback to default minecraft resource pack (custom_texturepack_path), the resource pack defined in MCprep's addon preferences
-	try:
-		addon_prefs = util.get_user_preferences(context)
-		resource_folder = bpy.path.abspath(addon_prefs.custom_texturepack_path)
-		path_active = os.path.join(
-			resource_folder, "assets", namespace, "textures", texture_path_fragment)
-		path_active = os.path.normpath(path_active)
-
-		if os.path.isfile(path_active):
-			return path_active
-	except Exception as e:
-		env.log(f"Error checking default minecraft Resource Pack: {e}", vv_only=True)
-	# If neither found, return the path that would be expected in the Active Resource Pack (B) for error messages/debugging.
-	return path_active
+	# ---------------------------------------------------------------
+	#  FALLBACK
+	#  If nothing is found, return the last constructed candidate path.
+	# ---------------------------------------------------------------
+	candidate = str(candidate if 'candidate' in locals() else model_path.parent / cleaned_png)
+	env.log(f"All Failed trying to find pack path, fallbacking to {candidate}", vv_only=True)
+	return candidate
 
 def read_model(
 	context: Context, model_filepath: Path) -> Tuple[Element, Texture]:
--- a/MCprep_addon/spawner/mcmodel.py
+++ b/MCprep_addon/spawner/mcmodel.py
@@ -508,7 +508,6 @@ def add_model(
 	uv_layer = bm.loops.layers.uv.verify()
 
 	materials = []
-	materials_remap = {}
 
 	if textures:
 		for img in textures: # img is the texture key (e.g., 'base', 'side')
@@ -531,7 +530,6 @@ def add_model(
 				if len(textures) > 1 and img != "particle":
 					# Use unique name for non-particle textures if multiple textures exist
 					name = f"{obj_name}_{img}"
-					materials_remap[mat_key] = textures[img] # Store original reference for face assignment
 				else:
 					# Use generic name if only one main texture, or for the 'particle' texture
 					name = f"{obj_name}"
--- a/MCprep_addon/spawner/mcmodel.py
+++ b/MCprep_addon/spawner/mcmodel.py
@@ -265,7 +265,7 @@ def normalize_texture_path(
 	return path
 
 def get_final_texture_key(
-	texture_ref: str, textures: Dict[str, str], visited: Optional[set] = None) -> Tuple[str, Optional[MCprepError]]:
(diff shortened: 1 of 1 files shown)


MCprep_addon/spawner/mcmodel.py
```diff
@@ -179,33 +175,142 @@ def add_get_material(
...
-				local_path = local_path.split(":")[1]
+			# The value is an actual path (e.g., 'block/stone'), return the key
+			return key
 
-			directory = os.path.join(
-				resource_folder, "assets", namespace, "textures")
-		return os.path.realpath(os.path.join(directory, local_path) + ".png")
+	# If the key is not in textures, return the key itself as the final key.
+	# This handles cases where a face references a key not explicitly defined.
+	return key
```
StandingPadAnimations: If `key` is the alternative return value regardless, then the `else` clause above could be removed.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -179,33 +175,142 @@ def add_get_material(
...
-	else:
-		if local_path[0] == '.':  # path is local to the model file
-			directory = os.path.dirname(model_filepath)
+	# Strip leading '#' for key lookup
+	key = texture_ref[1:] if texture_ref.startswith('#') else texture_ref
+
+	if key in visited:
+		# Detected circular dependency and break recursion (e.g., '#a' -> '#b' -> '#a')
+		env.log(f"Circular texture reference detected for key: {key}")
+		return None
```
StandingPadAnimations: Rather than return `None`, this function should return an `MCprepError` object (defined in `conf.py`), since (presumably) circular dependencies are an error with the model.

This would also mean changing the type signature of the function from `Optional[str]` to `Union[str, MCprepError]`.

Here's some reference on how to use `MCprepError` (note that the first parameter should be whatever exception class would be closest to describing the error):
https://github.com/Moo-Ack-Productions/MCprep/blob/4938a3036018feeacf2c9078f2b02784c99f2ffc/MCprep_addon/world_tools.py#L714-L715

MCprep_addon/spawner/mcmodel.py
```diff
@@ -179,33 +175,142 @@ def add_get_material(
...
+	else:
+		namespace = "minecraft"
+		local_path_remainder = local_path
+	
+	# Re-apply path cleaning after namespace removal (safe, catches stray separators)
+	local_path_clean = local_path_remainder.replace("/", os.sep).replace("\\", os.sep)
+	local_path_clean = local_path_clean.lstrip(os.sep) # No leading separator
+	if local_path_clean.lower().endswith(".png"):
+		local_path_clean = local_path_clean[:-4]
+	texture_path_fragment = local_path_clean + ".png"
```
StandingPadAnimations: I wouldn't use functions from `os` for this. We encourage the use of `pathlib` in new code for working with paths, since it handles a lot of this anyway.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -179,33 +175,142 @@ def add_get_material(
...
+		pack_root = None
+
+		# Walk up the directory tree to find the 'assets' folder
+		for parent in [model_path_obj] + list(model_path_obj.parents):
+			if (parent / "assets").is_dir():
+				pack_root = parent
+				break
+
+		if pack_root:
+			path_model = pack_root / "assets" / namespace / "textures" / texture_path_fragment
```
StandingPadAnimations: For this, [`joinpath`](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.joinpath) would be preferable since it's more clear what's going on.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -179,33 +175,142 @@ def add_get_material(
...
+	# --- B: Check Active Resource Pack ---
+	# If `A` failed, then try the active resource pack set in the scene properties
+	try:
+		resource_folder = bpy.path.abspath(context.scene.mcprep_texturepack_path)
+		path_active = os.path.join(
+			resource_folder, "assets", namespace, "textures", texture_path_fragment)
+		path_active = os.path.normpath(path_active)
+
+		if os.path.isfile(path_active):
+			return path_active
```
StandingPadAnimations: As noted earlier, we encourage the use of `pathlib` in new code, since it tends to be a lot cleaner than `os.path`.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -179,33 +175,142 @@ def add_get_material(
...
+	except Exception as e:
+		env.log(f"Error checking default minecraft Resource Pack: {e}", vv_only=True)
+	# If neither found, return the path that would be expected in the Active Resource Pack (B) for error messages/debugging.
+	return path_active
```
StandingPadAnimations: Since this is supposed to be an error case, `MCprepError` should be used here, to make it clear to any part of the code calling this function what is and isn't an error.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -179,33 +175,142 @@ def add_get_material(
...
+	# Fallback to default minecraft resource pack (custom_texturepack_path), the resource pack defined in MCprep's addon preferences
+	try:
+		addon_prefs = util.get_user_preferences(context)
+		resource_folder = bpy.path.abspath(addon_prefs.custom_texturepack_path)
+		path_active = os.path.join(
+			resource_folder, "assets", namespace, "textures", texture_path_fragment)
+		path_active = os.path.normpath(path_active)
+
+		if os.path.isfile(path_active):
+			return path_active
```
StandingPadAnimations: Being new code, this should use `pathlib`.

StandingPadAnimations (commented):
Did a basic skim of the code since this is still a draft. Will probably do something more in-depth this week now that university is out.

As an aside, it would be preferable if you had smaller, more focused commits as opposed to one large commit. It's easier to review stuff when they're smaller commits, since I (and anyone else reviewing this) can more easily review individual changes in isolation; a large commit with multiple changes is harder to review.

BolWaffy:
Meant to include the results from my automated test.

I tested **37 different mods, resource packs, and folder structures** that I just had lying around on my computer, attempting to import **9,846 models** (excluding the ones that threw “_JSON model does not contain any actual geometry_” warning).

| Old Importer | New Importer |
| :---: | :---: |
| OLD IMPORTER TEST RESULTS | NEW IMPORTER TEST RESULTS |

MCprep_addon/spawner/mcmodel.py
```diff
@@ -323,7 +323,7 @@ def read_model(
 	"""
 	try:
 		with open(model_filepath, 'r') as f:
-			obj_data = json.load(f)
+			obj_data = json.load(open(model_filepath, encoding="utf-8-sig"))
```
StandingPadAnimations: This additional `open` call is unneeded (and could cause issues with resource cleanup), the encoding can be passed to the `open` call in the `with` statement.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -175,6 +175,102 @@ def add_get_material(
 
 	return mat
 
+def find_all_pack_roots(
+	start_path):
```
StandingPadAnimations: This function needs a type signature to make it clear what it takes and what it returns

MCprep_addon/spawner/mcmodel.py
```diff
@@ -175,6 +175,102 @@ def add_get_material(
...
+	Finds ALL possible pack roots, including:
+		- normal packs with 'assets/'
+		- loose/unpacked mod folders containing `models/` and `textures/`
+		- nested packs
+
+	A pack root is ANY folder that contains:
+		• assets/<namespace>/...
+		OR
+		• textures/	  (for packs missing the assets folder)
+		• models/	  (same case)
```
StandingPadAnimations: Bullet points in code files should use `-` and not `•`. `-` is guaranteed to render in any text editor for the most part, while `•` isn't, and could appear as an empty box for devs who use terminal-based text editors (like myself)

MCprep_addon/spawner/mcmodel.py
```diff
@@ -175,6 +175,102 @@ def add_get_material(
...
+			roots.append(parent)
+
+		# Case B: loose/unpacked mod folder (your case)
+		# e.g. christmasfestivity/textures/block/*.png
+		if (parent / "textures").is_dir() and (parent / "models").is_dir():
+			roots.append(parent)
+
+		# Case C: even looser: any folder with a textures/ directory
+		if (parent / "textures").is_dir():
+			roots.append(parent)
```
StandingPadAnimations: Case B and C should use `elif`, so that if an earlier case succeeds, the runtime can just skip everything else

MCprep_addon/spawner/mcmodel.py
```diff
@@ -175,6 +175,102 @@ def add_get_material(
...
+	# Strip leading '#' if passed a reference by mistake
+	if path.startswith("#"):
+		path = path[1:]
+
+	# 1. Convert Windows slashes → MC slashes
+	path = path.replace("\\", "/")
+
+	# 2. Remove redundant "./"
+	while "/./" in path:
+		path = path.replace("/./", "/")
```
StandingPadAnimations: [`str.replace`](https://docs.python.org/3/library/stdtypes.html#str.replace) will change all occurrences of the substring, so the `while` loop is unneeded

MCprep_addon/spawner/mcmodel.py
```diff
@@ -175,6 +175,102 @@ def add_get_material(
...
+		"block\\foo"		-> "block/foo"
+		"block//foo"		-> "block/foo"
+		"../block/foo"		-> "block/foo"	(MC strips parent dirs)
+		"textures/block/foo" -> "block/foo" (MC removes leading folders)
+		"block/foo.png"		-> "block/foo"
+		"minecraft:block/foo" stays unchanged except path cleanup
+	"""
+
+	if not isinstance(path, str):
+		return path
```
StandingPadAnimations: Since the function has been annotated with `path` being a string, this check is unneeded. If someone calls the function with `path` not as a string, pretty much every LSP or static analyzer will catch it anyway

MCprep_addon/spawner/mcmodel.py
```diff
@@ -175,6 +175,102 @@ def add_get_material(
...
+		if (parent / "textures").is_dir():
+			roots.append(parent)
+
+	# Make unique and preserve order
+	seen = set()
+	final = []
+	for r in roots:
+		if r not in seen:
+			seen.add(r)
+			final.append(r)
```
StandingPadAnimations, line 212: More of a style comment, but the condition should be reversed to avoid excessive indentation.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -175,6 +175,102 @@ def add_get_material(
...
+	A pack root is ANY folder that contains:
+		• assets/<namespace>/...
+		OR
+		• textures/	  (for packs missing the assets folder)
+		• models/	  (same case)
+	"""
+	roots = []
+	start = Path(start_path).resolve()
+
+	for parent in [start] + list(start.parents):
```
StandingPadAnimations: This seems a bit ugly, and I don't understand why both `start` and `start.parents` are both being iterated over?

MCprep_addon/spawner/mcmodel.py
```diff
@@ -271,41 +271,39 @@ def normalize_texture_path(
...
+	# Strip '#' prefix
 	key = texture_ref[1:] if texture_ref.startswith('#') else texture_ref
 
+	# Avoid infinite recursion
 	if key in visited:
-		# Detected circular dependency and break recursion (e.g., '#a' -> '#b' -> '#a')
-		env.log(f"Circular texture reference detected for key: {key}")
-		return None
-
+		# env.log(f"Circular texture reference detected for key: {key}")
```
StandingPadAnimations: This should be removed if it's ultimately going to stay commented

MCprep_addon/spawner/mcmodel.py
```diff
@@ -271,41 +271,39 @@ def normalize_texture_path(
 
 	return path
 
-def get_final_texture_key(
-	texture_ref: str, textures: Dict[str, str], visited: Optional[set] = None) -> Optional[str]:
+def get_final_texture_key(texture_ref, textures, visited=None):
```
StandingPadAnimations: Why was the type signature removed? All that needed to change was the return signature to `Union[str, MCprepError]`. Without it, it's ambiguous what type of arguments this function expects, and what values it'll return.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -304,107 +304,115 @@ def get_final_texture_key(texture_ref, textures, visited=None):
 	return key
 
 
-def locate_image(
-	context: Context, textures: Dict[str, str], img: str, model_filepath: Path) -> str:
+def locate_image(context, textures, img, model_filepath):
```
StandingPadAnimations: Again, why was the type signature removed here, when only the return signature needed to change?

StandingPadAnimations:
Due to some weird things I've noticed in this PR (especially in subsequent changes), I have to ask this question: was AI used in the creation of this PR? While we don't forbid use of AI in contributions, [we do have rules in place](https://github.com/Moo-Ack-Productions/MCprep/blob/master/CONTRIBUTING.md#use-of-ai-in-development) on its use.

TheDuckCow:
Just chiming in to say thanks @BolWaffy for the continued contributions, and appreciate you Standingpad for the review upkeep

BolWaffy:
To answer your question directly: Yes, I do use AI tools, but I keep usage within the guidelines.
I use them primarily as a productivity aid for polishing commit messages, rewording docstrings/comments for clarity, and handling boilerplate or verifying API references (along with me checking the actual documentation). All code logic, decisions, and solutions are my own, and I verify everything I commit.
Regarding the "weird things" you noticed in the subsequent changes: I realize now that I opened this Draft PR earlier than I should have. You likely saw some "scratchpad" commits where I was still working on the implementation and cleanup, rather than polished final code.
As this is my first contribution to a shared project of this scale, I am still tuning my workflow regarding commits and PR timing. Going forward, I will make sure the final code is clean before marking it ready for review.
Thanks for the heads-up!

MCprep_addon/spawner/mcmodel.py
```diff
@@ -277,12 +277,10 @@ def get_final_texture_key(texture_ref: str, textures: Dict[str, str], visited: O
...
 	if key in visited:
-		line, file = env.current_line_and_file()
-		return MCprepError(
-			ValueError("Circular texture reference"),
-			line,
-			file,
-			f"Circular texture reference detected for key: {key}")
+		(line, file) = env.current_line_and_file()
+		error_collector.collect(MCprepError(Exception("Texture recursion"), line, file, f"Texture recursion on key '{key}' | Using '{key}'"))
+		return key
```
StandingPadAnimations: I'm not a fan of this, cause the caller of the function is not aware that this technically failed. Functions should return errors if something fails, and leave it up to the caller of the function to handle however they wish.

Regarding bmesh corruption: if a model has an error in it, in my view, it would be better to abort creating the model entirely, and let the user know, than to silently error in the end.

StandingPadAnimations: Also as an aside, the parenthesis around `line` and `file` is entirely unnecessary. `env.current_line_and_file` returns a tuple, and Python will already interpret `line, file = env.current_line_and_file()` as a tuple unpacking operation.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -1185,6 +1177,43 @@ def draw_callback(self, context) -> None:
...
+		if model_error_msgs:
+			for msg in model_error_msgs:
+				self.report({'ERROR'}, msg)
+				
+		if unhandled_errors_msgs:
+			for msg in unhandled_errors_msgs:
+				self.report({'ERROR'}, msg)
+				
+		self._model_errors = [] # Clear errors after reporting
+		self._unhandled_errors = []
```
StandingPadAnimations: As noted earlier, this isn't really a good way to gracefully handle errors. Low level functions should return error objects, and graceful handling should be done in the higher level functions that call them.

Perhaps if you *really* wanted to, you could make low level functions have the return signature of `Union[type_of_return, Tuple[type_of_return, MCprepError], MCprepError]`, where:

- `type_of_return` is the type of whatever object is returned when the function is successful
- `Tuple[type_of_return, MCprepError]` is returned if an issue occurs, but it's really a warning
- `MCprepError` for errors where the spawner needs to stop immediately

Then the caller could check what type was returned using `isinstance` calls and, if an warning or error is returned, handle accordingly. 

Again, IMO any error in the model should abort the model creation process entirely, but that's an option if complete abortion isn't desired

MCprep_addon/spawner/mcmodel.py
StandingPadAnimations: One thing I have realized here is if `roots` is empty, then `final` will also be empty. That case should be handled as well

MCprep_addon/spawner/mcmodel.py
BolWaffy: I placed parentheses around `line` and `file` since I saw them in the other files, which made it a tad easier on my eyes.
I’ll remove them before taking this PR out of draft.

MCprep_addon/spawner/mcmodel.py
BolWaffy: Ohh, gotcha. That makes more sense now. Thanks for the explanation. I'll rework that.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -289,23 +290,23 @@ def get_final_texture_key(texture_ref: str, textures: Dict[str, str], error_coll
 
 		# Is another reference
 		if isinstance(value, str) and value.startswith("#"):
-			return get_final_texture_key(value, textures, error_collector, visited)
+			return get_final_texture_key(value, textures, visited)
```
StandingPadAnimations: While it's great that the function now returns an error, the return signature doesn't match the actual return type. The return signature is `Tuple[str, Optional[MCprepError]]`, which indicates that all return values will be a tuple with a string, and either an `MCprepError` object, or `None`. However, the return paths here don't match this (the error path returns just an `MCprepError` object, and the main path returns just a string).

I know I mentioned `Tuple[str, Optional[MCprepError]]` as an option for warnings as opposed to complete errors (if that was desired). What that would mean is, in the case of a successful run, the function would return `("some key", None)`, and in the case of a warning, `("some key", MCprepError(...))`. In both cases, the key would be returned, and a warning could also be propagated (if the desired behavior was for the spawner to continue on, and not abort completely).

Personally, I do like the return paths here being "absolute" (i.e. string or error, not both), so my suggestion would be to change the return signature to `Union[str, MCprepError]`, and keep the existing return paths as is. If, however, the desired behavior is to treat error objects as warnings, refer to the above.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -289,23 +290,23 @@ def get_final_texture_key(texture_ref: str, textures: Dict[str, str], error_coll
...
+
 	# Key not found - use the literal name
 	return key
 
-
-def locate_image(context: Context, textures: Dict[str, str], img: str, model_filepath: str, error_collector: "ModelErrorCollector") -> str:
+def locate_image(
+	context: Context, textures: Dict[str, str], img: str, model_filepath: str) -> Union[str, Optional[MCprepError]]:
```
StandingPadAnimations: `Optional[MCprepError]` is equivalent to `Union[MCprepError, None]`. Based on the code paths, the `Optional` is unnecessary, and the return signature can be simplified to `Union[str, MCprepError]`

MCprep_addon/spawner/mcmodel.py
```diff
@@ -481,14 +485,15 @@ def read_model(
 	return elements, textures
 
 def add_model(
-	model_filepath: Path, obj_name: str = "MinecraftModel", error_collector: "ModelErrorCollector" = None) -> Tuple[int, bpy.types.Object]:
+	model_filepath: Path, obj_name: str = "MinecraftModel") -> Union[tuple[int, bpy.types.Object], Optional[MCprepError]]:
```
StandingPadAnimations: As note earlier, `Optional[MCprepError]` is a shorthand for `Union[MCprepError, None]`. Since this function only returns an object of either `Tuple[int, bpy.types.Object]` or `MCprepError` the `Optional` can be removed, and `MCprepError` can be kept.

Also, since MCprep supports Blender 2.8 (which uses Python 3.7), the built-in `tuple` type annotation is unfortunately unavailable (that was introduced in Python 3.9). Instead, the `Tuple` annotation from `typing` has to be used (just a simple case of swapping `t` for `T`).

StandingPadAnimations (commented):
I like the move towards absolute errors, and returning `MCprepError` objects in lower level functions. Just some comments about type annotations that either don't match the actual return values of the function, or could be simplified.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -96,14 +94,12 @@ def rotate_around(
...
 
-
 def add_element(
 	elm_from: VectorType = [0, 0, 0],
 	elm_to: VectorType = [16, 16, 16],
 	rot_origin: VectorType = [8, 8, 8],
 	rot_axis: str = 'y',
-	rot_angle: float = 0
-) -> list:
+	rot_angle: float = 0) -> list:
```
StandingPadAnimations: Similar to `tuple`, `list` is a Python 3.9 type annotation, so it has to be replaced with `List` for Python 3.7 support. 

Also, the type of the elements in the list needs to be added, so the return signature would be something like `List[type_of_elements]`

MCprep_addon/spawner/mcmodel.py
```diff
@@ -179,33 +173,228 @@ def add_get_material(
...
+	A pack root is ANY folder that contains:
+		- assets/<namespace>/...
+		OR
+		- textures/	  (for packs missing the assets folder)
+		- models/	  (same case)
+	"""
+	roots = []
+	start = Path(start_path).resolve()
+
+	for parent in list(start.parents):
```
StandingPadAnimations: `PurePath.parents` is a sequence, so list conversion isn't needed for iteration.

MCprep_addon/spawner/mcmodel.py
```diff
@@ -179,33 +173,228 @@ def add_get_material(
...
+		path = path[9:]
+
+	# 7. Remove empty leading slash
+	if path.startswith("/"):
+		path = path[1:]
+
+	return path
+
+def get_final_texture_key(
+	texture_ref: str, textures: Dict[str, str], visited: Optional[set] = None) -> Union[str, MCprepError]:
```
StandingPadAnimations: I'll have to double-check if `set` is also a Python 3.9 annotation, but iirc, this has to be swapped for `Set` (from the `typing` module)

MCprep_addon/spawner/mcmodel.py
```diff
@@ -813,16 +1094,17 @@ class MCPREP_OT_import_minecraft_model_file(
...
 
 	track_function = "model"
 	track_param = "file"
 	@tracking.report_error
 	def execute(self, context):
-		res = self.create_and_place_json_model(context, Path(self.filepath))
-		if res:
-			self.report({'ERROR'}, res.msg)
+		result = self.create_and_place_json_model(context, Path(self.filepath))
+		print(result)
```
StandingPadAnimations: I'm assuming this is a temporary debug print

StandingPadAnimations (commented):
Did a final skim over the pull request, just a few small things to change (and I need to test this locally) before we merge this into `dev`.

Since the requested changes are minimal, I'll just push the changes myself (probably some time this later this week).

MCprep_addon/spawner/mcmodel.py
StandingPadAnimations: Ok so I need some clarification, what are the intended types for `faces` and `edges` in this function? I know `verts` is supposed to be a `List[VectorType]`, but I can't seem to determine if `edges` is supposed to be that as well (and `faces` is a list of lists)? Also the return values don't seem to match the signature (at least according to my editor's LSP)

For now, I'll assume the return value is actually supposed to be `Tuple[List[VectorType], List[VectorType], List[Tuple[int, int, int, int]]]`, but correct me if that assumption is wrong

StandingPadAnimations: As an aside, if you haven't already, I would recommend making sure your Python LSP is configured to analyze type annotations. That way, at least your editor will warn you if the types of stuff don't match

MCprep_addon/spawner/mcmodel.py
```diff
@@ -179,33 +173,228 @@ def add_get_material(
...
+
+	# Make unique and preserve order
+	seen = set()
+	final = []
+	for r in roots:
+		if r not in seen:
+			seen.add(r)
+			final.append(r)
+
+	return final
```
StandingPadAnimations, line 214: My editor's LSP ([`based-pyright`](https://github.com/DetachHead/basedpyright)) is warning me that `final` is defined as `List[Path]`, while the `find_all_pack_roots` has the return signature of `str`. I'll make the assumption that `find_all_pack_roots` is supposed to return `List[Path]`, and see where that goes.

StandingPadAnimations:
Pushed a couple of commits and made some comments on a couple of LSP issues I encountered. *Seems* to be working fine with the vanilla resource pack, but of course I need to test this on some other models as well.

@zNightlord can you also review this for any issues, since you have more experience with JSON models than I do?

MCprep_addon/spawner/mcmodel.py
```diff
@@ -124,21 +120,19 @@ def add_element(
...
 
-	edges = []
-	faces = [
-		[0, 1, 2, 3],  # north
-		[5, 4, 7, 6],  # south
-		[1, 0, 4, 5],  # up
-		[7, 6, 2, 3],  # down
-		[4, 0, 3, 7],  # west
-		[1, 5, 6, 2]]  # east
+	edges: List[VectorType] = []
```
zNightlord: This is a list of two indices so `List[Tuple[int, int]]`

MCprep_addon/spawner/mcmodel.py
```diff
@@ -96,14 +94,12 @@ def rotate_around(
...
 
-
 def add_element(
 	elm_from: VectorType = [0, 0, 0],
 	elm_to: VectorType = [16, 16, 16],
 	rot_origin: VectorType = [8, 8, 8],
 	rot_axis: str = 'y',
-	rot_angle: float = 0
-) -> list:
+	rot_angle: float = 0) -> Tuple[List[VectorType], List[VectorType], List[Tuple[int, int, int, int]]]:
```
zNightlord: Edge is `List[Tuple[int, int]]`

MCprep_addon/spawner/mcmodel.py
```diff
@@ -323,292 +509,381 @@ def add_model(
...
+								obj_mats.append(mat)
+
+							# Find the index in the object's material list
+							mat_index = obj_mats.find(mat.name)
+						else:
+							env.log(f"Failed to generate placeholder material {placeholder_name}")
+
+					# uv can be rotated 0, 90, 180, or 270 degrees
+					uv_rot = d_face.get("rotation")
+					if uv_rot is None:
```
zNightlord: You can do `d_face.get("rotation", 0)` for default fallback here

zNightlord (commented):
Looking around it is fine in general, there are weird parts as mentioned probably how AI picture it and goes safe, again it is fine.

If something you check, not exists and purposely created it then it just makes no sense to check it again honestly.
https://github.com/Moo-Ack-Productions/MCprep/blob/ca3594b38905ca37ceb2c50f0ed50ecd9688515c/MCprep_addon/spawner/mcmodel.py#L635-L641

MCprep_addon/spawner/mcmodel.py
StandingPadAnimations: Ah that makes sense, I'll fix that real quick

MCprep_addon/spawner/mcmodel.py
StandingPadAnimations: Oddly, my LSP is stating `d_face` is a string, so I'll have to investigate that

StandingPadAnimations:
Could you clarify what the types of `faces` and `d_face` are supposed to be in this section? My LSP is stating that `faces` is either a `VectorType`, `Dict[str, str]`, or `None`. I've added a check for `Dict` in the same check for `None`, but this leaves `d_face` as a `str` (thus making `d_face.get(...)` be invalid, since `str` does not have the `get` method), since `faces` would be narrowed down to a `Dict[str, str]`, and `faces.get(...)` would, as such, return a `str`. 

```py
			faces = e.get("faces")
			for i in range(len(element[2])):
				try:
					f = element[2][i]


					if not faces:
						continue


					face_name = FACE_DIRECTIONS[i]
					d_face = faces.get(face_name)
					if not d_face:
						continue


					face_mat = d_face.get("texture")


					if not face_mat:
						continue
```

StandingPadAnimations:
It looks like the `e` that's being iterated in the for loop is assumed to be a `Dict[str, Dict[str, str]]` (aliased as `TexFace`). I'll try to refactor this so static analyzers and LSPs are aware of this assumption. @BolWaffy if you haven't already, I highly encourage using an LSP with type checking enabled when writing new code, as it highlights issues with types being assumed where those assumptions may not always hold true (assuming you're using VScode and Pyright, [make sure `python.analysis.typeCheckingMode` is set to something higher than `basic`](https://github.com/microsoft/pyright/blob/main/docs/settings.md))

StandingPadAnimations:
I've pushed a couple of type annotation fixes, as well as just general fixes based on what the LSP caught (such as an attempt to divide a variable that could either be a `int` or `str`, without first checking the type). I've also added a `TODO` for a later date (not now, maybe some time in the future, like next year) to go through and validate all the types being used in the algorithm (the main issue I've found is assuming variables are of a certain type, when they could be other types).

StandingPadAnimations:
Alright testing with modded blocks, and this seems to work with both drag-and-drop and regular imports

https://github.com/user-attachments/assets/ebe0320b-3564-49be-8a46-9474ca2879b4

StandingPadAnimations (approved):
Unless there are any objections, I'll merge this into `dev`

PR #699, opened 2025-12-03
feat: Set shading to flat for Mineways imports

StandingPadAnimations:
Mineways OBJs seem to have weird artifacts caused by vertex normals being set (and thus in Blender being shaded smooth). While this is being tracked in erich666/Mineways#151, a workaround is needed for existing OBJs to reduce issues.

This is done by simply setting the `use_smooth` property to `False` on the active selection, which will be the OBJ after importing, only if the OBJ is a Mineways export. This behavior can be disabled by calling the function with `mineways_fix_smooth_shading_artifacts` set to `False`. As this is done through the `foreach_set` function, there shouldn't be much of a hit to import performance. However, just in case, the end user can also disable the workaround.

If future releases of Mineways resolve the smooth shading issue, we could encapsulate it with a header check, either for the version (if future versions of Mineways completely eliminate setting vertex normals) or for an export option (if setting vertex normals can be disabled in the future, which is a more likely outcome).

### DIFF
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -883,7 +883,7 @@ class MCPREP_OT_import_world_split(bpy.types.Operator, WorldImporterBase, Import
         name=env._("Center Import"),
         description=env._("Decide if the import should be centered, and if so, how"),
         items=(
-            ('NONE', env._("Don't Center"), env._("Doesn't center the import")),
+            ('NONE', env._("Keep Original Position"), env._("Keeps the position of the import as-is")),
             ('XY', env._("Center by XY Plane"), env._("Centers the import by X and Y only")),
 			('XYZ', env._("Center by all 3 axis"), env._("Centers the import by all 3 axis")),
         ),
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -880,12 +880,12 @@ class MCPREP_OT_import_world_split(bpy.types.Operator, WorldImporterBase, Import
 		options={'HIDDEN'})	
 
 	center_import: bpy.props.EnumProperty(
-        name="Center Import",
-        description="Decide if the import should be centered, and if so, how",
+        name=env._("Center Import"),
+        description=env._("Decide if the import should be centered, and if so, how"),
         items=(
-            ('NONE', "Don't Center", "Doesn't center the import"),
-            ('XY', "Center by XY Plane", "Centers the import by X and Y only"),
-			('XYZ', "Center by all 3 axis", "Centers the import by all 3 axis"),
+            ('NONE', env._("Don't Center"), env._("Doesn't center the import")),
+            ('XY', env._("Center by XY Plane"), env._("Centers the import by X and Y only")),
+			('XYZ', env._("Center by all 3 axis"), env._("Centers the import by all 3 axis")),
         ),
         default='XY',
     )
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -877,15 +877,7 @@ class MCPREP_OT_import_world_split(bpy.types.Operator, WorldImporterBase, Import
 	fileselectparams = "use_filter_blender"
 	skipUsage: bpy.props.BoolProperty(
 		default=False,
-		options={'HIDDEN'})
-
-	minewaysShadingWorkaround: bpy.props.BoolProperty(
-		name=env._("Mineways Shading Workaround"),
-		description=env._(("Mineways imports have custom vertex normals defined, which may "
-			"cause artifacts. If enabled, this option will enforce flat shading on "
-			"Mineways imports.")),
-		default=True if bpy.app.version >= (4, 5, 0) else False
-	)
+		options={'HIDDEN'})	
 
 	center_import: bpy.props.EnumProperty(
         name="Center Import",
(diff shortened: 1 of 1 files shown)


TheDuckCow (commented):
Didn't live test - could you share before/after screenshots of what the artifacts look like, so we can have that for histprical reference on this PR?

Other question is whether we should try and be more specific - do all of the obj's exported from the affected versions of Mineways have metadata that we could use to identify it? I'd be weary of breaking or influencing shading for older model imports, unless you think the shading change is safe even for older ones too. It would be good to have tests images comparing imports for both old and new exports.

StandingPadAnimations:
> Didn't live test - could you share before/after screenshots of what the artifacts look like, so we can have that for histprical reference on this PR?

Here are some examples of artifacts:

image

image

image

> Other question is whether we should try and be more specific - do all of the obj's exported from the affected versions of Mineways have metadata that we could use to identify it? I'd be weary of breaking or influencing shading for older model imports, unless you think the shading change is safe even for older ones too. It would be good to have tests images comparing imports for both old and new exports.

This is caused by Mineways setting custom vertex normals, which has been the case for a long time now. Upstream, Eric noted that the issue seems to not occur on Blender 4.4, [but it does happen in Blender 5.0](https://github.com/erich666/Mineways/issues/151#issuecomment-3624592541) (and there's reports of this issue in Blender 4.5). I suspect it's related to some Cycles change in Blender 4.5 (though what change that may be, I'm not sure).

We could limit the fix to Blender 4.5 and above, though in my opinion, it's still helpful in older versions of Blender for when users do eventually migrate and open older files.

StandingPadAnimations:
I've merged `dev` into this branch (since there were some conflicts), and I've made it so the workaround now is only enabled by default in Blender 4.5 and above. The option will still exist in older versions, just that (for versions before Blender 4.5) the default will be `False`

StandingPadAnimations:
The upstream issue on the Mineways repo [has been closed](https://github.com/erich666/Mineways/issues/151#issuecomment-3712411600), on the basis that this is a Cycles-side issue.

TheDuckCow:
Oh wow, Brecht even commented on the above linked issue. I'm currently trying to test out the neutral case on my side on my windows machine, to double check the change here does in fact fix it.

MCprep_addon/world_tools.py
```diff
@@ -936,15 +958,22 @@ class MCPREP_OT_import_objs_as_chunks(bpy.types.Operator, WorldImporterBase, Imp
...
+        description=env._("Decide if the final assembly should be centered, and if so, how"),
         items=(
-            ('NONE', "Don't Center", "Doesn't center the final assembly"),
-            ('XY', "Center by XY Plane", "Centers the final assembly by X and Y only"),
-			('XYZ', "Center by all 3 axis", "Centers the final assembly by all 3 axis"),
+            ('NONE', env._("Don't Center"), env._("Doesn't center the final assembly")),
+            ('XY', env._("Center by XY Plane"), env._("Centers the final assembly by X and Y only")),
+			('XYZ', env._("Center by all 3 axis"), env._("Centers the final assembly by all 3 axis")),
         ),
         default='XY',
```
TheDuckCow, line 968: Per top-level comment, can we make the default be 'NONE'?

MCprep_addon/world_tools.py
```diff
@@ -936,15 +958,22 @@ class MCPREP_OT_import_objs_as_chunks(bpy.types.Operator, WorldImporterBase, Imp
...
-        description="Decide if the final assembly should be centered, and if so, how",
+        name=env._("Center Final Assembly"),
+        description=env._("Decide if the final assembly should be centered, and if so, how"),
         items=(
-            ('NONE', "Don't Center", "Doesn't center the final assembly"),
-            ('XY', "Center by XY Plane", "Centers the final assembly by X and Y only"),
-			('XYZ', "Center by all 3 axis", "Centers the final assembly by all 3 axis"),
+            ('NONE', env._("Don't Center"), env._("Doesn't center the final assembly")),
+            ('XY', env._("Center by XY Plane"), env._("Centers the final assembly by X and Y only")),
+			('XYZ', env._("Center by all 3 axis"), env._("Centers the final assembly by all 3 axis")),
```
TheDuckCow, line 966: And then for names, I'd suggest calling it "Default" instead of "Don't Center" as that feels a bit more intuitive, but open to other naming ideas.

TheDuckCow (approved):
Confirmed before state (even before prep materials)

Before state (with prep materials done after)

image


And with this branch, with the tickbox on (and prepping materials, though visible even before prepping)

image

So, appears to be working as expected.

Very side/unrelated note: I notice that centering import feature is in place, but "don't center" isn't the default. And to the average user, the wording is probably the opposite of what they expect it does. I'd argue for calling it "Default centering" or something like this, reference it matches blender's default for importing an obj into the scene, while the others are doing so differently.

MCprep_addon/world_tools.py
StandingPadAnimations: I assume you're referring to single OBJ imports and not chunk importing?

MCprep_addon/world_tools.py
StandingPadAnimations: I've renamed it (or at least the single-import version) to `Keep Original Position` to make it more explicit on what it actually does. Chunked imports will remain `Don't Center` since centering in that context relates to the final assembly

StandingPadAnimations:
Since the suggestions have been applied, and this PR is working as expected with resolving artifacts, I'll merge this into `dev`

PR #701, opened 2025-12-16
feat: Add centering options for single OBJ imports

StandingPadAnimations:
This is mostly for parity with importing OBJs as chunks. Like chunk imports, this will only be performed on OBJs that export CommonMCOBJ metadata.

In addition, this also adds a warning for OBJs that don't use the CommonMCOBJ header. Given it's been almost two years, I think it's safe to begin warning users about old OBJs, since most at this point should be exported with newer exporters (usually for newer Minecraft updates).

### DIFF
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -870,18 +870,45 @@ class MCPREP_OT_import_world_split(bpy.types.Operator, WorldImporterBase, Import
 		default=False,
 		options={'HIDDEN'})
 
+	center_import: bpy.props.EnumProperty(
+        name="Center Import",
+        description="Decide if the import should be centered, and if so, how",
+        items=(
+            ('NONE', "Don't Center", "Doesn't center the final assembly"),
+            ('XY', "Center by XY Plane", "Centers the final assembly by X and Y only"),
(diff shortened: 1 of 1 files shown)


StandingPadAnimations:
Looks like past me decided to enforce CommonMCOBJ headers if offsets are provided at all, even if they're all `0`. I've just made an adjust where `None` will be passed if the final offsets end up being `(0, 0, 0)` (since the behavior will be the same at that point)

TheDuckCow (approved):
Didn't live-test, but changes look good and make sense to me. Contemplated whether to suggest the default entry should be the first in the dropdown, but makes sense to have None first.

Just to confirm - with this, when users import their world, it will continue to the same default behavior as before if it was non-chunked right?

StandingPadAnimations:
> Just to confirm - with this, when users import their world, it will continue to the same default behavior as before if it was non-chunked right?

I've set the default to `XY` (i.e. center based on X and Y in Blender), both for consistency with chunked imports, and based on what users generally expect when importing worlds. Thus, the behavior won't be the same as before, but it should be more convenient for users

StandingPadAnimations:
If there are no objections to having the default behavior be center OBJs on the X and Y axis, then I'll go ahead and merge this

PR #702, opened 2025-12-30
Added Pixel Hashing to Image Consolidation

BolWaffy:
The current `Combine Images` operator can accidentally merge different image textures, which can sometimes cause a headache. So I upgraded the operator to use MD5 Pixel Hashing and compare the actual image data to verify they are identical before merging.

Also added 'UNDO' to the bl_options so Blender does not crash if the user undoes that step.

### DIFF
--- a/MCprep_addon/materials/material_manager.py
+++ b/MCprep_addon/materials/material_manager.py
@@ -288,11 +288,6 @@ class MCPREP_OT_combine_images(bpy.types.Operator):
 	def execute(self, context):
 		removeold = True
 
-		if bpy.app.version < (2, 78):
-			self.report(
-				{'ERROR'}, "Must use blender 2.78 or higher to use this operator")
-			return {'CANCELLED'}
-
 		if self.selection_only and len(context.selected_objects) == 0:
 			self.report(
 				{'ERROR'},
--- a/MCprep_addon/materials/material_manager.py
+++ b/MCprep_addon/materials/material_manager.py
@@ -21,7 +21,6 @@ import os
 
 import bpy
 
-import hashlib
 import numpy as np
 
 from . import generate
@@ -370,7 +369,7 @@ class MCPREP_OT_combine_images(bpy.types.Operator):
 
 	def get_image_hash(self, img: bpy.types.Image, w: int, h: int, buffer: np.ndarray = None) -> str:
 		"""
-		Generates a MD5 hash from image pixel data
+		Generates a byte-string hash from sampled image pixel data.
 		"""
 		if not img.has_data:
 			return None
@@ -381,12 +380,10 @@ class MCPREP_OT_combine_images(bpy.types.Operator):
 
 			img.pixels.foreach_get(pixels)
 
-			# Calculate stride (sample every Nth pixel), only start when image is larger than 4096 pixels (bigger than 64 x 64)
+			# Calculate stride: if pixels > 4096, sample every Nth pixel
 			stride = max(1, total_pixels // 4096)
 
-			sample = np.ascontiguousarray(pixels[::stride * 4])
-
-			return hashlib.md5(sample.data).hexdigest()
+			return pixels[::stride * 4].tobytes()
 
 		except Exception as e:
 			env.log(f"Failed to hash {img.name}: {e}", vv_only=True)
--- a/MCprep_addon/materials/material_manager.py
+++ b/MCprep_addon/materials/material_manager.py
@@ -32,6 +32,14 @@ from .. import util
 
 from ..conf import MCprepError, env
 
+# 4096 * 4, used to cap the amount of
+# pixels summed when not using the fast
+# prefilter with the Combine Images operator
+#
+# 4096 is the amount of pixels capped to,
+# and 4 is the RGBA channels
+RGBA_PIXELS_4096_MAX = 16384
+
 
 # -----------------------------------------------------------------------------
 # UI and utility functions
@@ -331,6 +339,9 @@ class MCPREP_OT_combine_images(bpy.types.Operator):
 		precount = len(bpy.data.images)
 
 		# Group images
+		#
+		# Here we use a list, since we're just trying to get the
+		# raw images at this point for comparison
 		groups: List[Tuple[str, int, int, bpy.types.Image]] = []
 		for img in images_to_check:
 			if img.type in ('RENDER_RESULT', 'COMPOSITING') or img.use_fake_user:
@@ -348,25 +359,29 @@ class MCPREP_OT_combine_images(bpy.types.Operator):
 		# Value: List of (image_obj, comparison_array, sum)
 		# store both the image (for remapping later on), the pixel array,
 		# and pixel array sum for super fast prefilter
-		unique_images: Dict[Tuple[str, int, int], List[Tuple[bpy.types.Image, NDArray[np.float32], float]]] = {}
+		unique_images: Dict[Tuple[str, int, int], List[Tuple[bpy.types.Image, NDArray[np.float32], Optional[np.float64]]]] = {}
 		images_to_remove = []
 
 		for base_name, w, h, img in groups:
 			cur_img_key = (base_name, w, h)
 
-			num_pixels = w * h * 4
-			img_array = np.empty(num_pixels, dtype=np.float32)
-			img.pixels.foreach_get(img_array)	# faster than `np.asarray(img.pixels)`
+			img_array = np.asarray(img.pixels)
+			pix_sum: Optional[np.float64] = None
 
-			# Use sampling unless strict is enabled
-			# 16384 = 4096 * 4
-			stride = max(1, num_pixels // 16384) if not self.strict_comparison else 1
-			cur_img_data = img_array[::stride] if stride > 1 else img_array
+			# Approximation of a image content check used
+			# if strict comparison is disabled. Unlike strict
+			# comparison, which checks the actual pixels, this
+			# sums up the pixels and uses that to represent the
+			# image contents
+			if not self.strict_comparison:
+				num_pixels = w * h * 4
+				stride = max(1, num_pixels // RGBA_PIXELS_4096_MAX)
+				cur_img_data = img_array[::stride] if stride > 1 else img_array
 
-			pix_sum = np.sum(cur_img_data)
-
+				pix_sum = np.sum(cur_img_data)
+	
 			if cur_img_key not in unique_images:
-				unique_images[cur_img_key] = [(img, cur_img_data, pix_sum)]
+				unique_images[cur_img_key] = [(img, img_array, pix_sum)]
 				continue
 
 			image_present = False
@@ -374,10 +389,11 @@ class MCPREP_OT_combine_images(bpy.types.Operator):
 
 			# Look through existing items in this Name/Size group
 			for uni_image, uni_data, uni_sum in unique_images[cur_img_key]:
-				# FAST PREFILTER
-				if not pix_sum == uni_sum:
+				# Fast prefilter, used as an alternative
+				# to strict comparison
+				if not self.strict_comparison and not pix_sum == uni_sum:
 					continue
-				elif not np.array_equal(uni_data, cur_img_data):
+				elif not np.array_equal(uni_data, img_array):
 					continue
 				image_present = True
 				present_image = uni_image
@@ -392,7 +408,7 @@ class MCPREP_OT_combine_images(bpy.types.Operator):
 				continue
 
 			# If the image is not present, add it to the list
-			unique_images[cur_img_key].append((img, cur_img_data, pix_sum))
+			unique_images[cur_img_key].append((img, img_array, pix_sum))
 
 		# Cleanup
 		to_delete = [
--- a/MCprep_addon/materials/material_manager.py
+++ b/MCprep_addon/materials/material_manager.py
@@ -294,6 +294,7 @@ class MCPREP_OT_combine_images(bpy.types.Operator):
 	track_function = "combine_images"
 	@tracking.report_error
 	def execute(self, context):
+		start_time = time.time()
 		# Setup image list
 		if self.selection_only:
 			if not context.selected_objects:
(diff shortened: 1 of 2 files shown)


MCprep_addon/materials/material_manager.py
```diff
@@ -285,90 +289,105 @@ class MCPREP_OT_combine_images(bpy.types.Operator):
 	def execute(self, context):
 		removeold = True
 
-		if self.selection_only is True and len(context.selected_objects) == 0:
+		if bpy.app.version < (2, 78):
```
TheDuckCow: We already support only Blender 2.8+ so we can just remove this block (appreciate trying to maintain the functionality though)

MCprep_addon/materials/material_manager.py
```diff
@@ -285,90 +289,105 @@ class MCPREP_OT_combine_images(bpy.types.Operator):
...
+			pixels = np.empty(width * height * 4, dtype=np.float32)
+			img.pixels.foreach_get(pixels)
+			
+			# Calculate buffer (sample every Nth pixel), only start buffering when image is larger than 4096 pixels (bigger than 64 x 64)
+			buffer = max(1, (width * height) // 4096)
+
+			# Reshape so buffer skips WHOLE pixels, keeping RGBA grouped
+			sampled = pixels.view().reshape(-1, 4)[::buffer]
+			
+			return hashlib.md5(sampled.tobytes()).hexdigest()
```
TheDuckCow: I had a high level over the speed concerns, so I did some timing tests. My results:

## Test script

```python
import time
t0 = time.time()
bpy.ops.mcprep.combine_images(skipUsage=True)
t1 = time.time()
print("Time in seconds: ", t1-t0)
```

## Testing the control, ie current dev branch

-- MC scene
Time in seconds:  0.02409982681274414
Time in seconds:  0.026123046875

Outcome: basically instant


-- General PBR scene (mixture of 1K and 2K assets, 95 images total across only ~13 materials)
Time in seconds:  0.013833999633789062
Time in seconds:  0.012299776077270508
Time in seconds:  0.012421846389770508

Outcome: basically instant

-- MC PBR scene (same as the MC one above, but swapped for a PBR pack; 224 initial images, )
Time in seconds:  0.1075751781463623
Time in seconds:  0.05999588966369629
Time in seconds:  0.06019115447998047

Outcome: basically instant


## Now with the new changes of this PR:


-- MC version:
Time in seconds:  0.03257012367248535
Time in seconds:  0.04587864875793457
Time in seconds:  0.0442049503326416

Outcome: still basically instant

-- General PBR scene (mixture of 1K and 2K assets, 95 images total across only ~13 materials)
Time in seconds:  3.233705997467041
Time in seconds:  3.362300157546997
Time in seconds:  3.2018487453460693

Outcome: Noticeably slower; didn't have any 8K assets so this could definitely end up getting even slower

-- MC PBR scene (same as the MC one above, but swapped for a PBR pack)
Time in seconds:  0.1335008144378662
Time in seconds:  0.12597393989562988
Time in seconds:  0.12844491004943848

Outcome: not instant, but fast enough


## Summary:

For typical MCprep user use cases, this is still fast enough, especially since it's not an operation that users are running that often. It can be quite slow if this were ever used in a larger PBR scene that uses higher resolution assets, but could be still mitigated by users working based off of selection.

MCprep_addon/materials/material_manager.py
```diff
@@ -285,90 +289,105 @@ class MCPREP_OT_combine_images(bpy.types.Operator):
...
+			
+		except Exception as e:
+			print(f"Failed to hash {img.name}: {e}")
+			return None
```
TheDuckCow: Code comment: let's avoid bare exceptions. If there can be an exception, let's try and focus the handling around the most narrow line(s) of code possible. Ideally, that would mean only 1 line of in a try/except block.

TheDuckCow (approved):
Approving, but making a small suggestion for the exception handling. Per my analysis below, the speed is indeed a little slower, but for most use cases, it will not be perceptible or will be largely quite tolerable. (I contemplated suggesting a secondary option in case users want to do a "fast" check, but this is not an operator that benefits from having multiple options in practice, and just increases complexity while being harder to understand).

BolWaffy:
I found a way to speed it up a little bit more. I will make a commit once I get back from work.

StandingPadAnimations:
I'll take a look at it as well in a couple of days (currently out of town, will be back by January 5th)

TheDuckCow:
Great, just re-request review once you do and we can do a final review and approval - from my standpoint, it's already pretty good so I'm curious what speedup you identified (is it lowering the max resolution a bit more?)

StandingPadAnimations:
Now that I think about it, I think we could actually just compare file contents. After all, the file contents have to be read anyway to perform the hash operation, so we can simply compare numpy arrays and skip the hashing process. That would probably help with performance as well. 

(Also avoids the possibility of hash collisions, which although rare can still happen)

BolWaffy:
Forgot it was New Year's yesterday. Anyway, I thought I found a way around not needing to load the image into memory, but I was incorrect. I just did a few micro optimizations. In my test, I think the code is 1.15x faster than the first edit.

BolWaffy:
I tried your method @StandingPadAnimations, and it works. 
Originally, I wasn’t using a buffer/stride logic, so I relied on `hashlib.md5()` for better memory overhead. However, now that the stride logic limits the sample data to 4096 pixels, the data size is small enough that the memory savings from MD5 is negligible compared to the speed of using `tobytes()` instead of `hashlib`

BolWaffy:
I already have code that retrieves image data blocks from materials, so if you want, I can add the Combine by Selection option, which should be straightforward.

I can also implement a similar hashing comparison approach with the Combine Materials tool. I also already have code that converts the material node tree into JSON. The added execution time would be minimal (just a couple of extra milliseconds).

MCprep_addon/materials/material_manager.py
```diff
@@ -268,106 +270,150 @@ def getMaterials(self, context):
...
+		description="Build images to consolidate based on selected objects' materials only",
 		default=False)
-	skipUsage: bpy.props.BoolProperty(
-		default=False,
-		options={'HIDDEN'})
+
+	strict_comparison: bpy.props.BoolProperty(
+		name="Strict Comparison",
+		description="Check every single pixel. Slower, but 100% accurate",
+		default=False)
```
StandingPadAnimations, line 299: I would rephrase the description as "Compare image contents, which while slower, will be more accurate"

Also @TheDuckCow given this is intended to solve an issue that exists, do you think this should be set to `True` by default?

MCprep_addon/materials/material_manager.py
```diff
@@ -268,106 +270,150 @@ def getMaterials(self, context):
...
+					if not mat or mat in processed_materials:
+						continue
 
-		name_cat = {}
-		data = bpy.data.images
+					processed_materials.add(mat)
+					if mat.use_nodes and mat.node_tree:
+						for node in mat.node_tree.nodes:
+							if node.type == 'TEX_IMAGE' and node.image:
+								target_images.add(node.image)
```
StandingPadAnimations: Stylistic comment: I would invert the condition to avoid extreme nesting.


```suggestion
							if node.type != 'TEX_IMAGE' or not node.image:
								continue
							target_images.add(node.image)
```

MCprep_addon/materials/material_manager.py
```diff
@@ -268,106 +270,150 @@ def getMaterials(self, context):
...
 
-			postcount = len(["x" for x in bpy.data.materials if x.users > 0])
-			self.report(
-				{"INFO"},
-				f"Consolidated {precount - postcount} materials, down to {postcount} overall")
-			return {'FINISHED'}
+		# Group images
+		groups = {}
+		for img in images_to_check:
+			if img.type in {'RENDER_RESULT', 'COMPOSITING'} or img.use_fake_user:
```
StandingPadAnimations: Since the comparison is against a static list of values, it's better to make this a tuple.


```suggestion
			if img.type in ('RENDER_RESULT', 'COMPOSITING') or img.use_fake_user:
```

MCprep_addon/materials/material_manager.py
```diff
@@ -268,106 +270,150 @@ def getMaterials(self, context):
...
+			w, h = img.size
+
+			base_name = util.nameGeneralize(img.name) if self.group_by_name else "GLOBAL"
+
+			key = (base_name, w, h)
+			groups.setdefault(key, []).append(img)
+
+		images_to_remove = []
+
+		# Hashing and Remapping
```
StandingPadAnimations: Since hashing requires reading the image contents into the hash function anyway, it would be a lot simpler to just compare image contents. Also probably a lot faster, especially when using Numpy.

Hashing in general is really only advised if you need an identifier based on data, like if you have to make files. However, since we already have the image data, it's both simpler and (most likely) faster to check the data directly, especially if we use vectorized Numpy operations. It also completely eliminates the rare (but non-zero) possibility of hash collisions.

StandingPadAnimations (changes requested):
Mostly stylistic comments, but one fundamental change is needed: instead of hashing, it would be simpler (and most likely faster) to compare image contents directly with Numpy. This would avoid the calculation time for a hash on image data that we already have, and it would avoid the possibility of hash collisions.

MCprep_addon/materials/material_manager.py
BolWaffy: I’ve been attempting to implement the direct data comparison using Numpy as you suggested, but so far every Numpy based comparison approach I've tried has actually been slower. I may be missing something obvious with how to structure the comparison efficiently.

BolWaffy:
Were you meaning this?

Also, I finished the `Combine Materials` to a similar approach to not remove materials that are truly different. Just need to clean it up a bit.

MCprep_addon/materials/material_manager.py
StandingPadAnimations: Looking at the code, I think the main issue is dictionaries that use the image contents as the key, and using keys for comparison. I've done some refactoring that's closer to what I meant by direct comparison (attached patch file below, can be applied with `git-apply`)

[comparision-refactor.patch](https://github.com/user-attachments/files/24501782/comparision-refactor.patch)

StandingPadAnimations:
I've pushed a commit that's closer to what I meant by direct image comparison (also regarding the force-push, I'm using [jj](https://github.com/jj-vcs/jj/) as my VCS)

EDIT: Since the array isn't being converted to a bytestring to act as a key to a dictionary, I've removed the 4096 limit as well, since *in theory* it shouldn't be needed

BolWaffy:
Updated to the suggested approach. I opted for a 'hybrid' solution to maintain performance; my testing (4060 images @ 256px, 2369 duplicates) with the `Fast Prefilter` decreased the execution time by ~93%.

MCprep_addon/materials/material_manager.py
```diff
@@ -331,64 +331,69 @@ def execute(self, context):
...
-			elif not img.has_data:
-				continue
-			# Empty images can't be compared, don't do anything
-			# with them
-			elif w == 0 or h == 0:
+			cur_img_key = (base_name, w, h)
+
+			num_pixels = w * h * 4
+			img_array = np.empty(num_pixels, dtype=np.float32)
+			img.pixels.foreach_get(img_array)	# faster than `np.asarray(img.pixels)`
```
StandingPadAnimations: Image pixels in Blender convert to `NDArray[np.float64]`. IMO to avoid precision loss, that datatype should be used.

Also do we have actual metrics for `img.pixels.foreach_get(img_array)` being faster than using `np.asarray(img.pixels)`? If not, the latter should be used to reduce room for error.

MCprep_addon/materials/material_manager.py
```diff
@@ -331,64 +331,69 @@ def execute(self, context):
...
 			present_image: Optional[bpy.types.Image] = None
-			img_array = np.asarray(img.pixels)
-			for uni_image, uni_image_array in unique_images[(base_name, w, h)]:
-				if not np.array_equal(uni_image_array, img_array):
+
+			# Look through existing items in this Name/Size group
+			for uni_image, uni_data, uni_sum in unique_images[cur_img_key]:
+				# FAST PREFILTER
+				if not pix_sum == uni_sum:
+					continue
```
StandingPadAnimations: Since we're dealing with (usually) small images, the sum of all pixels isn't a good metric to determine if 2 images are unique (especially since we cap it in the end anyway). Since strict comparison is supposed to actually check pixel contents, it would be preferable to do just that. Performance is important, but since this is currently opt-in, we don't need something as fast as the non-strict check.

If this is intended to run when strict check is disabled, then I would add another condition for checking if strict check is disabled.

MCprep_addon/materials/material_manager.py
```diff
@@ -331,64 +331,69 @@ def execute(self, context):
 		precount = len(bpy.data.images)
 
 		# Group images
-		#
-		# Here we use a list, since we're just trying to get the
-		# raw images at this point for comparison
```
StandingPadAnimations: Since this is an explanatory comment, I would avoid removing it

MCprep_addon/materials/material_manager.py
```diff
@@ -331,64 +331,69 @@ def execute(self, context):
...
+			num_pixels = w * h * 4
+			img_array = np.empty(num_pixels, dtype=np.float32)
+			img.pixels.foreach_get(img_array)	# faster than `np.asarray(img.pixels)`
+
+			# Use sampling unless strict is enabled
+			# 16384 = 4096 * 4
+			stride = max(1, num_pixels // 16384) if not self.strict_comparison else 1
+			cur_img_data = img_array[::stride] if stride > 1 else img_array
+
+			pix_sum = np.sum(cur_img_data)
```
StandingPadAnimations: Since this does seem to be intended to run only when strict check is disabled, I would wrap this around a condition:

```py
pix_sum = None
if self.strict_comparision:
	stride = max(1, num_pixels)
	cur_img_data = img_array[::stride] if stride > 1 else img_array
	
	pix_sum = np.sum(cur_img_data)

if cur_img_key not in unique_images:
	unique_images[cur_img_key] = [(img, cur_img_data, pix_sum)]
```

In this case, I would also wrap the `float` type in `Optional`, and update later sections to check if the sum is `None`

```py
unique_images: Dict[Tuple[str, int, int], List[Tuple[bpy.types.Image, NDArray[np.float32], Optional[float]]]] = {}
```

StandingPadAnimations (changes requested):
A couple of small changes, which I'll make on my end and push. The main thing is making it clearer in code that the fast prefilter is only supposed to run when strict comparison is disabled.

I did make a note regarding array creation performance; as far as I'm aware, there shouldn't be any major performance issues with `np.asarray(img.pixels)`, so unless there's a major performance issue with it, I don't think there should be any practical issues.

StandingPadAnimations:
I'll add some tests for strict comparison (mostly just a copy-paste of the existing combine images test to make sure the code works), and then merge it

StandingPadAnimations (approved):
Since tests past, and the code looks good, I'll merge this in

PR #703, opened 2026-01-16
Improved Material Consolidation with Node Tree Analysis

BolWaffy:
The current `Combine Materials` operator is limited and can accidentally merge materials that are actually different. I upgraded the operator to use a "fingerprinting" system that analyzes the full node tree, drivers, and animation data to verify that materials are only merged if they are functionally identical.

For the node_tree fingerprinting, it uses `bl_rna` properties to make it multi-version Blender compatible.
Also, it only considers nodes that actually contribute to the `Material Output` node; dead nodes are not included in the fingerprint.

**Note**: I've set this to a Draft as it's still a WIP. I have left comments in the code for a few things I'm still working on:
- [x] Toggle to allow users to choose if Action names should be used to differentiate fingerprints.
- [x] Users can choose the precision for how closely property values must match (adjusting the `round()` precision in the `to_primitive`& `serialize_keyframe` functions).
- [x] Automatically strip the `.001` suffixes from the chosen master material
- [x] Allow users to define the "Master" material (the material that will be kept) based on: Highest/Lowest node count | User count (popularity)*

### DIFF
--- a/MCprep_addon/materials/material_manager.py
+++ b/MCprep_addon/materials/material_manager.py
@@ -232,36 +232,36 @@ class MaterialFingerprint:
 
 # --- SERIALIZATION HELPERS ---
 
-def serialize_fcurve(fcurve: bpy.types.FCurve) -> FCurveData:
+def serialize_fcurve(precision: int, fcurve: bpy.types.FCurve) -> FCurveData:
 	"""
 	Converts an F-Curve data-block into a dictionary of primitive values for structural comparison.
 	"""
 	fcurve.keyframe_points.sort()
 
 	return FCurveData(
-        array_index=fcurve.array_index,
-        extrapolation=fcurve.extrapolation,
-        auto_smoothing=fcurve.auto_smoothing,
-        mute=fcurve.mute,
-        keyframes=[serialize_keyframe(kp) for kp in fcurve.keyframe_points]
-    )
+		array_index=fcurve.array_index,
+		extrapolation=fcurve.extrapolation,
+		auto_smoothing=fcurve.auto_smoothing,
+		mute=fcurve.mute,
+		keyframes=[serialize_keyframe(precision, kp) for kp in fcurve.keyframe_points]
+	)
 
-def serialize_keyframe(kp: bpy.types.Keyframe) -> KeyframeData:
+def serialize_keyframe(precision: int, kp: bpy.types.Keyframe) -> KeyframeData:
 	"""
 	Serializes individual keyframe points, rounding float values to ensure consistent hashing.
 	"""
 	return KeyframeData(
-        co=(round(kp.co[0]), round(kp.co[1])),
-        handles=(
-            round(kp.handle_left[0]), round(kp.handle_left[1]),
-            round(kp.handle_right[0]), round(kp.handle_right[1])
-        ),
-        types=(kp.handle_left_type, kp.handle_right_type, kp.interpolation, kp.easing),
-        dynamics=(round(kp.back), round(kp.amplitude), round(kp.period))
-    )
+		co=(round(kp.co[0], precision), round(kp.co[1], precision)),
+		handles=(
+			round(kp.handle_left[0], precision), round(kp.handle_left[1], precision),
+			round(kp.handle_right[0], precision), round(kp.handle_right[1], precision)
+		),
+		types=(kp.handle_left_type, kp.handle_right_type, kp.interpolation, kp.easing),
+		dynamics=(round(kp.back, precision), round(kp.amplitude, precision), round(kp.period, precision))
+	)
 
 # --- ANIMATION & DRIVER LOGIC ---
-def get_animation_fingerprint(id_data: bpy.types.ID, data_path: Optional[str] = None, array_index: Optional[int] = None) -> Optional[List[FCurveData]]:
+def get_animation_fingerprint(precision: int, id_data: bpy.types.ID, data_path: Optional[str] = None, array_index: Optional[int] = None) -> Optional[List[FCurveData]]:
 	"""
 	Retrieves the animation data for a specific property or data-block as a serializable fingerprint.
 	"""
@@ -273,12 +273,12 @@ def get_animation_fingerprint(id_data: bpy.types.ID, data_path: Optional[str] = 
 	if data_path:
 		fcurve = next((f for f in action.fcurves if f.data_path == data_path and
 					  (array_index is None or f.array_index == array_index)), None)
-		return [serialize_fcurve(fcurve)] if fcurve else None
+		return [serialize_fcurve(precision, fcurve)] if fcurve else None
 
 	fcurves = sorted(action.fcurves, key=lambda f: (f.data_path, f.array_index))
-	return [serialize_fcurve(f) for f in fcurves]
+	return [serialize_fcurve(precision, f) for f in fcurves]
 
-def get_driver_fingerprint(id_data: bpy.types.ID, data_path: str, array_index: Optional[int] = None) -> Optional[DriverFingerprint]:
+def get_driver_fingerprint(precision: int, id_data: bpy.types.ID, data_path: str, array_index: Optional[int] = None) -> Optional[DriverFingerprint]:
 	"""
 	Captures driver expressions, variables, and modifier stacks into a serializable structure.
 	"""
@@ -336,7 +336,7 @@ def get_driver_fingerprint(id_data: bpy.types.ID, data_path: str, array_index: O
 		m_data: Dict[str, Primitive] = {}
 		for prop in mod.bl_rna.properties:
 			if not prop.is_readonly and prop.identifier not in {'name', 'type', 'is_active'}:
-				m_data[prop.identifier] = to_primitive(getattr(mod, prop.identifier))
+				m_data[prop.identifier] = to_primitive(getattr(mod, prop.identifier), precision)
 		
 		modifiers.append(DriverModifierData(mod_type=mod.type, settings=m_data))
 
@@ -350,7 +350,7 @@ def get_driver_fingerprint(id_data: bpy.types.ID, data_path: str, array_index: O
 
 # --- NODE TREE ANALYSIS ---
 
-def get_material_fingerprint(material: bpy.types.Material, exclude_settings: bool, use_action_names: bool) -> MaterialFingerprint:
+def get_material_fingerprint(material: bpy.types.Material, exclude_settings: bool, use_action_names: bool,	precision: int) -> MaterialFingerprint:
 	"""
 	Generates a unique signature of a material's node tree and render settings to identify duplicates.
 	"""
@@ -367,12 +367,12 @@ def get_material_fingerprint(material: bpy.types.Material, exclude_settings: boo
 
 	if not material.node_tree:
 		mat_type = "FIXED_MAT"
-		diffuse = to_primitive(material.diffuse_color)
-		roughness = to_primitive(material.roughness)
-		metallic = to_primitive(material.metallic)
+		diffuse = to_primitive(material.diffuse_color, precision)
+		roughness = to_primitive(material.roughness, precision)
+		metallic = to_primitive(material.metallic, precision)
 	else:
 		mat_type = "NODE_TREE"
-		node_tree_data = get_node_group_fingerprint(material.node_tree)
+		node_tree_data = get_node_group_fingerprint(material.node_tree, precision)
 
 	# Gather Animation
 	if use_action_names:
@@ -394,7 +394,7 @@ def get_material_fingerprint(material: bpy.types.Material, exclude_settings: boo
 			if prop.is_readonly or prop.identifier in IGNORE_MAT_SETTINGS:
 				continue
 
-			val_data = PropertyEntry(val=to_primitive(getattr(material, prop.identifier)))
+			val_data = PropertyEntry(val=to_primitive(getattr(material, prop.identifier), precision))
 
 			# Check for driver
 			driver = get_driver_fingerprint(precision, material, prop.identifier)
@@ -413,9 +413,9 @@ def get_material_fingerprint(material: bpy.types.Material, exclude_settings: boo
 
 				value = getattr(la, prop.identifier)
 				if hasattr(value, "__iter__") and not isinstance(value, (str, bytes)):
-					value = [to_primitive(v) for v in value]
+					value = [to_primitive(v, precision) for v in value]
 				else:
-					value = to_primitive(value)
+					value = to_primitive(value, precision)
 
 				val_data = PropertyEntry(val=value)
 
@@ -439,7 +439,7 @@ def get_material_fingerprint(material: bpy.types.Material, exclude_settings: boo
 		line_art_settings=line_art_settings
 	)
 
-def get_node_group_fingerprint(node_tree: bpy.types.NodeTree) -> Optional[NodeTreeFingerprint]
+def get_node_group_fingerprint(node_tree: bpy.types.NodeTree, precision: int) -> Optional[NodeTreeFingerprint]:
 	"""
 	Recursively maps the connected node network, properties, and internal group contents.
 	"""
@@ -457,11 +457,11 @@ def get_node_group_fingerprint(node_tree: bpy.types.NodeTree) -> Optional[NodeTr
 			if prop.identifier in IGNORE_PROPS or prop.is_readonly: continue
 			data_path = f'nodes["{node.name}"].{prop.identifier}'
 
-			prop_entry = {"val": to_primitive(getattr(node, prop.identifier))}
+			prop_entry = {"val": to_primitive(getattr(node, prop.identifier), precision)}
 
 			# Check Driver & Anim
-			driver = get_driver_fingerprint(node_tree, data_path)
-			anim = get_animation_fingerprint(node_tree, data_path)
+			driver = get_driver_fingerprint(precision, node_tree, data_path)
+			anim = get_animation_fingerprint(precision, node_tree, data_path)
 			if driver: prop_entry["driver"] = driver
 			if anim: prop_entry["animation"] = anim
 
@@ -477,11 +477,11 @@ def get_node_group_fingerprint(node_tree: bpy.types.NodeTree) -> Optional[NodeTr
 
 			# Scalar
 			if isinstance(default_val, (int, float)):
-				input_data = {"idx": socket_idx, "val": to_primitive(default_val)}
+				input_data = {"idx": socket_idx, "val": to_primitive(default_val, precision)}
 
 				# Check Driver & Anim
-				driver = get_driver_fingerprint(node_tree, data_path)
-				anim = get_animation_fingerprint(node_tree, data_path)
+				driver = get_driver_fingerprint(precision, node_tree, data_path)
+				anim = get_animation_fingerprint(precision, node_tree, data_path)
 				if driver: input_data["driver"] = driver
 				if anim: input_data["animation"] = anim
 
@@ -497,8 +497,8 @@ def get_node_group_fingerprint(node_tree: bpy.types.NodeTree) -> Optional[NodeTr
 						"val": value if isinstance(value, str) else round(value, 6)}
 
 					# Check Driver & Anim
-					driver = get_driver_fingerprint(node_tree, data_path, array_idx)
-					anim = get_animation_fingerprint(node_tree, data_path, array_idx)
+					driver = get_driver_fingerprint(precision, node_tree, data_path, array_idx)
+					anim = get_animation_fingerprint(precision, node_tree, data_path, array_idx)
 					if driver: input_data["driver"] = driver
 					if anim: input_data["animation"] = anim
 
@@ -506,7 +506,7 @@ def get_node_group_fingerprint(node_tree: bpy.types.NodeTree) -> Optional[NodeTr
 
 
 		if node.bl_idname in ('ShaderNodeGroup', 'GeometryNodeGroup') and node.node_tree: # 'GeometryNodeGroup' included for future merge geoNode operator
-			node_data["group_content"] = get_node_group_fingerprint(node.node_tree)
+			node_data["group_content"] = get_node_group_fingerprint(node.node_tree, precision)
 		nodes_data.append(node_data)
 
 	links_data = []
@@ -542,18 +542,18 @@ def count_total_nodes(material: bpy.types.Material) -> int:
 
 	return _recursive_count(material.node_tree)
 
-def to_primitive(val: Any) -> Primitive:
+def to_primitive(val, decimals: int = 4) -> Primitive:
 	"""
 	Converts Blender-specific data types into standard Python primitives for serialization.
 	"""
 	if isinstance(val, (int, float)):
-		return round(val, 6)
+		return round(val, decimals)
 	elif isinstance(val, (str, bool, type(None))):
 		return val
 	elif hasattr(val, "to_list"):
-		return [round(x, 6) for x in val.to_list()]
+		return [round(x, decimals) for x in val.to_list()]
 	elif hasattr(val, "__iter__"):
-		return [to_primitive(x) for x in val]
+		return [to_primitive(x, decimals) for x in val]
 	return str(val)
 
 def trace_socket(socket: bpy.types.NodeSocket) -> bpy.types.NodeSocket:
@@ -648,6 +648,12 @@ class MCPREP_OT_combine_materials(bpy.types.Operator):
 		description="If true, materials with identical animation but different Action names will be kept separate",
 		default=False)
 
+	precision: bpy.props.IntProperty(
+		name="Matching Precision",
+		description="Number of decimal places to round values to when comparing. Higher = stricter",
+		min=0, max=6,
+		default=4)
+
 	def invoke(self, context, event):
 		return context.window_manager.invoke_props_dialog(self)
 
@@ -682,7 +688,7 @@ class MCPREP_OT_combine_materials(bpy.types.Operator):
 
 		for mat in materials_to_check:
 			group_key = util.nameGeneralize(mat.name) if self.group_by_name else "GLOBAL"
-			fp = get_material_fingerprint(mat, self.exclude_mat_settings, self.use_action_names)
+			fp = get_material_fingerprint(mat, self.exclude_mat_settings, self.use_action_names, self.precision)
 
 			if group_key not in fingerprint_groups:
 				fingerprint_groups[group_key] = []
--- a/MCprep_addon/materials/material_manager.py
+++ b/MCprep_addon/materials/material_manager.py
@@ -135,7 +135,7 @@ class ListMaterials(bpy.types.PropertyGroup):
 	# inherited: name
 	description: bpy.props.StringProperty()
 	path: bpy.props.StringProperty(subtype='FILE_PATH')
-	index: bpy.props.IntProperty(min=0, default=0)  # for icon drawing
+	index: bpy.props.IntProperty(min=0, default=0)	# for icon drawing
 
 
 # -----------------------------------------------------------------------------
@@ -323,7 +323,7 @@ def get_animation_fingerprint(precision: int, id_data: bpy.types.ID, data_path: 
 	# Return None if no valid (multi-keyframe) curves were found
 	return results if results else None
 
-def get_driver_fingerprint(precision: int, id_data: bpy.types.ID, data_path: str, array_index: int = 0) -> Optional[DriverFingerprint]:
+def get_driver_fingerprint(precision: int, id_data: bpy.types.ID, data_path: str, array_index: int = -1) -> Optional[DriverFingerprint]:
 	"""Captures drivers math (expression), inputs (variables), and modifiers."""
 
 	if not id_data.animation_data or not id_data.animation_data.drivers:
@@ -348,7 +348,7 @@ def get_driver_fingerprint(precision: int, id_data: bpy.types.ID, data_path: str
 				id_name=getattr(tar.id, "name_full", tar.id.name) if tar.id else ""
 			)
 
-			# 4. Contextual Logic: Only capture attributes relevant to the variable type.
+			# - Contextual Logic: Only capture attributes relevant to the variable type.
 			# This prevents "noise" / irrelevant unused attributes in the driver results
 			
 			# - Property based variables (getting a value from a UI field)
@@ -537,10 +537,29 @@ def get_node_group_fingerprint(node_tree: bpy.types.NodeTree, precision: int) ->
 	# doesn't change if the user re-wired nodes in a different sequence.
 	return NodeTreeFingerprint(nodes=nodes_data, links=sorted(links_data, key=lambda x: str(x)))
 
-def count_total_nodes(material: bpy.types.Material) -> int:
+def node_tree_counter(node_tree: bpy.types.NodeTree) -> int:
 	"""
 	Counts all functional nodes. If a NodeGroup is found, 
 	it enters that sub-tree and adds those nodes to the total count.
+	"""
+	count = 0
+	for node in node_tree.nodes:
+		# Frames and Reroutes are organizational, not functional
+		if node.bl_idname in ('NodeFrame', 'NodeReroute'):
+			continue
+		count += 1
+
+		# If it's a group, count its internal nodes too
+		if node.bl_idname == 'ShaderNodeGroup' and node.node_tree:
+			count += node_tree_counter(node.node_tree)
+	return count
+
+def count_nodes_in_material(material: bpy.types.Material) -> int:
+	"""Wrapper that performs a recursive count of all functional nodes within a material and its nested node groups."""
+	if not material.node_tree:
+		return 0
+	return node_tree_counter(material.node_tree)
+
 def serialize_value(val, decimals: int = 4) -> Primitive:
 	"""
 	Standardizes Blender's internal math types into Python-native primitives.
@@ -568,20 +587,25 @@ def serialize_value(val, decimals: int = 4) -> Primitive:
 		
 	return val
 
-def trace_socket(socket: bpy.types.NodeSocket) -> bpy.types.NodeSocket:
+def trace_socket(dest_socket: bpy.types.NodeSocket) -> bpy.types.NodeSocket:
 	"""Follows a node socket link back to its source, traversing reroute nodes."""
-	Follows a node socket link back to its source, traversing reroute nodes.
-	"""
-	if not socket or not socket.is_linked:
-		return socket
-	current_link = socket.links[0]
+	if not dest_socket or not dest_socket.is_linked:
+		return dest_socket
+	
+	# Start at the first link connected to the destination socket
+	current_link = dest_socket.links[0]
 	source_node = current_link.from_node
+
+	# If the source is a reroute, we need to walk "behind" it recursively
 	while source_node and source_node.bl_idname == 'NodeReroute':
-		if source_node.inputs[0].is_linked:
-			current_link = source_node.inputs[0].links[0]
-			source_node = current_link.from_node
-		else:
-			return source_node.inputs[0]
+		reroute_input = source_node.inputs[0]
+		if not reroute_input.is_linked:
+			return reroute_input
+		
+		current_link = reroute_input.links[0]
+		source_node = current_link.from_node
+
+	# Once we hit a non-reroute node, return the actual source output socket
 	return current_link.from_socket
 
 def get_connected_nodes(node_tree: bpy.types.NodeTree) -> Set[bpy.types.Node]:
@@ -592,18 +616,22 @@ def get_connected_nodes(node_tree: bpy.types.NodeTree) -> Set[bpy.types.Node]:
 	"""
 	if not node_tree:
 		return set()
-	outputNodes = [n for n in node_tree.nodes if n.bl_idname in ('ShaderNodeOutputMaterial', 'NodeGroupOutput')]
+	
+	# Identify the starting output node of the data flow - typically the Material Output node, but also Group Outputs for nested groups
+	outputs = [n for n in node_tree.nodes if n.bl_idname in ('ShaderNodeOutputMaterial', 'NodeGroupOutput')]
 	connected = set()
-	queue = deque(outputNodes)
+	queue = deque(outputs)
+
+	# Breadth-first search through the node tree, following links backwards from outputs to inputs
 	while queue:
 		node = queue.popleft()
-		if node in connected: continue
+		if node in connected:
+			continue
 		connected.add(node)
-		for input_socket in node.inputs:
-			for link in input_socket.links:
-				from_node = link.from_node
-				if from_node not in connected:
-					queue.append(from_node)
+		for socket in node.inputs:
+			for link in socket.links:
+				if link.from_node not in connected:
+					queue.append(link.from_node)
 	return connected
 
 # -----------------------------------------------------------------------------
@@ -668,6 +696,8 @@ class MCPREP_OT_combine_materials(bpy.types.Operator):
 		min=0, max=6,
 		default=4)
 
+	skipUsage: bpy.props.BoolProperty(default=False, options={'HIDDEN'})
+
 	def invoke(self, context, event):
 		return context.window_manager.invoke_props_dialog(self)
 
@@ -682,10 +712,13 @@ class MCPREP_OT_combine_materials(bpy.types.Operator):
 		if self.selection_only:
 			materials_to_check = set()
 			for obj in context.selected_objects:
-				if hasattr(obj.data, "materials"):
-					for mat in obj.data.materials:
-						if mat and not mat.library and not mat.is_library_indirect:
-							materials_to_check.add(mat)
+				if not hasattr(obj.data, "materials"):
+					continue
+				
+				for mat in obj.data.materials:
+					if not mat or mat.library or mat.is_library_indirect:
+						continue
+					materials_to_check.add(mat)
 			materials_to_check = list(materials_to_check)
 		else:
 			materials_to_check = [m for m in bpy.data.materials if not m.library and not m.is_library_indirect]
@@ -728,6 +761,7 @@ class MCPREP_OT_combine_materials(bpy.types.Operator):
 
 				# Sort materials by total node count (add option for user to decide, least or greatest node count. What about number of users?), then by name length as a tie-breaker
 				mats.sort(key=lambda m: (count_total_nodes(m), len(m.name)))
+				mats.sort(key=lambda m: (count_nodes_in_material(m), len(m.name)))
 
 				master_mat = mats[0] # The one with the least nodes
 				for i in range(1, len(mats)):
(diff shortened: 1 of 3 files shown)


BolWaffy:
What is the 'skipUsage' boolean property of the operators, and is it needed?

MCprep_addon/materials/material_manager.py
```diff
@@ -117,6 +135,362 @@ class ListMaterials(bpy.types.PropertyGroup):
 	path: bpy.props.StringProperty(subtype='FILE_PATH')
 	index: bpy.props.IntProperty(min=0, default=0)  # for icon drawing
 
+# -----------------------------------------------------------------------------
+# Material structural comparison helpers
+# -----------------------------------------------------------------------------
+
+# --- SERIALIZATION HELPERS ---
+
+def serialize_fcurve(fcurve: bpy.types.FCurve) -> Dict[str, Any]:
```
StandingPadAnimations: We don't allow the use of `Any` in new code, since it defeats the purpose of type annotations. For this use case, it would be best to use dataclasses with proper type annotations.

MCprep_addon/materials/material_manager.py
```diff
@@ -117,6 +135,362 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+		"extrapolation": fcurve.extrapolation,
+		"auto_smoothing": fcurve.auto_smoothing,
+		"mute": fcurve.mute,
+		"keyframes": [
+			serialize_keyframe(kp)
+			for kp in fcurve.keyframe_points
+		],
+	}
+
+def serialize_keyframe(kp: bpy.types.Keyframe) -> Dict[str, Any]:
```
StandingPadAnimations: Same as earlier, `Any` isn't allowed in new code. A dataclass here would be better.

MCprep_addon/materials/material_manager.py
```diff
@@ -117,6 +135,362 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+		"handle_right_type": kp.handle_right_type,
+		"interpolation": kp.interpolation,
+		"easing": kp.easing,
+		"back": round(kp.back, 6),
+		"amplitude": round(kp.amplitude, 6),
+		"period": round(kp.period, 6),
+	}
+
+# --- ANIMATION & DRIVER LOGIC ---
+def get_animation_fingerprint(id_data: bpy.types.ID, data_path: Optional[str] = None, array_index: Optional[int] = None) -> Optional[Dict[str, Any]]:
```
StandingPadAnimations: Again, dataclasses should be used here instead of a dictionary with `Any`. Also, as a stylistic comment, since the function has three arguments and a long type, some newlines should be inserted to keep things from overflowing.

MCprep_addon/materials/material_manager.py
```diff
@@ -117,6 +135,362 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+	return {"fcurves": [serialize_fcurve(f) for f in fcurves]}
+
+def get_driver_fingerprint(id_data: bpy.types.ID, data_path: str, array_index: Optional[int] = None) -> Optional[Dict[str, Any]]:
+	"""
+	Captures driver expressions, variables, and modifier stacks into a serializable structure.
+	"""
+	if not id_data.animation_data:
+		return None
+
+	fcurve = next((d for d in id_data.animation_data.drivers if d.data_path == data_path and (array_index is None or d.array_index == array_index)), None)
```
StandingPadAnimations: I'm not sure why we're using a generator here? It would be best to simply instantiate a tuple here, and then check if it's empty.

MCprep_addon/materials/material_manager.py
```diff
@@ -117,6 +135,362 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+
+	if data_path:
+		fcurve = next((f for f in action.fcurves if f.data_path == data_path and
+					  (array_index is None or f.array_index == array_index)), None)
+		return serialize_fcurve(fcurve) if fcurve else None
+
+	fcurves = sorted(action.fcurves, key=lambda f: (f.data_path, f.array_index))
+	return {"fcurves": [serialize_fcurve(f) for f in fcurves]}
+
+def get_driver_fingerprint(id_data: bpy.types.ID, data_path: str, array_index: Optional[int] = None) -> Optional[Dict[str, Any]]:
```
StandingPadAnimations: This function should return a dataclass.

StandingPadAnimations (changes requested):
Haven't reviewed the entire thing, since I don't want to end up repeating the same comment in multiple places. The main issue I have with this is the use of `Dict[str, Any]`.

`Any` is just dynamic typing, so any of the guarantees that type annotations should provide cease to exist. For example:

```py
from typing import Any, Dict

data: Dict[str, Any] = {
	"location": (1, 0, 1),
	"rotation": (1.3, 0.2, 10.6)
}

location = data["location"]
rotation = data["rotation"]
```

As far as an LSP, type checker, or any other static analyzer is concerned, `location` and `rotation` are both of the type `Any`, even though `location` should be marked as `Tuple[int, int, int]`, and `rotation` marked as `Tuple[float, float, float]`. No guarantees about the types are given, effectively eliminating any of the benefits. `location.append` for example would be invalid, but an LSP wouldn't catch that because as far as it's concerned, `location` could be any type, including a `List`.

My suggestion would be to use dataclasses here. To give an example:

```py
from dataclasses import dataclass
from typing import Tuple

@dataclass
class ObjectData:
	location: Tuple[int, int, int]
	rotation: Tuple[float, float, float]

data: ObjectData = ObjectData(
	location=(1, 0, 1),
	rotation=(1.3, 0.2, 10.6)
)
```

Here `data.location` is correctly marked as `Tuple[int, int, int]`, and `data.rotation` is correctly marked as `Tuple[float, float, float]`. Thus, when an LSP, type checker, or any other static analyzer goes to review the code and how variables are being used, they're working with proper type guarantees. An LSP will catch the invalid statement `data.location.append`, since it knows the type of `data.location`.

MCprep_addon/materials/material_manager.py
BolWaffy: I was wondering if there was a more efficient one-liner way of finding the first match in Iterable, and found an answer on Stack Overflow.
But now reading the comments, this use of next() is discouraged.
The orginal code was:
```
fcurve: Optional[bpy.types.FCurve] = None

for d in id_data.animation_data.drivers:
	if d.data_path == data_path and (array_index is None or d.array_index == array_index):
		fcurve = d
		break
```

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+@dataclass
+class DriverVariableTarget:
+	id_name: str
+	data_path: str = ""
+	bone_target: str = ""
+	id_type: str = ""
+	transform_type: str = ""
+	transform_space: str = ""
+	rotation_mode: str = ""
+	context_property: str = ""
```
StandingPadAnimations: If these could be empty, it would be best to wrap them in `Optional` and use `None`.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+class PropertyEntry:
+	val: Primitive
+	driver: Optional[DriverFingerprint] = None
+	animation: Optional[List[FCurveData]] = None
+
+class NodeLinkData:
+	from_socket: str
+	from_node: str
+	to_socket: str
+	to_node: str
```
StandingPadAnimations, line 223: Looks like this is missing the `@dataclass` decorator

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+
+@dataclass
+class MaterialFingerprint:
+	mat_type: str = ""
+	diffuse: Optional[List[float]] = None
+	roughness: Optional[float] = None
+	metallic: Optional[float] = None
+	node_tree_data: Optional[NodeTreeFingerprint] = None
+	mat_action_name: str = 'None'
+	node_tree_action_name: str = 'None'
```
StandingPadAnimations: Clarify what values `mat_action_name` and `node_tree_action_name` could be set to? If there's a limited set of specific string values they could be, [enums](https://docs.python.org/3/library/enum.html) would be a better option.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+# -----------------------------------------------------------------------------
+# Material structural comparison helpers
+# -----------------------------------------------------------------------------
+
+# --- SERIALIZATION HELPERS ---
+
+def serialize_fcurve(precision: int, fcurve: bpy.types.FCurve) -> FCurveData:
+	"""
+	Converts an F-Curve data-block into a dictionary of primitive values for structural comparison.
+	"""
```
StandingPadAnimations, line 266: Style comment: single-line doc strings should just be done as `"""Some doc-string"""`.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
 
 
+# -----------------------------------------------------------------------------
+# Dataclasses for material comparison 
+# -----------------------------------------------------------------------------
+
+Primitive = Union[int, float, str, bool, list, None]
+
+@dataclass
+class KeyframeData:
```
StandingPadAnimations, line 159: Dataclasses should have some doc-string summarizing what they represent.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+		dynamics=(round(kp.back, precision), round(kp.amplitude, precision), round(kp.period, precision))
+	)
+
+# --- ANIMATION & DRIVER LOGIC ---
+def get_animation_fingerprint(precision: int, id_data: bpy.types.ID, data_path: Optional[str] = None, array_index: Optional[int] = -1) -> Optional[List[FCurveData]]:
+	"""
+	Retrieves the animation data for a specific property or data-block as a serializable fingerprint.
+	"""
+	if not id_data.animation_data or not id_data.animation_data.action:
+		return None
```
StandingPadAnimations: `return None` can be simplified to just `return`.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+			elif var.type == 'TRANSFORMS':
+				tar_info["transform_type"] = tar.transform_type
+				tar_info["transform_space"] = tar.transform_space
+				if tar.transform_type.startswith("ROT"):
+					tar_info["rotation_mode"] = tar.rotation_mode
+
+			elif var.type == 'LOC_DIFF':
+				tar_info["transform_space"] = tar.transform_space
+
+			targets.append(DriverVariableTarget(**tar_info))
```
StandingPadAnimations: This form of dataclass creation makes me skeptical, since there would be no warnings if a typo was made here, and it's just overly verbose. It would be better to create the `DriverVariableTarget` object ahead of time, and then assign each of the attributes directly.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+	
+	# 4. F-Curve Modifiers (Skip if muted or influence is 0)
+	modifiers: List[DriverModifierData] = []
+	for mod in fcurve.modifiers:
+		if mod.mute or getattr(mod, "influence", 1.0) <= 0.0:
+			continue
+
+		m_data: Dict[str, Primitive] = {}
+		for prop in mod.bl_rna.properties:
+			if not prop.is_readonly and prop.identifier not in {'name', 'type', 'is_active'}:
```
StandingPadAnimations: Dictionary syntax (`{}`) shouldn't be used when comparing something to a list of possible values, Tuple syntax (`()`) should be used for that.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+	mat_type = "FIXED_MAT"
+	diffuse = None
+	roughness = None
+	metallic = None
+	node_tree_data = None
+	mat_action_name = 'None'
+	node_tree_action_name = 'None'
+	anim = None
+	render_settings = None
+	line_art_settings = None
```
StandingPadAnimations: To avoid creating these additional variables, it would be simpler to create an instance of `MaterialFingerprint` and set the attributes directly. That would of course require changing much of `MaterialFingerprint`'s definition to have default values, to make things simpler.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+		node_tree_data = get_node_group_fingerprint(material.node_tree, precision)
+
+	# Gather Animation
+	if use_action_names:
+		# Material level action
+		mat_action_name = material.animation_data.action.name if (material.animation_data and material.animation_data.action) else 'None'
+
+		# Node Tree level action
+		if material.node_tree and material.node_tree.animation_data and material.node_tree.animation_data.action:
+			node_tree_action_name = material.node_tree.animation_data.action.name
```
StandingPadAnimations: To avoid excessive overflow with the `material.animation_data and material.animation_data.action` condition, it should be extracted to its own variable.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+		else:
+			node_tree_action_name = 'None'
```
StandingPadAnimations: As noted in an earlier comment, [enums](https://docs.python.org/3/library/enum.html) are better for a limited set of values.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+		animation=anim,
+		render_settings=render_settings,
+		line_art_settings=line_art_settings
+	)
+
+def get_node_group_fingerprint(node_tree: bpy.types.NodeTree, precision: int) -> Optional[NodeTreeFingerprint]:
+	"""
+	Recursively maps the connected node network, properties, and internal group contents.
+	"""
+	if not node_tree: return None
```
StandingPadAnimations: Please don't use these single-line if statements, they're out of place.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+	active_nodes = get_connected_nodes(node_tree)
+	nodes_data = []
+
+	for node in sorted(list(active_nodes), key=lambda n: (n.bl_idname, n.name)):
+		if node.bl_idname in {'NodeFrame', 'NodeReroute'}: continue
+
+		node_data = {"type": node.bl_idname, "muted": node.mute, "properties": {}, "inputs": []}
```
StandingPadAnimations: This (and the rest of this function) are extremely verbose, and it's hard to tell what's exactly going on here. I would recommend adjusting this to avoid the verbosity of dictionaries.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+
+		if node.bl_idname in ('ShaderNodeGroup', 'GeometryNodeGroup') and node.node_tree: # 'GeometryNodeGroup' included for future merge geoNode operator
+			node_data["group_content"] = get_node_group_fingerprint(node.node_tree, precision)
+		nodes_data.append(node_data)
+
+	links_data = []
+	for node in active_nodes:
+		if node.bl_idname in {'NodeFrame', 'NodeReroute'}: continue
+		for socket in node.inputs:
+			if socket.is_linked:
```
StandingPadAnimations: This condition can be inverted to reduce excessive nesting.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+		for node in node_tree.nodes:
+			if node.bl_idname in {'NodeFrame', 'NodeReroute'}:
+				continue
+			count += 1
+			# If it's a group, count its internal nodes too
+			if node.bl_idname in ('ShaderNodeGroup', 'GeometryNodeGroup') and node.node_tree:
+				count += _recursive_count(node.node_tree)
+		return count
+
+	return _recursive_count(material.node_tree)
```
StandingPadAnimations: I would recommend moving this function out into the global scope with a name similar to `node_tree_count`, and rename the parent function to `count_nodes_in_material`. Also, the argument for `node_tree` should be annotated with the proper type.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+		return 0
+
+	def _recursive_count(node_tree):
+		count = 0
+		for node in node_tree.nodes:
+			if node.bl_idname in {'NodeFrame', 'NodeReroute'}:
+				continue
+			count += 1
+			# If it's a group, count its internal nodes too
+			if node.bl_idname in ('ShaderNodeGroup', 'GeometryNodeGroup') and node.node_tree:
```
StandingPadAnimations: Since this is dealing with shader nodes, checking for a geometry node group doesn't make sense.

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
 	index: bpy.props.IntProperty(min=0, default=0)  # for icon drawing
 
 
+# -----------------------------------------------------------------------------
+# Dataclasses for material comparison 
+# -----------------------------------------------------------------------------
+
+Primitive = Union[int, float, str, bool, list, None]
```
StandingPadAnimations: `list` isn't allowed as a type annotation until Python 3.9, and all lists must define an inner type (i.e. is it a list of ints, floats, etc?)

MCprep_addon/materials/material_manager.py
```diff
@@ -118,6 +138,465 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+	"""
+	if isinstance(val, (int, float)):
+		return round(val, decimals)
+	elif isinstance(val, (str, bool, type(None))):
+		return val
+	elif hasattr(val, "to_list"):
+		return [round(x, decimals) for x in val.to_list()]
+	elif hasattr(val, "__iter__"):
+		return [to_primitive(x, decimals) for x in val]
+	return str(val)
```
StandingPadAnimations: A couple of things:

- Integers don't need rounding, so that should be broken out into its own thing
- `None` can be checked using the `val is None` condition

More broadly though, is there a specific reason Blender's datatypes need to be converted to Python primitives? This isn't be outputted as a JSON file and whatnot from what I can see, so conversion shouldn't be needed. If this is just analyzing node-trees in Python, the analysis should be written around Blender's datatypes, not the other way around.

MCprep_addon/materials/material_manager.py
```diff
@@ -151,132 +630,130 @@ def execute(self, context):
...
-		# materials blocks with the same base
-		name_cat = {}
+		# 1. Gather materials
+		if self.selection_only:
+			materials_to_check = set()
+			for obj in context.selected_objects:
+				if hasattr(obj.data, "materials"):
+					for mat in obj.data.materials:
+						if mat and not mat.library and not mat.is_library_indirect:
+							materials_to_check.add(mat)
```
StandingPadAnimations: One or more of these conditions should be inverted to reduce nesting.

StandingPadAnimations (changes requested):
Did some skimming with the changes. My main issue at the moment is the verbosity, mainly that it's extremely hard to follow along with everything. A couple of broad suggestions:

- I wouldn't worry too much about the amount of lines something takes. It seems like in many places, there's a push towards trying to keep line counts as low as possible, and while that's understandable, IMO it does lead to verbosity.
	- If something is overly verbose, we'll comment on that in the review
- In a couple of places, there are some uses of dictionaries that could be avoided. My issue with dictionaries is that they're extremely verbose, especially in situations where dataclasses could be used directly, and they don't have proper type guarantees<sup>[1](#footnote1)</sup>.
	- More broadly, I think the main issue is not taking proper advantage of dataclasses. Dataclasses (well at least non-frozen ones, like what we have here) can have their attributes reassigned at any point, so there's no need to create a temporary dictionary or a set of temporary variables, you can just instantiate a dataclass ahead of time, reassign attributes where needed, and return that object at the end.
- From what I can tell, there's no stage at which the material is outputted to say a JSON file, so is there a reason to convert Blender's datatypes? After all, if it's just analysis, and it's all being done without being converted to a file, I don't see a reason to convert datatypes.


<a name="footnote1">1</a>: `TypedDict` in theory would give those guarantees, but that's a Python 3.8 feature, and MCprep targets Python 3.7

MCprep_addon/materials/material_manager.py
StandingPadAnimations: I wouldn't worry about condensing things to one-liners right now. If the length of something becomes an issue, it will be brought up in review, but otherwise it doesn't matter.

MCprep_addon/materials/material_manager.py
BolWaffy: That makes sense. The JSON handling was carried over from the original script I wrote that this was based on, but in this use case, those conversions are unnecessary. I'll refactor it to use Blender's types. Thanks

MCprep_addon/materials/material_manager.py
```diff
@@ -133,363 +135,503 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+@dataclass
+class NodeTreeFingerprint:
+	"""Stores the list of nodes and links within a node tree."""
+	nodes: List[NodeFingerprint]
+	links: List[NodeLinkData]
+
+@dataclass
+class MaterialFingerprint:
+	"""Stores a unique signature of a material for structural comparison."""
+	mat_type: str = "FIXED_MAT"
```
StandingPadAnimations: Is `mat_type` limited to a couple of values? If so, it would be better to use an enum.

MCprep_addon/materials/material_manager.py
```diff
@@ -133,363 +135,503 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+	links_data = [
+		NodeLinkData(
+			from_socket=(source_socket := trace_socket(dest_socket)).name,
+			from_node=source_socket.node.bl_idname,
+			to_socket=dest_socket.name,
+			to_node=node.bl_idname)
+
+		for node in active_nodes if node.bl_idname not in ('NodeFrame', 'NodeReroute')
+		for dest_socket in node.inputs if dest_socket.is_linked and (source_socket := trace_socket(dest_socket)).node in active_nodes
+	]
```
StandingPadAnimations: I would move this out into an actual loop. It's weird in its current state.

MCprep_addon/materials/material_manager.py
```diff
@@ -133,363 +135,503 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+
+	# Sorting links by a string representation ensures the link order 
+	# doesn't change if the user re-wired nodes in a different sequence.
+	return NodeTreeFingerprint(nodes=nodes_data, links=sorted(links_data, key=lambda x: str(x)))
```
StandingPadAnimations: From what I understand, `links_data` is supposed to be a `List[NodeLinkData]`? In that case, you could define `NodeLinkData` with [`order` set to `True`](https://docs.python.org/3/library/dataclasses.html#dataclasses.dataclass), which will define the necessary methods needed for `sorted` to work natively with the class.

MCprep_addon/materials/material_manager.py
```diff
@@ -133,363 +135,503 @@ class ListMaterials(bpy.types.PropertyGroup):
...
-			count += 1
-			# If it's a group, count its internal nodes too
-			if node.bl_idname in ('ShaderNodeGroup', 'GeometryNodeGroup') and node.node_tree:
-				count += _recursive_count(node.node_tree)
-		return count
-
-	return _recursive_count(material.node_tree)
-
-def to_primitive(val: Any) -> Union[int, float, str, bool, list, None]:
+def serialize_value(val, decimals: int = 4) -> Primitive:
```
StandingPadAnimations: Do we still need to convert types? From what I can tell, we're not writing to an external file, so why not just use the native types without conversion?

MCprep_addon/materials/material_manager.py
```diff
@@ -133,363 +135,503 @@ class ListMaterials(bpy.types.PropertyGroup):
...
 	path: bpy.props.StringProperty(subtype='FILE_PATH')
-	index: bpy.props.IntProperty(min=0, default=0)  # for icon drawing
+	index: bpy.props.IntProperty(min=0, default=0)	# for icon drawing
+
+
+# -----------------------------------------------------------------------------
+# Dataclasses for material comparison 
+# -----------------------------------------------------------------------------
+
+Primitive = Union[int, float, str, bool, Tuple['Primitive', ...], None]
```
StandingPadAnimations: Can `Tuple[Primative, ...]` contain anther tuple (i.e. are nested tuples possible)? If not, I would split this into `PrimativeBasic` (which has the base types) and `Primative` (a union of `PrimativeBasic` and `Tuple[PrimativeBasic, ...]`).

MCprep_addon/materials/material_manager.py
```diff
@@ -133,363 +135,503 @@ class ListMaterials(bpy.types.PropertyGroup):
...
 		connected.add(node)
-		for input_socket in node.inputs:
-			for link in input_socket.links:
-				from_node = link.from_node
-				if from_node not in connected:
-					queue.append(from_node)
+		for socket in node.inputs:
+			for link in socket.links:
+				if link.from_node not in connected:
+					queue.append(link.from_node)
```
StandingPadAnimations: I would invert that last condition to reduce nesting in this loop.

StandingPadAnimations (commented):
Currently still sorting things with my living situation (long story), but I've done a basic skim of the changes made since the last review (though I'll leave the in-depth review to @TheDuckCow). From what I was able to gather while skimming the changes, it looks mostly good, my comments are mainly with some of the more obvious issues.

TheDuckCow:
Yup I'm fine to do the in depth review - @BolWaffy I see you have two open tick box options:

- For clearing animation if you think this will be helpful for you, that's fine to have as a default off option - though how are you planning to surface "options" since there's not really an undo workflow for this? Or perhaps it just works
- The choice by popularity/user count is definitely one I support

Neither are blocking and we could always do as incremental improvement if you're looking to just wrap this up. I'll take a  deeper review tomorrow.

TheDuckCow:
(Marking as ready for review, as that's what we're doing at this stage)

MCprep_addon/materials/material_manager.py
TheDuckCow, line 1: Feedback from my functional review, I think these minor UX changes will make it easier to understand to someone unfamiliar with the feature:

- In both cases, let's put the "selection only" tick box at the top, default on is still good
- Combine materials:
  - The options here feel a bit more confusing. Suggesting we align on the "Compare..." terminology
  - Exclude Material Settings -> Let's invert it so it's "positive" framed, ie "Compare material settings", and if it needs to be on by default then that's fine.
  - "Group method: Material Name" can be changed to "Compare material name"
  - Then "Differentiate by Action Name" -> "Compare Action Name"
  - then "Matching Precision" -> "Compare fcurves", though I wonder if this should just be an enum with values of "Off", "Loose", and "Strict" so that users don't have to hover and guess what 2 or 6 means. Off would be not comparing curves at all (just animation block name, if enabled right?), Loose could be a value of e.g. 1, and strict could be a value of e.g. 6. This is reducing the options in favor of ease of use, what do you think? And this way it's also more clear when it's off (whereas, I don't know what it would mean if it says 4 or 6 for fcurve but then I'm not comparing by actions anyways).
- Combine images:
  - "Group by name" -> "Compare Image Name"
  - Rename "Strict Comparison" to "Compare Pixels (slower)", tooltip value is fine.
  - Let's default to also have selection on by default, so it's consistent with materials (or, materials should show off by default, but I like the idea of starting more conservatively and letting the user expand it)

MCprep_addon/materials/material_manager.py
TheDuckCow: I was going to say: typing here is getting rather complex, perhaps we should just use Any at this rate, it's not that helpful for the developer if we need to abstract this so much (at the very least, I'm not worried about changing it further).

MCprep_addon/materials/material_manager.py
TheDuckCow: Fwiw though I also advocate for reducing indents, in this case you still end up with a continue line as intended, so I'd say an optional change since there's still net 1 LOC indented at that level.

MCprep_addon/materials/material_manager.py
```diff
@@ -151,129 +667,131 @@ def execute(self, context):
...
 	bl_idname = "mcprep.combine_materials"
 	bl_label = "Combine materials"
-	bl_description = (
-		"Consolidate the same materials together e.g. mat.001 and mat.002")
+	bl_description = "Consolidate duplicate materials based on nodes or name"
 	bl_options = {'REGISTER', 'UNDO'}
 
-	# arg to auto-force remove old? versus just keep as 0-users
+	group_by_name: bpy.props.BoolProperty(
+		name="Group Method: Material Name",
```
TheDuckCow: See my naming feedback at the top of the file, and slight re-ordering. Sorry for being picky, but these little renaming will help a lot to new users. Descriptions are mostly fine, just refine if needed (e.g. where I suggested inverting the logic)

TheDuckCow (approved):
I'm going to give my approval in that any remaining changes that are critical I'd be happy to take over. If you want to throw the towl in @BolWaffy I'd be happy to push these changes myself, though see my first comment where I leave some UX feedback and ask one question. Please just let me know either way so we don't double up and both try to make the changes.

As a side note, we're adding non trivial complexity here. That's fine, but a note to self that we'll want to improve our test coverage over these different areas of functionality to make sure things continue to work (e.g. creating tests that compare when we have different cases of merging materials together base don adding nodes or changing settings etc). Thankfully the tests did at least cover some issue for the defaults applied to these merging options, but expanded tests might reveal other cases of syntax unsupported in older blender (python) versions,

In the meantime, I at least confirmed in the meantime tests are all passing with the latest code and latest blender versions - only really two kinds of failures for the older blender versions, with multiples of them:

```
(5.0.1)   	all_tests	75	2	0	No errors
(4.5.3)   	all_tests	75	2	0	No errors
(4.3.1)   	all_tests	75	2	0	No errors
(4.2.3)   	all_tests	75	2	0	No errors
(4.0.2)   	all_tests	75	2	0	No errors
(3.6.2)   	all_tests	75	2	0	No errors
(3.5.1)   	all_tests	75	2	0	No errors
(3.1.0)   	all_tests	75	2	0	No errors
(3.0.0)   	all_tests	75	2	0	No errors
(2.93.0)   	all_tests	6	0	7	addon_test;asset_qa_test;materials_test;spawner_test;tracker_test;UtilOperatorsTest);world_tools_test
(2.80.75)   	all_tests	6	0	7	addon_test;asset_qa_test;materials_test;spawner_test;tracker_test;UtilOperatorsTest);world_tools_test
Compiled in 42.3s + tests ran in 226.7s
Total of 268.9s with exit code 1
```

Detailing on failures:

Some are due to numpy typing not being available, I'd prefer to skip typing if it would prevent the tool to work properly.

```
ERROR: asset_qa_test (unittest.loader._FailedTest)
ImportError: Failed to import test module: asset_qa_test
  File "/Applications/blender 2.93/Blender.app/Contents/Resources/2.93/python/lib/python3.9/unittest/loader.py", line 436, in _find_test_path
    module = self._get_module_from_name(name)
  File "/Applications/blender 2.93/Blender.app/Contents/Resources/2.93/python/lib/python3.9/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/Users/patrickcrawford/Documents/gits/mcprep/test_files/asset_qa_test.py", line 25, in <module>
    from MCprep_addon.spawner import spawn_util
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/__init__.py", line 59, in <module>
    from . import load_modules
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/load_modules.py", line 43, in <module>
    from .materials import material_manager
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/materials/material_manager.py", line 27, in <module>
    from numpy.typing import NDArray
ModuleNotFoundError: No module named 'numpy.typing'
```

This syntax isn't supported apparently, so let's just remove the colon and just do assignment to support Pre 3.0 blender versions:

```
Error: Traceback (most recent call last):
  File "/Applications/blender 2.80/Blender.app/Contents/Resources/2.80/scripts/modules/addon_utils.py", line 351, in enable
    mod = __import__(module_name)
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/__init__.py", line 59, in <module>
    from . import load_modules
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/load_modules.py", line 43, in <module>
    from .materials import material_manager
  File "/Users/patrickcrawford/Documents/gits/mcprep/MCprep_addon/materials/material_manager.py", line 527
    from_socket=(source_socket := trace_socket(dest_socket)).name,
                               ^
SyntaxError: invalid syntax
```

These two errors I think are all the automated tests. Again, if you don't want to deal with these, I'd be happy to resolve myself. But these are changes we should make before merging.

BolWaffy:
Sorry about the close/reopen accident.

@StandingPadAnimations Read your long story, all good.

@TheDuckCow Thanks for the offer to take over, but I'd definitely like to apply these edits and finish this up myself. Its a great chance for me to learn more.
I'll tackle those UX naming changes, the logic flips, and the edits StandingPadAnimations suggested. I'll ping you when it's ready.

TheDuckCow:
> Sorry about the close/reopen accident.
> 
> @StandingPadAnimations Read your long story, all good.
> 
> @TheDuckCow Thanks for the offer to take over, but I'd definitely like to apply these edits and finish this up myself. Its a great chance for me to learn more. I'll tackle those UX naming changes, the logic flips, and the edits StandingPadAnimations suggested. I'll ping you when it's ready.

No worries, I can definitely appreciate taking it as a learning opportunity! Just don't want to hold it up unnecessary, but no need to rush from my standpoint. When you want either of us to review again, just press the refresh arrow under our names in the Reviewers section.

BolWaffy:
Forgot to mention, I wanted to clarify the purpose of the "Matching Precision" setting (now called "Value Matching").

While it does affect f-curves, its primary purpose is to merge near identical materials by rounding all float values in the material, including node input properties like Roughness, Metallic, Colors, Vectors, etc. It basically acts as a "Fuzzy Matching" threshold.

For example:

* Material A Diffuse: 0.15
* Material B Diffuse: 0.15245

With **Value Matching** set, these values are rounded and compared as identical ($0.15 = 0.15$), allowing the script to merge those materials. This helps prevent duplicate materials from staying separate due to negligible floating-point differences.

I also finalized adding the "**Material to Keep**" setting so users can choose which material becomes the "Master."

StandingPadAnimations:
I feel we could rename "Value Matching" to "Value Rounding", and have it be an integer representing the amount of numbers behind the decimal. That way, its behavior is more clear to the end user (and it becomes more flexible).

TheDuckCow:
> I feel we could rename "Value Matching" to "Value Rounding", and have it be an integer representing the amount of numbers behind the decimal. That way, its behavior is more clear to the end user (and it becomes more flexible).

I'm still not sure if Value Rounding is intuitive when I see the options to put 0, 1, or 6. If I'm thinking like "merge by distance", then I'd assume 0 is only for perfect matches. 

I do like the idea of just making it the number of decimals though, so maybe we could phrase it: "Fcurve decimal rounding". Then it's very explicit, better than my "loose" vs "strict" suggestion before.

Edit: also I see we had [these failures](https://github.com/Moo-Ack-Productions/MCprep/actions/runs/22309806290/job/64538478722?pr=703) in the last run, do the semantics of the tests need updating or is something actually wrong now?

> (4.1.1)   	all_tests	75	2	2	test_combine_images;test_combine_images_strict_comparision

BolWaffy:
Hey @TheDuckCow , just checking in on this PR. Let me know if you need any changes to get this merged.

TheDuckCow:
> Hey @TheDuckCow , just checking in on this PR. Let me know if you need any changes to get this merged.

So sorry, a lot of life has been happening. Appreciate the nudge, doing my final review today for you

TheDuckCow (approved):
I'm approving as I've done my testing, and did a run through of the more recent changes. I will say I believe the typing, since it's more for developer benefit, has gotten more intricate to be useful (e.g. instead of Primitive, we should have just been happy enough with Any if we weren't really going to have a narrow typing response). But, since the work is already done, it's fine to stay.

The only changes I'd make (which I'm going to say are optional for you, if we merge then I'll do them in a fast-follow):
- Default "Compare Material Name" to on to be similar to old behavior and similar to the default Combine Images options
- Based on the scenario below, I feel like it's a bit unintuitive that materials with different image blocks aren't merged together. I believe previously that was the default behavior, and would result in 0-reference images that would get cleaned up later, and would let the combine materials "just work" right away. But now, you essentially have to run the combine images first, and _then_ attempt combine materials to actually get them to merge down.
  - So, the change I'd consider is whether we have another tick box on Combine Materials for "Compare image blocks", and basically have this control whether it considers the material unique based on image blocks or not. If off, then it just merges materials based on names and not the image blocks

Curious what you think of the proposal above (including from @StandingPadAnimations).  Main thing is to make sure users who are accustomed to using Combine Materials can generally do so w the same as before, and not be left thinking it's broken whereas really it's just being more specific now by checking the image blocks (hence the tick box).

Edit: demo video of what I mean / an easy way to reproduce:

https://github.com/user-attachments/assets/6214a820-bf6a-437d-954c-45d6c349bb39

BolWaffy:
Running `combine images` first, then `combine materials` was something I also found unintuitive. 

To make sure I'm on the same page regarding the 'Compare image blocks' toggle, are you envisioning:

1.  **Attribute Masking:** Keeping the current logic but simply ignoring the image data during the material 'fingerprinting' phase (e.g., dynamically adding `image` to `IGNORE_PROPS`)? This would treat materials as identical even if their image nodes point to different files.
2.  **Chained Operators:** Actually triggering the `combine_images` operator logic *inside* the combine materials execution if the toggle is set? 

The second option seems closer to the old behavior you mentioned, as it actively cleans up the data. If we go with **Option 2**, the implementation would look something like this:

```python
# Inside MCPREP_OT_combine_materials operator:
def execute(self, context):
    if self.selection_only and not context.selected_objects:
        self.report({'ERROR'}, "Select objects with materials first")
        return {'CANCELLED'}

    # Trigger image combining first to consolidate the data
    if self.combine_images:
        bpy.ops.mcprep.combine_images(
            'EXEC_DEFAULT',
            selection_only=self.selection_only,
            group_by_name=self.compare_name,
            strict_comparison=True
        )

    # < 0 means Exact Matching (no rounding).
    precision_val = float('inf') if self.value_rounding < 0 else self.value_rounding

    # 1. Gather materials...
```

MCprep_addon/materials/material_manager.py
```diff
@@ -115,9 +135,550 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+				fp.animation[f.data_path] = []
+			fp.animation[f.data_path].append(f)
+	else:
+		fp.animation = None
+
+	# 4. Gather Settings and Line Art
+	if compare_settings:
+		# Material Render Settings
+		fp.render_settings = {
+			prop.identifier: extract_property(getattr(material, prop.identifier), material, prop.identifier, precision)
```
StandingPadAnimations: This and line `474` should have the function parameters spread across multiple lines, to avoid overflowing text

MCprep_addon/materials/material_manager.py
```diff
@@ -115,9 +135,550 @@ class ListMaterials(bpy.types.PropertyGroup):
...
+
+	# Sort: We sort by type and name so that the order of nodes in
+	# the list is always identical (deterministic) for the same setup.
+	for node in sorted(active_nodes, key=lambda n: (n.bl_idname, n.name)):
+		if node.bl_idname in ('NodeFrame', 'NodeReroute'):
+			continue
+
+		# Capture Attributes: Every input property, slider, checkbox, enum, etc. on the node.
+		properties = {
+			p.identifier: extract_property(getattr(node, p.identifier), node_tree, f'nodes["{node.name}"].{p.identifier}', precision)
```
StandingPadAnimations: This (and line `507`) should also be spread across multiple lines for overflow reasons.

MCprep_addon/materials/material_manager.py
```diff
@@ -151,129 +712,167 @@ def execute(self, context):
...
+		for old_mat, master_mat in merge_map.items():
+			if master_mat not in renamed_masters:
+				new_name = util.nameGeneralize(master_mat.name)
+				if new_name != master_mat.name and new_name not in bpy.data.materials:
+					master_mat.name = new_name
+				renamed_masters.add(master_mat)
+			env.log(f"Replaced '{old_mat.name}' with '{master_mat.name}'")
+			old_mat.user_remap(master_mat)
+
+		to_delete = [m for m in merge_map.keys() if m.users == 0 and not m.use_fake_user]
```
StandingPadAnimations: The check for `m.users == 0` can be simplified to `not m.users`.

MCprep_addon/materials/material_manager.py
```diff
@@ -151,129 +712,167 @@ def execute(self, context):
...
+		if to_delete:
+			bpy.data.batch_remove(ids=to_delete)
+
+		postcount = len(bpy.data.materials)
+		consolidated_count = precount - postcount
 
+		if consolidated_count > 0:
+			self.report({"INFO"}, f"Consolidated {consolidated_count} material{'s' if consolidated_count > 1 else ''} (Total: {precount} -> {postcount})")
```
StandingPadAnimations: The part that makes `material` plural could be reduced down to `'s'[:consolidated_count^1]`. Granted, the syntax is a strange for anyone not aware of XOR and string slicing, but it would be shorter.

StandingPadAnimations (approved):
All the comments I have are style related. Otherwise, I'd be fine merging this in.

TheDuckCow:
Thanks for re-requesting, will review today - and sorry for my delayed reply [over here](https://github.com/Moo-Ack-Productions/MCprep/pull/703#issuecomment-4162795215):

I'm thinking more in line with option 2. Option one feels like a good idea in a way, but will gloss over that different materials might have different images of course and if we already have the benefit of the pre-existing operator, may as well use that (with its default settings). I see that's what you put in your PR, so let's go with this. I'll review it today to see if I have any different thoughts about the settings to apply.

TheDuckCow (approved):
Approving and merging shortly - have a very small diff I'll apply in a post merge commit (since the upstream doesn't have source repo contrib access). Thanks for bearing with us on the back and forths - I know it took awhile but I hope it was a valuable experience. If there are any other areas you are curious to get involved with, be sure to let us know so we can chat and set you up for success.

PR #704, opened 2026-03-06
Implement internal support for layered resource packs

StandingPadAnimations:
(Currently a draft)

This PR aims to implement support for layered resource packs internally; in other words, support layered resource packs in material generation, but without a proper UI to add layers (reserved for a future PR). The goal is to make it so that, internally, material generation can handle multiple textures, with texture swap being functionally equivalent to having two resource packs in Minecraft (the custom resource pack, and the default resource pack as a fallback). That way, when the time does come for implementing the UI/UX side of things, the workload should be lighter overall.

### DIFF
--- a/MCprep_addon/materials/generate.py
+++ b/MCprep_addon/materials/generate.py
@@ -27,7 +27,7 @@ from bpy.types import Context, Material, Image, Texture, Nodes, NodeLinks, Node
 
 from .. import util
 from ..conf import MCprepError, env, Form
-from .resource_pack import find_from_texturepack
+from .resource_pack import MCResourcePack, find_texture_from_layers, get_default_pack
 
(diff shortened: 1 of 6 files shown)


StandingPadAnimations:
Huh, according to the tests, in several places `MCprepError` is being passed to the texture finding function, but in all of those places, an explicit check for `MCprepError` is done with an early return, so it shouldn't be happening. Looks like something for me to investigate tomorrow.

StandingPadAnimations:
I'll have to dive back into this code to figure out where I left off and see where I still need to work on, but for now, I think the core code is in a reviewable state

TheDuckCow:
Good to note - I'm thinking a bit more about the sequencing  of things, but probably won't fully revisit this in earnest until we have more of the world importing in progress.

PR #710, opened 2026-05-27
World Bridge - May 2026

StandingPadAnimations:
This PR tracks the development of the world bridge for May 2026. For the most part, this is a proof of concept.

Goals for this PR:
- [x] Create interface code for calling exporters
	- At the moment, we're only going to support jmc2OBJ, since A. that's what I have on my system and B. Mineways script creation adds a bit of additional complexity
- [x] Basic reloading of imported OBJS
	- Starting with reloading since that's the simplest to implement; at most, we'd need one additional property on OBJs in the form of the OBJ path
	- [ ] ~~Options between `Back-up and Reload` vs `Replace`~~
	- [ ] ~~Chunked import support (both for individual chunks and full assembly)~~

***Breaking***: This PR raises the minimum Blender requirements to Blender 4.2 (and by extension, Python 3.11).


MCprep_addon/exporter_glue/jmc2obj.py
```diff
@@ -0,0 +1,84 @@
...
+    # or otherwise figure out a way to determine if the original
+    # OBJ was centered or used an arbitrary offset
+    @override
+    def set_export_offset(self, offset: tuple[float, float, float]) -> None | MCprepError:
+        # Since there's no way to have arbitrary Y offsets
+        # with jmc2OBJ, it's better to alert the end-user
+        if offset[1] != 0:
+            line, file = env.current_line_and_file()
+            return MCprepError(NotImplementedError(), line, file, "jmc2OBJ cannot take an arbitrary Y offset!")
+        self.args.append(f"--offset={offset[0]},{offset[2]}")
```
StandingPadAnimations: jmc2OBJ seems to only allow arbitrary X and Z offsets. Will need to figure out how to distinguish centering from user-defined offsets on the X and Z axis.

TheDuckCow:
> Breaking: This PR raises the minimum Blender requirements to Blender 4.3 (and by extension, Python 3.11).

MCprep v4.0, here we come!

StandingPadAnimations:
> > Breaking: This PR raises the minimum Blender requirements to Blender 4.3 (and by extension, Python 3.11).
> 
> MCprep v4.0, here we come!

If a 14 version jump doesn't meet the criteria for a super major bump, idk what will 😄

pyproject.toml
```diff
@@ -8,7 +8,7 @@ readme = "README.md"
 package-mode = false
 
 [tool.poetry.dependencies]
-python = "^3.9"
+python = "^3.10"
```
TheDuckCow: Also wouldn't this be 3.11?

MCprep_addon/_blender_manifest.toml
```diff
@@ -11,7 +11,7 @@ website = "https://TheDuckCow.com/MCprep"
 
 tags = ["Animation", "Material"]
 
-blender_version_min = "4.2.0"
+blender_version_min = "4.3.0"
```
TheDuckCow:
```suggestion
blender_version_min = "4.2.0"
```

Hopeful this is possible

pyproject.toml
StandingPadAnimations: Whops, I'll fix that real quick

MCprep_addon/_blender_manifest.toml
StandingPadAnimations: Looks like 4.2 is possible

StandingPadAnimations:
Since May is already over, I've marked the back-up and chunk import support parts for a later PR. Chunked imports would be the next logical step in June, whle back-up vs replace would be more of a UX thing (not sure if I'll do that in June). Hopefully tomorrow I should be able to get at least single-chunk reloading in.

StandingPadAnimations:
A couple of notes:
- `util` now has the `run_executable` function, which allows running an executable with arguments. In theory, this would supersede the old `open_program` function, but for now we'll keep it to avoid larger scale changes
- The single object reload operator is *technically* a third world importer, since it's from the base class used for the other world importers
- For now, the operator isn't exposed in the UI, since I want to try and get everything in the backend working first, and then overhaul the UI
- Only jmc2OBJ is supported

StandingPadAnimations:
*sigh* of course `xdg-open` doesn't support passing arguments

StandingPadAnimations:
I think for now we might have to replace the jmc2OBJ option in preferences with a simple string that has the basic launch command (i.e. `java -jar path/to/jMc2Obj.jar`)

StandingPadAnimations:
So I've gotten reloading working (at the moment, it requires `java` to be accessible system-wide, which I guess is a fair tradeoff for jmc2OBJ); *however*, for some reason, jmc2OBJ is returning different results. For context, here's the CommonMCOBJ headers (mostly the same), where the reloaded OBJ is on the left, and the original is on the right:

image

And here's what happens in Blender:

https://github.com/user-attachments/assets/c0f5febd-9bf0-4b21-b959-9c58154ffd64

It seems jmc2OBJ, when given the same settings, produces extremely different results in the command line than with the GUI (I've also tested manual importing of the reloaded OBJ, the OBJ itself is the issue).

@TheDuckCow could you test this on your end and see if the same issue occurs? I'm not sure if it's an issue on my end (since the settings appear to be correct, at least based on the header) or a jmc2OBJ bug.

StandingPadAnimations:
(Anyhow I'll mark this as ready for review, since the MCprep-side of things appears to be fine)

StandingPadAnimations:
Yeah nothing looks wrong in the command line (other than lack of quotes, but `subprocess.run` doesn't use a shell anyway, so that shouldn't be an issue, and everything file related is in proper order at runtime):

```
java -jar /path/to/jMc2Obj-128.jar --area=-696,802,-552,993 --height=12,314 --offset=center --scale=1.0 --output=path/to/reload_test --objfile=dream-house.obj --mtlfile=dream-house.mtl path/to/.minecraft/saves/Dream House _D
```

I tried seeing if maybe it's because we're writing over the original file, and nope, that does not appear to be the case. I also tried running the command on my own in bash (with quotes since now we're in the shell) and nope, issue still occurs. As far as I can tell, it seems to be a jmc2OBJ bug with the command line (since I cannot reproduce this with the GUI)

StandingPadAnimations:
Reported upstream: https://github.com/jmc2obj/j-mc-2-obj/issues/306

StandingPadAnimations:
Ok so this isn't actually a jmc2OBJ bug, the resource packs just need to be passed in the command line. For now, I'll probably create a basic `exporter_bridge_settings.json` file to make things simpler, since UI work would need a bit of extra time.

StandingPadAnimations:
Reloading works!

https://github.com/user-attachments/assets/28866f3e-fa98-43d7-9199-16ccfa10d0f0

TheDuckCow:
That's an amazing milestone already! Nice demonstration.

Adding some of my follow up questions and comments here, so I can easily get back to them in the future:


- Need to specify location to Minecraft install in a json like:
  ```
  {
      "resource_packs": [
          "path/to/minecraft.jar"
      ]
  }
  ``` 
  - Edit: I didn't need to do this in the end
- Need to have an initial world imported first, fresh world will come later
- For compiling, now need a python 3.11 venv (now configured on my side, one-time)
- Need to have an older Minecraft world - not sure if you facd the same, but it seems the latest jmc2obj (128) is not working with MC worlds 26.2.1 (the latest). I was able to get it to export one from v1.21
- And very much so remember: this is an _extension_ now, I had to manually uninstall the legacy MCprep addon before the new UI and operators were showing up



### Initial feedback after a brief look

Took me a bit of time, but I indeed got it working!

https://github.com/user-attachments/assets/949f639e-3ba7-404e-b49c-1ef6db84a81e

Some quick thoughts before doing a more thorough review:

- Good: Connection to jmc2obj work fined after I specified it in preferences, didn't have to give custom Minecraft jar path (though, instead of a resources file, I suggest we just define a setting for that)
  - But, in case I had needed to: I think the minecraft jar location should be located under the addon's settings rather than in a json in the resource pack, since you're likely to have even multiple resource packs used at once. Do you have an argument for keeping it under one like this?
- I think it would be nicer to have the export obj's (if retained, didn't check further) be saved to the same location as the original obj it was replacing, and if we can't know that or it doesn't exit, next to the blend file in a folder call `objs` or something like that. Keeps things connected to their corresponding projects.
- I would auto disable the collection of the prior/initially selected export, so you get the instant gratification of just seeing the new import
- I'd also argue for auto-applying Prep materials, but maybe that's a downstream effort as we unify the resources pack stack applying and such.
- We should try to find some sort of way to show a loading bar or other feedback while refreshing is in progress, especially if it can be large; right now Blender just hangs, which is expected. Curious if you have any ideas - simplest would be a modal operator.
  - May also need to contemplate cancellation of the action, in case someone realizes it's going too slow or got stuck.

StandingPadAnimations:
> But, in case I had needed to: I think the minecraft jar location should be located under the addon's settings rather than in a json in the resource pack, since you're likely to have even multiple resource packs used at once. Do you have an argument for keeping it under one like this?

Right now the JSON is a temporary measure. I do think it would be a good idea to eventually have a setting the Minecraft JAR, but UI/UX stuff I want to tackle at a later date.

> I think it would be nicer to have the export obj's (if retained, didn't check further) be saved to the same location as the original obj it was replacing, and if we can't know that or it doesn't exit, next to the blend file in a folder call objs or something like that. Keeps things connected to their corresponding projects.

Right now OBJs are not retained. I'm debating whether to make a folder of backed up OBJs, or add a numerical suffix (similar to blend1s) in the same location.

> I'd also argue for auto-applying Prep materials, but maybe that's a downstream effort as we unify the resources pack stack applying and such.

I think a better route would be to either call Combine Materials or use Blender's material name collision (5.0+), depending on the version. That way, user modifications to the materials are kept. Not sure why Blender 5.1 didn't use name collision for materials. 

> We should try to find some sort of way to show a loading bar or other feedback while refreshing is in progress, especially if it can be large; right now Blender just hangs, which is expected. Curious if you have any ideas - simplest would be a modal operator.

Yeah the UI hang-up isn't great. I'm not sure how we could do a loading bar though, given every exporter is different with console output. Maybe a time elapsed indicator at least?

TheDuckCow:
> UI/UX stuff I want to tackle at a later date.

Yup, totally fine with that.

> Right now OBJs are not retained

Ok great, in that case I'd say that's actually perfectly fine... let's not retain them, unless we find a reason to in the future. Would just clutter up local files anyways. 

> user modifications to the materials are kept

I guess if we do already have the pre-existing materials in the file and we can match to those existing ones, that is the safest short term option. For whatever it's worth, my longer term idea is that everything from resource packs and custom block models and prep settings ends up eventually being scene settings. In the case those materials don't exist right away, I'd argue for applying it anyways. Eventually, I could imagine prep materials is something that automatically runs over the connected worlds when you make changes like adding or removing resource packs, making it a higher level consideration. 

In the meantime, it's just nice for the demo's to have it pre-prepped ;) longer term UX flow around layered resources and such is another matter.

> I'm not sure how we could do a loading bar though
We could have the progress bar broken into phases, and just indicate whatever is possible for each of those phases. Maybe:
- first 10% just loading jmc2obj (if that can be split up)
- Next 60% is the actual jmc2obj export, if any data is streamed out during wowrld load vs export, we could try to capture that as additional text for a substage on top or something.
- Final 30% is loading into blender

Time lapsed either way would be nice. Beyond that, we could semi "fake" loading progress by having each stage lerp towards the target max for the given stage. This would be more to help showcase that things aren't stuck/working as normal, whereas looking at a loading bar that ins't moving at all is a signal to most any user that it's stuck (when it's more that we just don't have more information yet).

Again, being a UX thing, fine to addres slater - it' just nice to start thinking about what the options are.

StandingPadAnimations:
> I guess if we do already have the pre-existing materials in the file and we can match to those existing ones, that is the safest short term option. For whatever it's worth, my longer term idea is that everything from resource packs and custom block models and prep settings ends up eventually being scene settings. In the case those materials don't exist right away, I'd argue for applying it anyways. Eventually, I could imagine prep materials is something that automatically runs over the connected worlds when you make changes like adding or removing resource packs, making it a higher level consideration.

Whatever we do with resource packs, I think at the end of the day, user modifications should be preserved by default. In my view, a good middle ground (at least as a default setting) would be to only prep new materials, and carry over old materials (which as of Blender 5.0 can simply be handled in the OBJ importer). That said, that would require some larger changes, so maybe that could be done at a later point.

>  We could have the progress bar broken into phases, and just indicate whatever is possible for each of those phases. Maybe:
>
> - first 10% just loading jmc2obj (if that can be split up)
> - Next 60% is the actual jmc2obj export, if any data is streamed out during wowrld load vs export, we could try to capture that as additional text for a substage on top or something.
> - Final 30% is loading into blender

That might be a good option, though that's definitely something to tackle at a later date. Might be worth discussing more in depth in the general world importer discussion

StandingPadAnimations:
Interestingly, it seems there's a slight shift in the OBJ when reloading. I'll have to look into that

StandingPadAnimations:
Hmm, oddly the shift doesn't seem like it should happen in jmc2OBJ:

https://github.com/jmc2obj/j-mc-2-obj/blob/be42c6a2df0e0c946504de16da24cea32f35de72/src/org/jmc/CmdLineParser.java#L133-L140

Yet the CommonMCOBJ metadata is indicating a shift of -1 for `export_bounds_min` when reloading. I've resolved this on my end by adding 1, but @TheDuckCow let me know if this caues a shift on your end.

StandingPadAnimations:
I've converted the operator to a modal operator so now the UI remains responsive while the export is being done (it still hangs when importing, but that's to be expected.

Also made sure to set the optimization arguments in jmc2OBJ, since they don't seem to be the default when running in the CLI (but they are the default in the GUI). Reloaded OBJ without optimization arguments:
image

vs with:
image

MCprep_addon/exporter_glue/jmc2obj.py
```diff
@@ -0,0 +1,125 @@
...
+        if base_check is not None:
+            return base_check
+        if not self.world_path:
+            line, file = env.current_line_and_file()
+            return MCprepError(Exception(), line, file, "No world path set!")
+        self.args += [f"--output={str(self.output_obj_path.parent)}",
+                      f"--objfile={str(self.output_obj_path.name)}",
+                      f"--mtlfile={str(self.output_obj_path.with_suffix('.mtl').name)}",
+                      "--remove-dup",
+                      "--optimize-geometry",
```
TheDuckCow, line 122: This is one of those things where sometimes you want it, sometimes you don't. e.g. if you want to meshswap for some generic 3D blocks like grass, optimization breaks that.

I'd suggest we create some kind of way to pass in a flag for this specifically, or maybe a dataclass structure of settings so that users can set things up the way they need. Doesn't need to have everything, but would increase its usability. That goes for the block randomization as well I'd say. For Mineway's we'd have different settings as well, so it's worthwhile to create a structure for it I'd say

When we get to the UI layer, we probably should have some custom draw options for the initial import settings to leverage, so it's a _similar_ popup once you confirm which world you want to load, or a separate setting that can be applied on re-importing (though that could be annoying if you're always forced to export twice to set a specific setting).

I'm also fine if we want to defer this to a follow up task, but figured I'd mention it. This PR is already a big step forward.

MCprep_addon/conf.py
```diff
@@ -151,10 +151,21 @@ def __init__(self):
 		self.vivy_material_json: Dict = {}
 		self.vivy_enabled = False
 
+		# This is mainly intended for operations
+		# like world reloading, where the user can
+		# still move around the scene, but certain
+		# features shouldn't be used
+		self.global_lock: bool = False
```
TheDuckCow, line 158: Should we instead just have this be an actual python lock?  e.g. `world_lock = threading.Lock()` and then in various functions have `lock.acquire()` and `lock.release()` (if split over functions, if within the same you could use a `with` context manager). Then, the places we are checking for this lock would actually be acquirers using `acquire(blocking=False)` and releasing shortly after as necessary.

I know we use bool locks elsewhere, but they aren't safe and best to put our best foot forward to avoid downstream headaches I'd say.

MCprep_addon/exporter_glue/jmc2obj.py
```diff
@@ -0,0 +1,125 @@
...
+        if base_check is not None:
+            return base_check
+
+
+        self.xz_min_max = (min_bounds[0],min_bounds[2],max_bounds[0],max_bounds[2])
+
+        # Adjustment is required for the max bounds to avoid shifts
+        min_max_adjusted = (min_bounds[0],min_bounds[2],max_bounds[0] + 1,max_bounds[2] + 1)
+        self.args += [f"--area={','.join(str(x) for x in min_max_adjusted)}",
+                      f"--height={min_bounds[1]},{max_bounds[1] + 1}"]
```
TheDuckCow, line 56: Haven't tested yet, I can do next time I do a live testing review, but is also something that will become clear with more usage.

MCprep_addon/exporter_glue/jmc2obj.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Kind of a horrifying block, but this
+        # is enough to make Mypy not freak out from
+        # type narrowing
+        def average_offset(m: int, n: int) -> int:
+            return int(-1 * (m + (n - m) / 2))
+
+        # Use an assert as this function should
+        # be called after the bounds are set
+        assert self.xz_min_max is not None
```
TheDuckCow, line 80: Should this be a runtime error or value error instead?

MCprep_addon/exporter_glue/jmc2obj.py
```diff
@@ -0,0 +1,125 @@
...
+        z_average = average_offset(self.xz_min_max[1], self.xz_min_max[3])
+        print(offset[0], offset[2])
+        print(x_average, z_average)
+        if offset[0] == x_average and offset[2] == z_average:
+            self.args.append("--offset=center")
+        else:
+            self.args.append(f"--offset={offset[0]},{offset[2]}")
+
+    @override
+    def set_block_scale(self, scale: float) -> None | MCprepError:
```
TheDuckCow, line 92: More of a style Q, I know we have pretty much a consistent convention of using `|` but how do you feel about `Optional[MCprepError]`? Slightly more verbose but is also fairly clear. it's another import though. If we stick with `None | {Type}` then we should document in a styleguide doc somewhere.

(noting, we do use Optional in other files, but I definitely wouldn't say there's repo-wise consistency)

TheDuckCow (commented):
Generally looking good - I probably should do one more live test, let me know if there's anything else in progress with this base implementation. Then, we should talk about the breakdown of the next steps. I've had some thoughts about what the MVP vs extended versions are.

Also, do we want the world bridge to be a feature branch or merge into milestone 4.0? Sort of depends on how many steps we have left I'd say. I'm fine to merge into the milestone.

MCprep_addon/conf.py
StandingPadAnimations: I'm not too familiar with the Python threading API, but I'd imagine that might be a better option

MCprep_addon/exporter_glue/jmc2obj.py
StandingPadAnimations: Yeah I'm not sure if it's just me having issues on Linux or if jmc2OBJ requires adjustment regardless for consistency. I would suspect it's related to the block offset, but it's hard to know.

MCprep_addon/exporter_glue/jmc2obj.py
StandingPadAnimations: The main reason for `assert` as opposed to returning an `MCprepError` object is because this is situation would only happen if the sequence of functions called is incorrect (which in that case would be a legitimate bug). IMO it's cleaner to have a single line asserting `self.xz_min_max` isn't `None` (and satisfying type checkers) then having a conditional with a return branch.

MCprep_addon/exporter_glue/jmc2obj.py
StandingPadAnimations: I prefer using the `|` syntax where possible since it's less verbose (and covers the use cases of both `Union` and `Optional`). I do agree there should be a style guide; my personal conventions right now are:

- For a function that can return multiple types: no particular order (ex. `int | float`), but if `None` is present, then `None` should be last (ex. `int | float | None`)
- For a function that can return a valid object or an error object: `Object | Error`
- For a function that can return a valid object or `None`: `Object | None`
- For a function that can return `None` or an error object: `None | Error`

Essentially, if there's an error object in the type signature, it's always be the very last item, and `None` should be last *unless* there's also an error object in the type signature. I think as a general rule of thumb, it's simple enough to consistently apply in new code.

Edit: The conventions above could be simplified to something like this (taking some inspiration from the Zen of Python):

No particular order by default,
Except for `None`, which should be last,
Unless there's an error object, then it is last,
And `None` is second to last

(~~I never claimed to be good at poetry~~)

MCprep_addon/exporter_glue/jmc2obj.py
StandingPadAnimations: I think if we have settings expost in the UI, then we need to just have common settings found in most exporters, and then set flags accordingly. The options shown to the end-user would be simplified, and IMO, exporter-specific options are best done in an exporter anyway (ex. setting chunks as foreground in MiEx)

MCprep_addon/exporter_glue/jmc2obj.py
TheDuckCow: Ok, I'm fine with aligning on this. I do like the simplicity of having the error be last as a quick way to see it.

Not a rush (aka can be it's own separate PR), but might be nice to add a `docs/codestyle.md` file that covers this and things like the fact that we in theory extend from the official Python followed by Blender style guide, and that these are just rules on top.

MCprep_addon/exporter_glue/jmc2obj.py
TheDuckCow: I'm still torn about it, since needing a specific setting for your own use case can make or break your ability to use the tool itself. And I don't think it'd be that hard to define a UI pattern of mapping a database of prop types to a procedural UI draw function.

But on the other hand: I'm increasingly thinking that for the MVP of the world bridge, we may want to stick with assuming the user starts with their own initial export (as in loading the file from disk, not dealing with a pre-imported world). This gets around short term the whole UX of finding the right area to import, which almost certainly will be a worse or slower UX experience OR requires us to do direct world parsing ourselves (which when we get to that point, is probably what we'd need to do to make it snappy).

Then, the operations we are providing are to:

1. refresh the current export area
2. Extend the export area (possibly with different modes for "background extend" vs "foreground extension"). 

Especially for refreshing, I'd say it's important to have as consistent settings as the original export as possible. But, if optimize isn't in the spect that gets saved (which I believe is the case), and similar to Mineways show or merge occluding floating faces / merge verts etc, we already risk refreshing something that was different to before.

I wonder of we should add something like an "optimize_mesh" property to the CommonMCobj spec. We could bucket them all together, maybe it's both merge verts and optimize mesh (which creates single faces spanning multiple blocks where possible). Or, have the spec allow for exporters to add additional settings if they want, and then it's just up to any given importer to read those settings? We could then still use the spec, and just have checks for if optional property by specific name is found, it gets mapped back to a given flag for an exporter.

A bit late in the game to suggest a spec change, but the second best time to plant a tree is today as they say. What do you think?

TheDuckCow: We can revisit this after the initial MVP workflow, we'll probably do some user interviews to find out what is most important/key missing features once we have this MVP in place.

MCprep_addon/blender_manifest.toml
TheDuckCow, line 1: I realize we now have some of the 4.0 stuff in this branch - that's fine for now, I'm hoping to prioritize the world branch MVP to completion in the next month or two anyways

TheDuckCow (approved):
Approving as discussed - some follow ups to revisit after, but I'd say this is worth merging so we can start talking next steps.

To best understand the next priorities, I do think we're at the stage that we should think and decide what the actual UX flow will look like for direct imports so we can think how it'll all fit together.

PR #715, opened 2026-06-13
Fix: Silent handling of permission errors & animation material pass mapping errors.

mozi1924:
### 1. Permission errors now trigger a warning and cancel the operation
**Issue**: When importing texture packs on Unix systems, folder creation could fail due to permission issues; however, the error was silently suppressed, resulting in the import of incorrect textures.

**Fixes**:
- `sequences.py`: All calls to `animate_single_material()` now check the return value; if a permission error occurs, the operation is cancelled and an error dialog is displayed.
- `prep.py`: 3 call sites fixed.
- `material_manager.py`: 1 call site fixed.
- `mobs.py`: Permission errors for the icon folder now trigger a warning instead of just printing to the console.

### 2. Animated material pass mapping error
**Issue**: When applying a PBR texture pack to animated materials, the `magma_n` (normal) and `magma_s` (specular) sequences were incorrectly applied to the same Diffuse Texture node.

**Root Causes**:
- `np_is_mcprep_node_prop` destructively set `node.mnp.X` to `False` when the old dictionary property was missing.
- The `else` fallback in `get_node_for_pass` treated the first `TEX_IMAGE` node as the fallback for *all* passes.
- `set_cycles_texture` did not create missing normal/specular nodes.

**Fixes**:
- `util.py`: `np_is_mcprep_node_prop` no longer performs destructive writes.
- `util.py`: `apply_noncolor_data` now returns safely when `node.image` is `None`.
- `generate.py`: The fallback in `get_node_for_pass` now applies only to the diffuse pass.
- `generate.py`: Added `_create_missing_pass_nodes` to automatically create missing tagged nodes.

### 3. Cross-platform wording for permission errors
**Issue**: Permission error prompts suggested running Blender as administrator, which was not applicable to Linux/macOS users.

**Fixes**: All permission error prompts have been updated to offer generic advice regarding folder/file permissions. ### Involved files
- `MCprep_addon/materials/sequences.py`
- `MCprep_addon/materials/prep.py`
- `MCprep_addon/materials/material_manager.py`
- `MCprep_addon/materials/generate.py`
- `MCprep_addon/spawner/mobs.py`
- `MCprep_addon/spawner/mcmodel.py`
- `MCprep_addon/util.py`

### DIFF
--- a/MCprep_addon/materials/generate.py
+++ b/MCprep_addon/materials/generate.py
@@ -458,7 +458,7 @@ def set_cycles_texture(
 				node.image = new_img
 				node.mute = False
 				node.hide = False
-				
+
 				res = util.apply_noncolor_data(node)
 				if res is not None:
 					env.log(f"TypeError on {res.line} in {res.file}: {res.err_type}")
@@ -501,7 +501,56 @@ def set_cycles_texture(
 
 		changed = True
 
+	# If extra passes exist but no labeled node was found, create them
+	if extra_passes and img_sets:
+		_create_missing_pass_nodes(material, img_sets)
+
 	return changed
+
+
+def _create_missing_pass_nodes(
+	material: Material, img_sets: dict) -> None:
+	"""Create missing normal/specular texture nodes if extra passes exist
+	but no labeled nodes are present in the material."""
+	if not material.node_tree:
+		return
+	nodes = material.node_tree.nodes
+
+	# Check if labeled nodes already exist
+	has_normal = any(
+		util.np_is_mcprep_node_prop(n, MCPREP_NORMAL)
+		for n in nodes if n.type == 'TEX_IMAGE')
+	has_specular = any(
+		util.np_is_mcprep_node_prop(n, MCPREP_SPECULAR)
+		for n in nodes if n.type == 'TEX_IMAGE')
+
+	# Create normal node if needed
+	if "normal" in img_sets and not has_normal:
+		new_img = util.loadTexture(img_sets["normal"])
+		node_nrm = create_node(
+			nodes, 'ShaderNodeTexImage',
+			name="Normal Texture",
+			label="Normal Texture")
+		node_nrm.mnp.MCPREP_normal = True
+		node_nrm.image = new_img
+		res = util.apply_noncolor_data(node_nrm)
+		if res is not None:
+			env.log(f"TypeError on {res.line} in {res.file}: {res.err_type}")
+		env.log("Created missing normal texture node for extra passes")
+
+	# Create specular node if needed
+	if "specular" in img_sets and not has_specular:
+		new_img = util.loadTexture(img_sets["specular"])
+		node_spec = create_node(
+			nodes, 'ShaderNodeTexImage',
+			name="Specular Texture",
+			label="Specular Texture")
+		node_spec.mnp.MCPREP_specular = True
+		node_spec.image = new_img
+		res = util.apply_noncolor_data(node_spec)
+		if res is not None:
+			env.log(f"TypeError on {res.line} in {res.file}: {res.err_type}")
+		env.log("Created missing specular texture node for extra passes")
 
 
 def get_node_for_pass(material: Material, pass_name: str) -> Optional[Node]:
@@ -522,9 +571,10 @@ def get_node_for_pass(material: Material, pass_name: str) -> Optional[Node]:
 			return_node = node
 		elif util.np_is_mcprep_node_prop(node, MCPREP_DISPLACE) and pass_name == "displace":
 			return_node = node
-		else:
-			if not return_node:
-				return_node = node
+		# Only use unlabeled fallback for diffuse pass, to avoid
+		# incorrectly applying normal/specular sequences to diffuse node
+		elif pass_name == "diffuse" and not return_node:
+			return_node = node
 	return return_node
 
 
--- a/MCprep_addon/materials/prep.py
+++ b/MCprep_addon/materials/prep.py
@@ -272,10 +272,13 @@ class MCPREP_OT_prep_materials(bpy.types.Operator, McprepMaterialProps):
 				return {'CANCELLED'}
 
 			if self.animateTextures:
-				sequences.animate_single_material(
+				_, _, err = sequences.animate_single_material(
 					mat,
 					context.scene.render.engine,
 					export_location=sequences.ExportLocation.ORIGINAL)
+				if err:
+					self.report({'ERROR'}, str(err))
+					return {'CANCELLED'}
 
 		# Sync materials.
 		if self.syncMaterials is True:
@@ -509,13 +512,15 @@ class MCPREP_OT_swap_texture_pack(
 			self.preprocess_material(mat)
 			res += generate.set_texture_pack(mat, Path(folder), self.useExtraMaps)
 			if self.animateTextures:
-				sequences.animate_single_material(
+				_, _, err = sequences.animate_single_material(
 					mat,
 					context.scene.render.engine,
 					export_location=sequences.ExportLocation.ORIGINAL)
+				if err:
+					self.report({'ERROR'}, str(err))
+					return {'CANCELLED'}
 			# may be a double call if was animated tex
-			generate.set_saturation_material(mat)
-
+		generate.set_saturation_material(mat)
 		if self.prepMaterials:
 			bpy.ops.mcprep.prep_materials(
 				animateTextures=self.animateTextures,
@@ -660,10 +665,13 @@ class MCPREP_OT_load_material(bpy.types.Operator, McprepMaterialProps):
 		success = res == 0
 
 		if self.animateTextures:
-			sequences.animate_single_material(
+			_, _, err = sequences.animate_single_material(
 				mat,
 				context.scene.render.engine,
 				export_location=sequences.ExportLocation.ORIGINAL)
+			if err:
+				self.report({'ERROR'}, str(err))
+				return False, str(err)
 
 		return success, None
 
--- a/MCprep_addon/materials/material_manager.py
+++ b/MCprep_addon/materials/material_manager.py
@@ -420,10 +420,13 @@ class MCPREP_OT_replace_missing_textures(bpy.types.Operator):
 				count += 1
 				env.log(f"Updated {mat.name}")
 				if self.animateTextures:
-					sequences.animate_single_material(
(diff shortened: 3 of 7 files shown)


MCprep_addon/materials/generate.py
```diff
@@ -501,9 +501,58 @@ def set_cycles_texture(
...
 		changed = True
 
+	# If extra passes exist but no labeled node was found, create them
+	if extra_passes and img_sets:
+		_create_missing_pass_nodes(material, img_sets)
+
 	return changed
 
 
+def _create_missing_pass_nodes(
```
TheDuckCow, line 511: I need to spend a moment trying to replicate this more specifically - is this when trying to animate textures only *after* prep materials is run, or does this also happen after running the texture swap first?

TheDuckCow (commented):
Thanks for the contribution. Leaving an initial comment to get some clarification - in general the messaging and the error handling does look better (I'm pondering whether some aspects of it could be resused), the material code changes look to be a little different and almost warrant their own separate review.

mozi1924 (commented):
Good question — here's the specific reproduction scenario:

**The bug triggers specifically after "Swap Texture Pack" with "Animate Textures" enabled**, NOT after a standalone prep materials (though the `_create_missing_pass_nodes` fix also benefits `generate_base_material` → `Load Material` flow).

### Exact reproduction flow:

1. Import a jmc2obj model that has animated textures (e.g., magma block with `.mcmeta` animation data)
2. The material gets prepped with SIMPLE pack format → only one `TEX_IMAGE` node labeled `MCPREP_diffuse`, no normal/specular nodes exist
3. User opens "Swap Texture Pack" → selects a PBR resource pack → enables both "Use Extra Maps" AND "Animate Textures"
4. `set_texture_pack` → `set_cycles_texture(image, mat, extra_passes=True)` runs:
   - `find_additional_passes` discovers `magma_n.png` and `magma_s.png` in the pack
   - But no nodes have `MCPREP_normal`/`MCPREP_specular` labels → the `elif swap_all_imgs` branch labels ALL nodes as `MCPREP_diffuse`
   - Normal/specular images are never applied
5. `animate_single_material` runs → `generate_material_sequence` generates tile sequences for diffuse, normal, specular
6. `get_node_for_pass(mat, "normal")` can't find any `MCPREP_normal` node → old `else` fallback returns the diffuse node → **normal sequence gets applied to Diffuse Texture node** 💥

### The two-part fix:
- **`_create_missing_pass_nodes`**: When `extra_passes=True` but no labeled nodes exist, automatically creates them — so step 4 actually applies the images
- **`get_node_for_pass` fallback restriction**: Limits the `else` fallback to only diffuse pass, preventing the wrong node from being returned even if labels are somehow missing

### There's also a secondary path:
If the material was created via `generate_base_material` (Load Material), it initially sets `node.mnp.MCPREP_normal = True` correctly, but the old `np_is_mcprep_node_prop` destructively overwrote it to `False` on read — that's why the util.py fix is also needed.

---

Regarding testing: we can prepare a minimal test resource pack (just a magma block with animated textures + normal/specular variants) to make reproduction straightforward. Would that be helpful?

mozi1924:
I have prepared a branch for this:
https://github.com/mozi1924/MCprep/tree/demo/before-fix

It includes a `sample.blend` file.

The resource pack I used is: https://modrinth.com/resourcepack/vanilla-mashup-pbr/version/1.5

To reproduce the issue: select the object (the flame),
then click "Swap Texture Pack."
In the file picker, navigate to the resource pack directory and check the options for "Animated Materials" and "Prep Materials."
At this point, MCPrep incorrectly treats the Specular Texture as the Diffuse Texture.

MCprep_addon/materials/prep.py
```diff
@@ -272,10 +272,13 @@ def execute(self, context):
 				return {'CANCELLED'}
 
 			if self.animateTextures:
-				sequences.animate_single_material(
+				_, _, err = sequences.animate_single_material(
 					mat,
 					context.scene.render.engine,
 					export_location=sequences.ExportLocation.ORIGINAL)
+				if err:
+					self.report({'ERROR'}, str(err))
```
StandingPadAnimations, line 280: For `sequences.animate_single_material`, the last return value is a string anyway, so no conversion is needed.

MCprep_addon/materials/generate.py
```diff
@@ -501,9 +501,58 @@ def set_cycles_texture(
...
+		node_spec = create_node(
+			nodes, 'ShaderNodeTexImage',
+			name="Specular Texture",
+			label="Specular Texture")
+		node_spec.mnp.MCPREP_specular = True
+		node_spec.image = new_img
+		res = util.apply_noncolor_data(node_spec)
+		if res is not None:
+			env.log(f"TypeError on {res.line} in {res.file}: {res.err_type}")
+		env.log("Created missing specular texture node for extra passes")
```
StandingPadAnimations, line 553: Could you elaborate on what real world issue this function is resolving? Outside of a user using `Simple` with a PBR resource pack, I can't think of a case where specular and normal maps won't be created. In MCprep, the only material options with PBR support are `Specular` and `SEUS`; `Simple` is intended for non-PBR use cases, as in the past, there have been multiple PBR standards for MC resource packs (hence why we require the user to explicitly opt-in to one of the PBR options).

MCprep_addon/materials/generate.py
StandingPadAnimations, line 511: Style note: leading `_` should be avoided in function names, mainly because it looks weird.

MCprep_addon/materials/generate.py
```diff
@@ -501,9 +501,58 @@ def set_cycles_texture(
 
 		changed = True
 
+	# If extra passes exist but no labeled node was found, create them
+	if extra_passes and img_sets:
+		_create_missing_pass_nodes(material, img_sets)
```
StandingPadAnimations, line 506: See my comment on this function as a whole.

StandingPadAnimations (changes requested):
I think there's a bit of a misunderstanding with the behavior of `set_cycles_textures` and its `extra_passes` argument.

`set_cycles_textures` is intended to be ran after the nodes for a material have been created, with a single job: find the proper textures for the material, and set them. However, what textures it will use will ultimately be dictated by the material.

The `extra_passes` argument exists to tell the function to search for additional passes other than diffuse; it *does not* dictate that those passes *have* to be used, that is up to the material tree itself.

There are legitimate reasons for a user to want to use a PBR resource pack without using the PBR maps (ex. if they like the design of the blocks, but don't like the PBR aspects). The `Simple` option in prep materials would be the option to select, as it's intended to not be a PBR material.

There's also just issues with automatically adding PBR nodes, namely the lack of a single PBR standard for Minecraft resource packs. Just on the top of my head, we have:

- LabPBR ([which requires a lot of additional logic in node setups](https://shaderlabs.org/wiki/LabPBR_Material_Standard))
- SEUS (though this spec is slowly dying off)
- RTX and Vibrant Visuals (will become more important once Vibrant Visuals is implemented in Java Edition)

Manual selection is still needed to properly handle multiple specs. Automatically adding nodes would effectively break resource packs.

Unless there's a *real world case* where the nodes for specular and normal aren't being added *when using the `Specular` and/or `SEUS` options* in prep materials, the additional logic for PBR passes in `set_cycles_texture` should be removed.

As an aside: swap textures has the option to prep materials afterwards, intended for switching to the `Specular` or `SEUS` options if needed. The issue mentioned above should be resolved with this.

image

StandingPadAnimations:
Also I might be wrong, but if an LLM is being used for this contribution, [we have rules regarding their use](https://github.com/Moo-Ack-Productions/MCprep/blob/master/CONTRIBUTING.md#use-of-ai-in-development)

mozi1924:
> Also I might be wrong, but if an LLM is being used for this contribution, [we have rules regarding their use](https://github.com/Moo-Ack-Productions/MCprep/blob/master/CONTRIBUTING.md#use-of-ai-in-development)

Thanks for the review!
The initial fix was indeed created with the help of an LLM; however, I reviewed all the changes, tested them in my own environments (Mac/Windows) to confirm the issue was genuinely resolved rather than just coincidentally fixed, and performed comparative testing.
I fully understand the concerns regarding LLM-assisted contributions and am willing to close this PR if necessary. Otherwise, I am happy to address any review comments or make further improvements.
Thanks again for taking the time to review this.
(Apologies for using a translator.)

mozi1924:
I also agree that the `_create_missing_pass_nodes` logic should not be included. The purpose of the `extra_passes` flag is to allow `set_cycles_texture` to look for additional textures, not to mandate their existence. Automatically creating PBR nodes would indeed break resource packs that follow different standards (such as LabPBR, SEUS, or RTX), and the Simple format was intentionally designed to be non-PBR. I will remove it.

The fix remains effective even after removing that logic—the key change lies in the `get_node_for_pass` function: by restricting the fallback result in the `else` branch to the diffuse texture only, the function returns `None` when nodes marked as normal or specular are missing; meanwhile, `animate_single_material` already skips `None` values ​​via the `if not node: continue` logic. Consequently, the sequence is not applied, preventing it from being incorrectly applied to the diffuse node. If users require animated PBR textures, they can use the Specular/SEUS preprocessing options.

StandingPadAnimations:
> > Also I might be wrong, but if an LLM is being used for this contribution, [we have rules regarding their use](https://github.com/Moo-Ack-Productions/MCprep/blob/master/CONTRIBUTING.md#use-of-ai-in-development)
> 
> Thanks for the review! The initial fix was indeed created with the help of an LLM; however, I reviewed all the changes, tested them in my own environments (Mac/Windows) to confirm the issue was genuinely resolved rather than just coincidentally fixed, and performed comparative testing. I fully understand the concerns regarding LLM-assisted contributions and am willing to close this PR if necessary. Otherwise, I am happy to address any review comments or make further improvements. Thanks again for taking the time to review this. (Apologies for using a translator.)

Are LLMs being used for creating comments though (beyond simply translating while preserving meaning and being close to the original wording)? To me, it appears the comments describing the changes made in this pull request are generated by an LLM, beyond simply translating; while we don't explicitly mention it in `CONTRIBUTING.md`, at least in my opinion, comments describing changes and what issues they resolve fall under the first condition:

> - The AI is supplemental to the developer's work, not the other way around.

mozi1924:
Thanks for the reply.

I don't think there is any need to continue this discussion, so I will be closing this PR now and maintaining the changes in my own fork.

Thanks for your time and feedback.

PR #719, opened 2026-06-15
Handle 26.1 JSON model changes and properly resize elements

StandingPadAnimations:
Minecraft 26.1 introduced a change where textures can now be a dictionary defining a sprite (the texture itself) and some additional properties. Thus, we can no longer assume textures on JSON models will always be a string.

Really we should probably refactor the JSON importer code to make it less fragile, but for now, this band-aid works.

This fix should be backported to MCprep 3.6.X, as Minecraft 26.1 has already been out for some time, and Minecraft 26.2 is coming out soon.

### DIFF
--- a/MCprep_addon/spawner/mcmodel.py
+++ b/MCprep_addon/spawner/mcmodel.py
@@ -38,7 +38,7 @@ from .spawner_gizmo import draw_callback
 TexFace = Dict[str, Dict[str, Union[str, List[int]]]]
 
 Element = Sequence[Union[Dict[str, VectorType], TexFace]]
-Texture = Dict[str, str]
+Texture = Dict[str, Union[str, Dict[str, str]]]
 
 try:
 	from bpy.types import FileHandler
@@ -213,8 +213,7 @@ def find_all_pack_roots(
 
 	return final
 
-def normalize_texture_path(
-	path: str) -> str:
+def normalize_texture_path(path: Union[str, Dict[str, str]]) -> str:
 	"""
 	Path normalizer.
 	Takes different and any weird path and rewrites it into a formatted, useable form.
@@ -228,6 +227,11 @@ def normalize_texture_path(
 		"block/foo.png"		-> "block/foo"
 		"minecraft:block/foo" stays unchanged except path cleanup
 	"""
+
+	if not isinstance(path, str) and "sprite" in path:
+		path = path["sprite"]
+
+	assert isinstance(path, str)
 
 	# Strip leading '#' if passed a reference by mistake
 	if path.startswith("#"):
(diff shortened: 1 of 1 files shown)


StandingPadAnimations:
For some additional context, this change fixes a current bug where stained glass models in Minecraft 26.1 and above cannot be imported with MCprep's JSON importer.

TheDuckCow:
Can you share what the original error was and experience before this change, so I can help confirm it's indeed fixed? I tried with and without these changes after extracting the latest resources from 26.1, wasn't seeing an explicit error. Didn't try a world import though

(side note, I'm thinking this could be a good convention to add in general, a required PR format where you specify how to actually test the changes, and the clear before/after results, so it's very reproducible).

StandingPadAnimations:
> Can you share what the original error was and experience before this change, so I can help confirm it's indeed fixed? I tried with and without these changes after extracting the latest resources from 26.1, wasn't seeing an explicit error. Didn't try a world import though

In Minecraft 26.1, the texture key for JSON models can now be a dictionary with some additional properties; currently, this is mainly used by the stained glass models.

If you attempt to import certain JSON models (like stained glass from Minecraft 26.1), the importer will raise an exception when resolving textured as at the moment, textures are assumed to be a string.

TheDuckCow (approved):
I wasn't able to properly test as jmc2obj is failing to export these kinds of blocks, but trusting I'll spend some time further covering it in my release testing on windows where I can validate both further. Agreed, we should aim to make an update sooner than later.

StandingPadAnimations:
> I wasn't able to properly test as jmc2obj is failing to export these kinds of blocks, but trusting I'll spend some time further covering it in my release testing on windows where I can validate both further. Agreed, we should aim to make an update sooner than later.

This change relates to JSON models (specifically models such as stained glass), just to clarify (so local refresh with Minecraft 26.1 should be all that's required). I just want to make sure that it works on multiple machines before merging.

StandingPadAnimations:
@TheDuckCow  I've incorporated another fix in this branch related to JSON models. This one resolves a bug that occurs when spawning models such as `short_grass` and the `firefly_bush`, where the elements are not properly rescaled prior to rotation. To quote the commit:

```
fix: Use rescale parameter for block rotations where necessary

Some blocks in Minecraft have a rescale parameter to account for
distortions that may occur after rotating, where the elements are scaled
by a factor of 1 / cos(rotation_angle) along the other axis. Without
this, blocks such as short_grass and firefly_bush remain thin and tall.

Signed-off-by: Maryam Sheikh <mahid@standingpad.org>
```

Here's a before (current MCprep behavior, spawned blocks are too tall and thin compared to their in-game counterparts):

image

And an after:

image

Unlike the earlier bug, which is related to a 26.1 change, this bug (as far as I know) affects JSON models going back to Minecraft 1.8 (at least based on what [Minecraft.Wiki says](https://minecraft.wiki/w/Model#History)), so we should probably create an MCprep 3.6.3 release with these fixes.

Also, since I probably didn't demonstrate it, here's a demonstration on the earlier change, which is required due to a change in the `texture` parameter for blocks such as stained glass in Minecraft 26.1:

https://github.com/user-attachments/assets/a3151838-b811-43dc-9f0a-e49860b63aa7

TheDuckCow (approved):
Just live-tried it, works as described. Agreed with the new version too. Merging this PR shortly.

PR #720, opened 2026-06-15
docs: Clarify rules surrounding AI

StandingPadAnimations:
We added rules around AI three years ago. While back then, they were mostly reasonable, clarification is now needed with the advent of AI agents.

The main change here is regarding interaction with the MCprep project, inspired by Matplotlib (https://matplotlib.org/devdocs/devel/contribute.html#generative-ai). Under the clarified rules, interacting with the MCprep project through AI tooling (including but not limited to bots, agents, plugins, etc) will result in an automatic rejection of pull requests.


TheDuckCow (approved):
Reasonable updates

PR #722, opened 2026-06-30
fix: Limit Mineways shading workaround to under Blender 5.1

StandingPadAnimations:
The core issue we had to work around with Mineways imports isn't needed in Blender 5.1 and above, so we shouldn't apply it to Mineways imports done post-Blender 5.1.

- Initial bug report in Mineways: https://github.com/erich666/Mineways/issues/151
- Issue made by Eric upstream: https://projects.blender.org/blender/blender/issues/152496
- Pull request where issue is resolved: https://projects.blender.org/blender/blender/pulls/153836

Also some demonstrations of the fix in upstream Blender. Below is Blender 4.5, where a Mineways export is imported through the regular OBJ importer (not through MCprep) and Cycles has issues with custom normals:

Screenshot_20260630_020648

And here's the exact same file opened up in Blender 5.1, where the issue has been resolved:

Screenshot_20260630_020758

### DIFF
--- a/MCprep_addon/world_tools.py
+++ b/MCprep_addon/world_tools.py
@@ -799,7 +799,7 @@ class WorldImporterBase:
 				self.track_exporter = header.exporter
 
 				# Mineways workaround for smooth shading artifacts
-				if mineways_fix_smooth_shading_artifacts and header.exporter == "mineways":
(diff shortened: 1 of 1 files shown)


TheDuckCow (approved):
Change looks safe to me, sorry for the delay - good to merge

PR #724, opened 2026-07-11
Fixed moon and sun mesh importing

TheDuckCow:
Previously, the mesh sun+moon importer was not working at all. It must have accidentally been changed while making some other typing updates. 

Now, the mesh sun/moon do properly import as they should have, and tests now better ensure this behavior is not lost again by accident.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ mcprep_venv_*
 .python-version
 venv/
 .venv/
+venv39/
 *.sublime-*
 MCprep_addon/import_bridge/conf
 MCprep_addon/import_bridge/nbt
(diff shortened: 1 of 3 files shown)


TheDuckCow:
The error is _related_ but not the entire issue, my local tests will expand on this:

```
FAIL: test_add_mc_sky (world_tools_test.WorldToolsTest.test_add_mc_sky) [world_static_mesh]
  File "/home/runner/work/MCprep/MCprep/test_files/world_tools_test.py", line 365, in test_add_mc_sky
    self.assertFalse(obj, "Static scn should have no timeobj")
AssertionError: bpy.data.objects['MCprepSunMoonControl'] is not false : Static scn should have no timeobj
```

